PR #15, opened 2020-09-29
Adds metadata structs, global cut percentage, global delegator registration, and splits delegators stake

### LINKED ISSUE #12: Enable a NodeRecordStatus struct to be returned
### Issue To Be Solved
If you want to know the status of a NodeRecord right now you have to call lots of methods to fetch different fields. Having the ability to return a single NodeRecordStatus struct of a map of all of them would make it easier to write scripts on top of the api to show what is staked. 

### (Optional): Suggest A Solution

Create a NodeRecordStatus struct with calculated and safe properties and return that. Kind like versus does it here: https://github.com/versus-flow/auction-flow-contract/blob/master/contracts/Auction.cdc#L13

### (Optional): Context

Nothing, just want a nice API.

joshuahannan:
Based on some conversations this week, I've made a few changes.

* Changed the delegator cut percentage to a global value because it cannot be chosen by individual node operators. Also fixed it to 3%
* Made the register delegator function a global public function in the contract, so that node operators cannot prevent delegating to their nodes
* Split the delegators' stakes into their own buckets instead of being combined with the node's buckets
* Combined all the node and delegator info into their own structs so that transactions and scripts could get all the data associated with a user in one call and one struct instead of individual transactions. Thanks @bjartek! Closes #12 
* Updated the transactions and tests to reflect the changes


transactions/idTableStaking/get_node_info.cdc
```diff
@@ -0,0 +1,7 @@
+import FlowIDTableStaking from 0xIDENTITYTABLEADDRESS
+
+// This script gets all the info about a node and returns it
+
+pub fun main(nodeID: String): FlowIDTableStaking.NodeInfo {
+    return FlowIDTableStaking.NodeInfo(nodeID: nodeID)
+}
```
MaxStalker, line 7: This is getting repetitive after a couple of first scripts 😅 
How about we do it this way? 🤔

```suggestion
pub fun main(nodeID: String, _ key: String?): AnyStruct {
    let nodeInfo = FlowIDTableStaking.NodeInfo(nodeID: nodeID)
    
    if key == "initialWeight" { return nodeInfo.initialWeight } 
    if key == "networkingAddress" { return nodeInfo.networkingAddress } 
    if key == "networkingKey" { return nodeInfo.networkingKey } 
    if key == "tokensRewarded" { return nodeInfo.tokensRewarded } 
    if key == "role" { return nodeInfo.role } 
    if key == "tokensStaked" { return nodeInfo.tokensStaked } 
    if key == "stakingKey" { return nodeInfo.stakingKey } 
    if key == "tokensCommitted" { return nodeInfo.tokensCommitted }
    if key == "tokensUnlocked" { return nodeInfo.tokensUnlocked } 
    if key == "tokensUnstaked" { return nodeInfo.tokensUnstaked } 
    if key == "tokensRequestedToUnstake" { return nodeInfo.tokensRequestedToUnstake } 
    
    // In other cases, return full nodeInfo
    return nodeInfo
}
```

MaxStalker (commented):
Small suggestion about the script. Other than that looks great 👍

transactions/idTableStaking/get_node_info.cdc
joshuahannan: I don't know, I think it would probably be clearer to app developers if they were individual scripts for each field instead of one script for all of them. What do you think, @JeffreyDoyle @psiemens

MaxStalker (approved):

PR #16, opened 2020-09-29
Initial version of Lockbox contract

psiemens:


contracts/Lockbox.cdc
```diff
@@ -0,0 +1,130 @@
...
+
+pub contract Lockbox {
+
+  pub resource TokenBucket {
+    
+    pub var vault: @FlowToken.Vault
+
+    pub var withdrawalLimit: UFix64
+
+    pub fun setWithdrawalLimit(limit: UFix64) {
```
robmyers: Does this open up the contract to a variant of the ERC20Multiple Withdrawal Attack?

https://blockchain-projects.readthedocs.io/multiple_withdrawal.html

If so, and if we do not want to change the interface, the mitigation would be to use two transactions to change the withdrawal limit: one to set the limit to zero, then another to set it to its new value.

PR #18, opened 2020-09-30
Implementation of Flow Arcade Token, scripts and initial tests.

robmyers:
* Code is based on Flow Token. Transaction parameters have been re-ordered in some places to be more regular

* Tests are unstructured and will be improved after initial review.


contracts/FlowArcadeToken.cdc
```diff
@@ -0,0 +1,145 @@
...
+    // Event that is emitted when a new minter resource is created
+    pub event MinterCreated()
+
+    // Event that is emitted when a new burner resource is created
+    //pub event BurnerCreated()
+
+    // Vault
+    //
+    // Each user stores an instance of only the Vault in their storage
+    // The functions in the Vault and governed by the pre and post conditions
```
10thfloor, line 33:
```suggestion
    // The functions in the Vault are governed by the pre and post conditions
```

contracts/FlowArcadeToken.cdc
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    // createEmptyVault
+    //
+    // Function that creates a new Vault with a balance of zero
+    // and returns it to the calling context. A user must call this function
+    // and store the returned Vault in their storage in order to allow their
+    // account to be able to receive deposits of this token type.
+    //
+    pub fun createEmptyVault(): @FungibleToken.Vault {
```
10thfloor, line 94: Should return type be `@FlowArcadeToken.Vault`?

turbolent:
@robmyers 

> Transaction parameters have been re-ordered in some places to be more regular

Can you elaborate a bit, it's hard to see in the PR how it differs.
For consistency reasons I'd suggest keeping them the same across different contracts to avoid developer friction: when preparing a transaction the arguments for it have to be added in the correct order – if there's a difference, this might lead to confusion, bugs, etc.

transactions/flowArcadeToken/admin/add_minter.cdc
```diff
@@ -0,0 +1,22 @@
...
+        // Create a new minter and a private link to it
+        let minter: @FlowArcadeToken.Minter <- tokenAdmin.createNewMinter()
+        adminAcct.save<@FlowArcadeToken.Minter>(<- minter, to: /storage/flowArcadeTokenMinter)
+        let minterCap: Capability<&FlowArcadeToken.Minter> = adminAcct.link<&FlowArcadeToken.Minter>(
+            /private/flowArcadeTokenMinter,
+            target: /storage/flowArcadeTokenMinter
+        )!
+
+        // Transfer the capability to the minter account
+        minterAcct.save<Capability<&FlowArcadeToken.Minter> >(minterCap, to: /storage/flowArcadeTokenMinter)
```
turbolent, line 19: Maybe consider using type inference to make the code easier to read:

```suggestion
        // Create a reference to the admin admin resource in storage
       let tokenAdmin = adminAcct.borrow<&FlowArcadeToken.Administrator>(from: /storage/flowArcadeTokenAdmin)
            ?? panic("Could not borrow a reference to the admin resource")

        // Create a new minter and a private link to it
        let minter <- tokenAdmin.createNewMinter()
        adminAcct.save(<- minter, to: /storage/flowArcadeTokenMinter)
        let minterCap = adminAcct.link<&FlowArcadeToken.Minter>(
            /private/flowArcadeTokenMinter,
            target: /storage/flowArcadeTokenMinter
        )!

        // Transfer the capability to the minter account
        minterAcct.save(minterCap, to: /storage/flowArcadeTokenMinter)
```

contracts/FlowArcadeToken.cdc
robmyers: Assuming FlowToken, which returns the same type for the same function, is correct this is the proper type.

contracts/FlowArcadeToken.cdc
robmyers: Fixed. Thank you!

transactions/flowArcadeToken/admin/add_minter.cdc
robmyers: Yes that is much clearer, thank you! The verbose style was for my own learning.

contracts/FlowArcadeToken.cdc
turbolent: I think @10thfloor point was to change the return type to the concrete type. This was originally not possible in earlier versions of Cadence, but is supported now.

contracts/FlowArcadeToken.cdc
```diff
@@ -0,0 +1,145 @@
...
+    pub event TokensMinted(amount: UFix64)
+
+    // Event that is emitted when tokens are destroyed
+    //pub event TokensBurned(amount: UFix64)
+
+    // Event that is emitted when a new minter resource is created
+    pub event MinterCreated()
+
+    // Event that is emitted when a new burner resource is created
+    //pub event BurnerCreated()
```
turbolent, line 28: Why is this commented out?

transactions/flowArcadeToken/setup_account.cdc
```diff
@@ -0,0 +1,31 @@
...
+            )
+
+            // Create a public capability to the Vault that only exposes
+            // the balance field through the Balance interface
+            signer.link<&FlowArcadeToken.Vault{FungibleToken.Balance}>(
+                /public/flowArcadeTokenBalance,
+                target: /storage/flowArcadeTokenVault
+            )
+        }
+    }
```
turbolent, line 30: Consider using `check` instead of `borrow`, and maybe consider returning early to avoid deep nesting:
```suggestion
        if signer.check<&FlowArcadeToken.Vault>(from: /storage/flowArcadeTokenVault) {
            return
        }
        
        // Create a new FlowArcadeToken Vault and put it in storage
        signer.save(<-FlowArcadeToken.createEmptyVault(), to: /storage/flowArcadeTokenVault)

        // Create a public capability to the Vault that only exposes
        // the deposit function through the Receiver interface
        signer.link<&FlowArcadeToken.Vault{FungibleToken.Receiver}>(
            /public/flowArcadeTokenReceiver,
            target: /storage/flowArcadeTokenVault
        )

        // Create a public capability to the Vault that only exposes
        // the balance field through the Balance interface
        signer.link<&FlowArcadeToken.Vault{FungibleToken.Balance}>(
            /public/flowArcadeTokenBalance,
            target: /storage/flowArcadeTokenVault
        )
    }
}
```

transactions/flowArcadeToken/mint_tokens.cdc
```diff
@@ -0,0 +1,24 @@
...
+
+    prepare(minterAccount: AuthAccount) {
+        self.tokenMinter = minterAccount
+        .borrow<&FlowArcadeToken.Minter>(from: /storage/flowArcadeTokenMinter) 
+        ?? panic("No minter available")
+
+        self.tokenReceiver = getAccount(recipientAddress)
+        .getCapability(/public/flowArcadeTokenReceiver)!
+        .borrow<&{FungibleToken.Receiver}>()
+        ?? panic("Unable to borrow receiver reference")
```
turbolent, line 16: Minor: maybe consider indenting this to make it easier to read:

```suggestion
        self.tokenMinter = minterAccount
            .borrow<&FlowArcadeToken.Minter>(from: /storage/flowArcadeTokenMinter) 
            ?? panic("No minter available")

        self.tokenReceiver = getAccount(recipientAddress)
            .getCapability(/public/flowArcadeTokenReceiver)!
            .borrow<&{FungibleToken.Receiver}>()
            ?? panic("Unable to borrow receiver reference")
```

robmyers:
> @robmyers
> 
> > Transaction parameters have been re-ordered in some places to be more regular
> 
> Can you elaborate a bit, it's hard to see in the PR how it differs.
> For consistency reasons I'd suggest keeping them the same across different contracts to avoid developer friction: when preparing a transaction the arguments for it have to be added in the correct order – if there's a difference, this might lead to confusion, bugs, etc.

The parameters in transactions/flowToken/transfer_tokens.cdc are reversed relative to transactions/flowToken/mint_tokens.cdc . My own preference for parameters is "objects first then simple types, large to small, subject to others" and mint_tokens.cdc's `recipientAddress: Address, amount: UFix64` fits that better so I modified  transactions/flowArcadeToken/transfer_tokens.cdc to fit that instead.

I've reverted it so that the transactions have the same interface for each token.

Thank you!

contracts/FlowArcadeToken.cdc
robmyers: We do not currently support burning.

turbolent:
@robmyers Ah, thanks for explaining! Yeah, ideally it would be consistent even in the FT code itself. Maybe we can still fix it there @joshuahannan?

contracts/FlowArcadeToken.cdc
turbolent: Then I'd recommend removing it

contracts/FlowArcadeToken.cdc
robmyers: This would result in different behaviour from FlowToken. I'm happier returning the concrete type if it is OK to do so. :-)

contracts/FlowArcadeToken.cdc
robmyers: Done! Thank you.

contracts/FlowArcadeToken.cdc
robmyers: I've changed it to use the concrete type. Thank you!

lib/go/templates/flow_arcade_token_templates.go
```diff
@@ -0,0 +1,81 @@
...
+
+	getBalanceFilename = "flowArcadeToken/scripts/get_balance.cdc"
+	getSupplyFilename  = "flowArcadeToken/scripts/get_supply.cdc"
+
+	defaultFungibleTokenAddress   = "FUNGIBLETOKENADDRESS"
+	defaultFlowArcadeTokenAddress = "ARCADETOKENADDRESS"
+)
+
+// ReplaceAddresses replaces the contract addresses in the code
+func replaceAddresses(code, ftAddr, fatAddr string) string {
```
joshuahannan, line 24:
```suggestion
func replaceFATAddress(code, fatAddr string) string {
```

we already have the `replaceAddresses` function that replaces fungible token and flow token addresses, so we should probably rename this and make it only replace the FAT address

joshuahannan: And change all the templates in the file to call both replace functions. Do you think that sounds ok?

transactions/flowArcadeToken/admin/add_minter.cdc
```diff
@@ -0,0 +1,22 @@
+import FlowArcadeToken from 0xARCADETOKENADDRESS
+
+transaction() {
+    
+    prepare(adminAcct: AuthAccount, minterAcct: AuthAccount) {
+        // Create a reference to the admin admin resource in storage
```
joshuahannan, line 6:
```suggestion
        // Create a reference to the admin resource in storage
```

lib/go/test/flow_arcade_token_test.go
```diff
@@ -0,0 +1,559 @@
...
+			adminSigner,
+			oneAddress,
+			oneSigner,
+			twoAddress,
+			"9.09",
+			true,
+		)
+	})
+
+	t.Run("Account with insufficient tokens should not be able to transfer a karger amount", func(t *testing.T) {
```
joshuahannan, line 382:
```suggestion
	t.Run("Account with insufficient tokens should not be able to transfer a larger amount", func(t *testing.T) {
```

joshuahannan (approved):
This looks really good! I just had a few comments.

I like what you're doing with making the tests more modular. We definitely should have a conversation soon about how we can standardize that.

lib/go/templates/flow_arcade_token_templates.go
robmyers: Yes that makes sense. I'll make that change

transactions/flowArcadeToken/admin/add_minter.cdc
robmyers: Thank you for catching this!

contracts/FlowArcadeToken.cdc
```diff
@@ -0,0 +1,145 @@
+import FungibleToken from 0xFUNGIBLETOKENADDRESS
+
+
+pub contract FlowArcadeToken: FungibleToken {
+
+    // Total supply of Flow tokens in existence
```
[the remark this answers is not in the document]
robmyers: Thank you for catching this!

robmyers:
I recommend rebuilding the contract and transaction resources after merging this.

robmyers:
DO NOT MERGE.

There is a critical issue with the minter and the tests for it. Fixing now...

joshuahannan:
What was the issue @robmyers ?

robmyers:
> What was the issue @robmyers ?

@joshuahannan The FAT implementation of minting, in which a Capability for minting is deposited into the minter account's storage, is naive and attempts to use a capability from one account directly in another from /storage/. This always returns null for the capability.

This was concealed because both the resource and the capability have the same path in their respective accounts' storage and in the tests the minter was the admin due to the wrong account details being used to mint tokens. So it could borrow the resource (rather than the capability) and mint.

Using the minter's account details to mint with reveals the problem. This is very basic Cadence but I made this error at precisely the wrong point in my personal learning curve.

I think I need to wrap the minter capability in another resource and deposit *that* in the minter's storage. This wrapping of a capability in a resource given to another account will be like TokenForwarding.cdc in the FT.

The reason for the gymnastics around minting is that the capacity of a FAT minter is unlimited, so I feel they must be unilaterally revokable by us.

robmyers:
> DO NOT MERGE.
> 
> There is a critical issue with the minter and the tests for it. Fixing now...

Fixed. The Minter capability now works properly, with just one extra layer of indirection as per the programming axiom.

PR #21, opened 2020-10-01
Adds Lockbox and StakingProxy

joshuahannan:
* Adds lockbox contract, which manages users' locked tokens they purchased in the token sale
* Adds transactions for lockbox contract:
    * for the admin to create both user accounts (the shared account and the unlocked account)
    * admin deposit locked tokens and increase unlock limit
    * user to withdraw and deposit locked tokens
    * user to perform all staking and delegating actions with locked tokens

* Adds staking proxy contract, which allows node operators to set up their account to be in a staking helper relationship, by allowing a user to read their node info, use it to sign up for a node staker in the staking contract, and then hand a staking reference to the node operator. This can be used with locked tokens or with unlocked tokens.
* Adds transactions for staking proxy contract:
    * transactions for the node operator to manage their account
    * transaction for a user to register as a node and give the "reference" to the node operator
    * transactions for the node operator to perform staking actions

* Adds unit tests for all of the contracts and transactions


transactions/lockbox/administrator/createSharedAccount.cdc
```diff
@@ -0,0 +1,63 @@
...
+transaction(adminPublicKey: [UInt8], userPublicKey: [UInt8])  {
+    prepare(admin: AuthAccount) {
+        let sharedAccount = AuthAccount(payer: admin)
+        let userAccount = AuthAccount(payer: admin)
+
+        sharedAccount.addPublicKey(adminPublicKey)
+        sharedAccount.addPublicKey(userPublicKey)
+
+        // TODO: add separate full-weight key for user
+        userAccount.addPublicKey(userPublicKey)
```
psiemens, line 15: This is what we wrote up in the call yesterday, removes the TODO in there

```
transaction(
  partialAdminPublicKey: [UInt8], // Weight: 100
  partialUserPublicKey: [UInt8], // Weight: 900
  fullUserPublicKey: [UInt8], // Weight: 1000
) {
  prepare(admin: AuthAccount) {
    let sharedAccount = AuthAccount(payer: admin)
    let userAccount = AuthAccount(payer: admin)

    sharedAccount.addPublicKey(partialAdminPublicKey)
    sharedAccount.addPublicKey(partialUserPublicKey)

    userAccount.addPublicKey(fullUserPublicKey)
```

contracts/StakingProxy.cdc
```diff
@@ -0,0 +1,121 @@
+// This contract defines an interface for node stakers
+// to use to be able to perform common staking actions
+
+// It also defines a resource that a node operator can
+// use to store staking proxies for all of their node operation
+// relationships 
+
+pub contract StakingProxy {
+
+    pub struct interface NodeDelegatorProxy {
```
psiemens, line 10: Minor, but maybe move this right below `NodeStakerProxy`. It's sort of a secondary interface and probably doesn't need to be at the top of the contract

contracts/StakingProxy.cdc
```diff
@@ -0,0 +1,121 @@
...
+
+        pub fun borrowStakingProxy(nodeID: String): AnyStruct{NodeStakerProxy} {
+            return self.stakingProxies[nodeID]!
+        }
+    }
+
+    pub fun createProxyHolder(): @NodeStakerProxyHolder {
+        return <- create NodeStakerProxyHolder()
+    }
+}
```
psiemens, line 121: Nice, this file is looking great 👍

contracts/Lockbox.cdc
```diff
@@ -0,0 +1,329 @@
...
+    /// in the admin account
+    pub let LockedTokenAdminCollectionPath: Path
+
+    /// path to store the token holder resource
+    /// in the unlocked account
+    pub let TokenHolderStoragePath: Path
+
+    /// path to store the node operator struct
+    /// in the node operators account for staking helper
+    pub let NodeOperatorCapabilityStoragePath: Path
```
psiemens, line 29: I wonder if this should be defined here?

contracts/Lockbox.cdc
```diff
@@ -0,0 +1,329 @@
...
+    /// stored in the shared account 
+    pub resource LockedTokenManager: FungibleToken.Receiver, FungibleToken.Provider, TokenAdmin {
+    
+        pub var vault: Capability<&FlowToken.Vault>
+
+        pub var nodeStaker: @FlowIDTableStaking.NodeStaker?
+
+        pub var nodeDelegator: @FlowIDTableStaking.NodeDelegator?
+
+        pub var unlockLimit: UFix64
```
psiemens, line 45: Sorry to be pedantic, but I would probably place this underneath `vault` to signal coupling between the two variables.

contracts/Lockbox.cdc
```diff
@@ -0,0 +1,329 @@
...
+        // Lockbox.TokenAdmin actions
+
+        pub fun increaseUnlockLimit(delta: UFix64) {  
+            self.unlockLimit = self.unlockLimit + delta
+        }
+
+        // Lockbox.TokenHolder actions
+
+        pub fun registerNode(id: String, role: UInt8, networkingAddr: String, networkingKey: String, stakingKey: String) {
+            self.nodeStaker <- FlowIDTableStaking.addNodeRecord(self.nodeInfo)
```
psiemens, line 95: I guess `self.nodeInfo` should be replaced here?

contracts/Lockbox.cdc
```diff
@@ -0,0 +1,329 @@
...
+        }
+
+        pub fun registerDelegator(nodeID: String) {
+            self.nodeDelegator <- FlowIDTableStaking.registerDelegator(nodeID: nodeID)
+        }
+
+        pub fun createTokenHolder(tokenManager: Capability<&LockedTokenManager>): @TokenHolder {
+            return <- create TokenHolder(tokenManager: tokenManager)
+        }
```
psiemens, line 104: Can this just be defined directly on the `Lockbox` contract?

contracts/Lockbox.cdc
```diff
@@ -0,0 +1,329 @@
...
+
+        access(self) var tokenManager: Capability<&LockedTokenManager>
+
+        access(self) var nodeStakerProxy: LockedNodeStakerProxy?
+
+        access(self) var nodeDelegatorProxy: LockedNodeDelegatorProxy?
+
+        init(tokenManager: Capability<&LockedTokenManager>) {
+            pre {
+                tokenManager.borrow() != nil: "Invalid capability"
```
psiemens, line 118:
```suggestion
                tokenManager.borrow() != nil: "Must pass a LockedTokenManager capability"
```

contracts/Lockbox.cdc
```diff
@@ -0,0 +1,329 @@
...
+
+        // FungibleToken.Provider actions
+
+        pub fun withdraw(amount: UFix64): @FungibleToken.Vault {
+            let tokenManagerRef = self.tokenManager.borrow()!
+
+            return <- tokenManagerRef.withdraw(amount: amount)
+        }
+
+        pub fun createNodeStaker(id: String, role: UInt8, networkingAddr: String, networkingKey: String, stakingKey: String) {
```
psiemens, line 139: Hmm, should `role` be part of `StakingProxy.NodeInfo`? And then maybe these params can just be replaced with a single `nodeInfo` param?

```suggestion
        pub fun createNodeStaker(id: String, role: UInt8, networkingAddr: String, networkingKey: String, stakingKey: String) {
```

psiemens (commented):
Nice!

contracts/Lockbox.cdc
joshuahannan: Probably not. I'll define it in the StakingProxy contract

contracts/Lockbox.cdc
joshuahannan: Good suggestion! 👍

contracts/Lockbox.cdc
joshuahannan: yep, I missed that. Good catch!

contracts/Lockbox.cdc
joshuahannan: yeah I guess it could. I'll do that

contracts/Lockbox.cdc
joshuahannan: Yeah you're right. I'll replace those

contracts/Lockbox.cdc
```diff
@@ -0,0 +1,431 @@
...
+            // register node, which stores the NodeStaker object in the LockedTokenManager
+            tokenManagerRef.registerNode(nodeInfo: nodeInfo, amount: amount)
+
+            // Create a new staker proxy that can be accessed in transactions
+            self.nodeStakerProxy = LockedNodeStakerProxy(tokenManager: self.tokenManager)
+        }
+
+        /// The user calls this function if they want to register as a node operator
+        /// They have to provide all the info for their node
+        pub fun createNodeDelagtor(nodeID: String) {
```
robmyers, line 202: Typo: Delagtor -> Delegator

contracts/Lockbox.cdc
```diff
@@ -0,0 +1,468 @@
...
+
+            // register node, which stores the NodeStaker object in the LockedTokenManager
+            tokenManagerRef.registerNode(nodeInfo: nodeInfo, amount: amount)
+
+            // Create a new staker proxy that can be accessed in transactions
+            self.nodeStakerProxy = LockedNodeStakerProxy(tokenManager: self.tokenManager)
+        }
+
+        /// The user calls this function if they want to register as a node operator
+        /// They have to provide all the info for their node
```
robmyers, line 232:
```suggestion
        /// They have to provide the node ID for their node
```

robmyers (changes requested):

contracts/Lockbox.cdc
```diff
@@ -0,0 +1,468 @@
+/*
+
+    Lockbox implements the functionality required to manage FLOW
+    buyers locked tokens from the token sale.
+
+    Each token holer gets two accounts. One account is their locked token
+    account. It will be jointly controlled by the user and Dapper Labs.
```
psiemens, line 7: Thinking about this again, it might be better to say "token administrator" rather than "Dapper Labs" -- eventually this will be controlled via governance so probably better to use a generic identifier.

contracts/Lockbox.cdc
```diff
@@ -0,0 +1,468 @@
...
+    No actions will be able to be performed with this account without authorization
+    from Dapper Labs. Locked tokens are stored in this account and Dapper Labs
```
psiemens, line 9:
```suggestion
    No actions can be performed by this account without authorization
    from both the token holder and the token administrator. 
```

The current phrasing sort of implies that the token admin can move funds without approval from the token holder -- might be better to be extra explicit here.

contracts/Lockbox.cdc
```diff
@@ -0,0 +1,468 @@
...
+    with authorize withdrawals every time a milestone is passed in the
+    vesting period.
```
psiemens, line 11: Similar to the above, I think we should be very clear about what the token admin can do. Something like this maybe?

```
The token administrator must co-sign the transfer of any locked tokens. The token admin cannot interact with the account without approval from the token holder, except to deposit additional locked FLOW or to unlock existing FLOW at each milestone in the token vesting period.

File: contracts/Lockbox.cdc [hunk 15]
```diff
@@ -0,0 +1,468 @@
...
+        
+    The second account is the unlocked user account. This account is
+    in full possesion and control of the user and they can do whatever
+    they want with it. This account will store a capability that allows
+    them to withdraw tokens when they become unlocked and also to
+    perform staking operations with their locked tokens.
+
+    When a user account is created, both accounts are initialized with
+    their respective objects, LockedTokenManager for the shared account,
```
psiemens, line 20:
```suggestion
    their respective objects: LockedTokenManager for the shared account,
```

contracts/Lockbox.cdc
```diff
@@ -0,0 +1,468 @@
...
+
+    /// path to store the token holder resource
+    /// in the unlocked account
+    pub let TokenHolderStoragePath: Path
+
+    /// Public path to store the capability that allows
+    /// reading the unlock limit of an account
+    pub let UnlockLimitPublicPath: Path
+
+    /// stored in Admin account to use to increase the unlock
```
psiemens, line 62:
```suggestion
    /// The TokenAdmin capability allows the token administrator to unlock tokens at each
    /// milestone in the vesting period.
```

contracts/Lockbox.cdc
```diff
@@ -0,0 +1,468 @@
...
+    /// token limit every time a vesting release happens
+    pub resource interface TokenAdmin {
+        pub fun increaseUnlockLimit(delta: UFix64)
+    }
+
+    /// stored in the shared account to manage access to the locked token vault
+    /// and to the staking/delegating resources
+    pub resource LockedTokenManager: FungibleToken.Receiver, FungibleToken.Provider, TokenAdmin {
```
psiemens, line 70: Do you think it would be more intuitive to flip the naming around here?

We could call the contract `LockedTokens` and then rename this resource to `Lockbox`. I feel like "lockbox" more accurately describes the actual resource that holds the tokens, and `LockedTokenManager` also feels a bit generic.

contracts/Lockbox.cdc
```diff
@@ -0,0 +1,468 @@
...
+    
+        /// Capability to the normal Flow Token Vault in the shared account
```
psiemens, line 72: I feel like this field could use a bit more explanation 

```
// This is a reference to the default FLOW vault stored in the shared account. 
//
// All locked FLOW tokens are stored in this vault, which can be accessed in two ways:
//   1) Directly, in a transaction co-signed by both the token holder and token administrator
//   2) Indirectly via the LockedTokenManager, in a transaction signed by the token holder

File: contracts/Lockbox.cdc [hunk 19]
```diff
@@ -0,0 +1,468 @@
...
+        }
+
+        destroy () {
+            destroy self.nodeStaker
+            destroy self.nodeDelegator
+        }
+
+        // FungibleToken.Receiver actions
+
+        pub fun deposit(from: @FungibleToken.Vault) {
```
psiemens, line 101: Unfortunately I don't think it's very clear that `deposit` and `withdraw` are for moving **unlocked** FLOW in and out of the lockbox.

Maybe we could make this a passthrough to a more explicitly-named function?

```
pub fun deposit(from: @FungibleToken.Vault) {
  self.depositUnlockedTokens(from: from)
}
```

contracts/Lockbox.cdc
```diff
@@ -0,0 +1,468 @@
...
+            let balance = from.balance
+
+            vaultRef.deposit(from: <- from)
+
+            self.increaseUnlockLimit(delta: balance)
+        }
+
+        // FungibleToken.Provider actions
+
+        pub fun withdraw(amount: UFix64): @FungibleToken.Vault {
```
psiemens, line 113: Same here:

```
pub fun withdraw(amount: UFix64): @FungibleToken.Vault {
  self.withdrawUnlockedTokens(amount: amount)
}
```

contracts/Lockbox.cdc
```diff
@@ -0,0 +1,468 @@
...
+        /// Registers a new Delegator with the Flow Staking contract
+        /// the caller has to specify the ID of the node operator
+        /// they are delegating to
+        pub fun registerDelegator(nodeID: String) {
+            self.nodeDelegator <-! FlowIDTableStaking.registerNewDelegator(nodeID: nodeID)
+        }
+    }
+
+    /// allows anyone to read the unlock limit of an account with TokenHolder
+    pub resource interface UnlockLimit {
```
psiemens, line 163: 👍

contracts/Lockbox.cdc
```diff
@@ -0,0 +1,468 @@
...
+            pre {
+                tokenManager.borrow() != nil: "Must pass a LockedTokenManager capability"
+            }
+            self.tokenManager = tokenManager
+            self.nodeStakerProxy = nil
+            self.nodeDelegatorProxy = nil
+        }
+
+        pub fun getUnlockLimit(): UFix64 {
+            let tokenManagerRef = self.tokenManager.borrow()!
```
psiemens, line 192: Could be nice to factor this common line out into a helper:

```
pub fun borrowTokenManager(): &LockedTokenManager {
  return self.tokenManager.borrow()!
}

pub fun getUnlockLimit(): UFix64 {
  return self.borrowTokenManager().unlockLimit
}

pub fun deposit(from: @FungibleToken.Vault) {
  self.borrowTokenManager().deposit(from: <-from)
}

// etc...
```

contracts/Lockbox.cdc
```diff
@@ -0,0 +1,468 @@
...
+
+            return tokenManagerRef.unlockLimit
+        }
+
+        /// Deposits tokens in the locked vault, which marks them as 
+        /// unlocked and available to withdraw
+        pub fun deposit(from: @FungibleToken.Vault) {
```
psiemens, line 199: Should `TokenHolder` also implement `FungibleToken.Receiver` and `FungibleToken.Provider`? It might be nice to allow the user to route standard token transfers to the lockbox. The methods are already defined -- we'd just need to add the conformances to this resource defintion.

contracts/StakingProxy.cdc
```diff
@@ -0,0 +1,156 @@
...
+        
+        pub fun addStakingProxy(nodeID: String, proxy: AnyStruct{NodeStakerProxy})
+
+        pub fun getNodeInfo(nodeID: String): NodeInfo?
+    }
+
+    /// The resource that node operators store in their accounts
+    /// to manage relationships with token holders who pay them off-chain
+    /// instead of with tokens
+    pub resource NodeStakerProxyHolder: NodeStakerProxyHolderPublic {
```
psiemens, line 85:
```suggestion
    pub resource NodeStakerProxyContainer: NodeStakerProxyHolderPublic {
```

Maybe this should be "container" to be consistent with other resources and to avoid confusion with the _token_ holder?

psiemens (commented):
Contracts are looking great! I'll go through the rest of the transactions tomorrow morning

contracts/Lockbox.cdc
joshuahannan: Good suggestion. I changed it to token administrator

contracts/Lockbox.cdc
joshuahannan: yeah that is a good idea

contracts/Lockbox.cdc
joshuahannan: yeah, I've been thinking something similar to this, but hadn't had the courage to do it, but I think I will now that you've mentioned it

psiemens: power to you!

contracts/LockedTokens.cdc
```diff
@@ -0,0 +1,476 @@
...
+    }
+
+    /// Public function to create a new token admin collection
+    pub fun createTokenAdminCollection(): @TokenAdminCollection {
+        return <- create TokenAdminCollection()
+    }
+
+    /// Public function to create a new Locked Token Manager
+    /// every time a new user account is created
+    pub fun createNewLockedTokenManager(vault: Capability<&FlowToken.Vault>): @LockedTokenManager {
```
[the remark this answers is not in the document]
joshuahannan: yep, I'll remove the `New`

contracts/testContracts/TestFlowIDTableStaking.cdc
```diff
@@ -0,0 +1,188 @@
+/*
+
+    TestFlowIDTableStaking
```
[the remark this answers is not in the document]
joshuahannan: There are a bunch of other setup tasks, ending staking, unstaking, and paying rewards that need to be done with the real staking contract before accounts can call all of the functions and I didn't want to spend the time adding all the setup tests for that since it has already been tested, so I just made the test one to have all the same function signatures and returns so I can just test the interface. Is that ok?

transactions/stakingProxy/sh_get_node_info.cdc
```diff
@@ -0,0 +1,15 @@
...
+    let nodeAccount = getAccount(account)
+
+    let proxyCapability = nodeAccount.getCapability
+        <&StakingProxy.NodeStakerProxyHolder{StakingProxy.NodeStakerProxyHolderPublic}>
+        (StakingProxy.NodeOperatorCapabilityPublicPath)!
+
+    let proxyRef = proxyCapability.borrow()
+        ?? panic("Could not borrow public reference to staking proxy")
+
+    return proxyRef.getNodeInfo()!
```
psiemens, line 14: This actually needs to take the `nodeID` as an argument. Probably best to change the function to:

```
pub fun main(account: Address, nodeID: String): StakingProxy.NodeInfo {

robmyers: Good catch! Fixed in https://github.com/onflow/flow-core-contracts/pull/29

joshuahannan: oops. Yeah I didn't test this one

psiemens (approved):
I've gone through all of the contracts, transactions and tests. I'm ready to give this a 👍 !

psiemens (approved):
LGTM!

robmyers (approved):
Looking good! All tests pass here.

PR #25, opened 2020-10-02
rob/lockedtokens staking proxy templates

robmyers:
Adds staking proxy transactions and GenerateXXXXXScript methods.

Test simply checks that the GenerateXXXXXScript methods can find their resources.


joshuahannan:
Thank you Rob! We won't be using these though, so I'll close this PR.

PR #26, opened 2020-10-02
Implement the lockbox staking templates (as I was meant to).

robmyers:
Includes the contracts, GenerateXXXScript functions, and a minimal test to make sure those work.


transactions/lockbox/staker/request_unstaking.cdc
```diff
@@ -0,0 +1,18 @@
+import Lockbox from 0xf3fcd2c1a78f5eee
+import StakingProxy from 0x179b6b1cb6755e31
+
+transaction(amount: UFix64) {
+
+    let holderRef: &LockBox.TokenHolder
+
+    prepare(acct: AuthAccount) {
+        self.holderRef = acct.borrow<&LockBox.TokenHolder>(from: LockBox.TokenHolderStoragePath)
```
joshuahannan, line 9:
```suggestion
        self.holderRef = acct.borrow<&LockBox.TokenHolder>(from: LockBox.TokenHolderStoragePath)
            ?? panic("Could not borrow a reference to the TokenHolder object")
```
Just so you are aware, we need to unwrap the optional on all of these. You don't need to worry about fixing it yourself though! I'll merge this and fix it in my local branch.

joshuahannan: I missed it in my sample transaction so I apologize if that misled you!

joshuahannan (approved):

PR #28, opened 2020-10-05
rename unstaked to unstaking and unlocked to unstaked

joshuahannan:
only in the core staking contract and its transactions


contracts/FlowIDTableStaking.cdc
```diff
@@ -420,7 +420,7 @@ pub contract FlowIDTableStaking {
 
         /// Request amount tokens to be removed from staking
         /// at the end of the next epoch
-        pub fun requestUnStaking(amount: UFix64) {
+        pub fun requestUnstaking(amount: UFix64) {
```
psiemens, line 423: 👍

psiemens (approved):
Nice! This reads a lot clearer IMO

PR #29, opened 2020-10-06
Provide nodeId to sh_get_node_info and use it. Thanks @psiemens .

robmyers:
Fix for https://github.com/onflow/flow-core-contracts/pull/21/files#r499947276


joshuahannan:
I fixed this in my branch already, but thank you for thinking of it!

PR #33, opened 2020-10-07
Update all contracts and templates to use placeholder addresses

psiemens:
Screen Shot 2020-10-06 at 11 49 56 PM


Kay-Zee (approved):
Nice! Noticed some inconsistencies when i was doing the testnet tests as well. Looking clean!

PR #34, opened 2020-10-07
adding events and making create admin collection not public

joshuahannan:


contracts/LockedTokens.cdc
```diff
@@ -549,5 +560,8 @@ pub contract LockedTokens {
 
         self.LockedAccountCreatorStoragePath = /storage/lockedAccountCreator
         self.LockedAccountCreatorPublicPath = /public/lockedAccountCreator
+
+        /// create a single admin collection and store it
+        self.account.save(<-create TokenAdminCollection(), to: self.LockedTokenAdminCollectionStoragePath)
```
psiemens, line 565: What do you think about using the same pattern from `FlowToken`, where a separate administrator is passed into `init`? https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowToken.cdc#L168-L197

This would allow us to make the service account the token admin, which I think is what we want. The token admin needs to have access to the main token supply, which the service account already has.

contracts/LockedTokens.cdc
```diff
@@ -166,6 +173,7 @@ pub contract LockedTokens {
         /// Called by the admin every time a vesting release happens
         pub fun increaseUnlockLimit(delta: UFix64) {  
             self.unlockLimit = self.unlockLimit + delta
+            emit UnlockLimitIncreased(address: self.owner!.address, amount: delta)
```
robmyers, line 176: Is it worth adding a field to represent the new limit amount? This would make tracking the history of an account's limit easier as it can be established without keeping a running total.

I suggest this rather than changing the event to simply reflect the new level as the amount information is obviously also useful, and this naming reflects the function intention well.

joshuahannan: yeah I can add that field also

contracts/LockedTokens.cdc
joshuahannan: yeah I can do that. Do you know where the transaction template is that we use to deploy that? I'd rather reuse that then have to write a new one

psiemens: Here's how it's done for the `FlowToken`: https://github.com/onflow/flow-go/blob/master/fvm/bootstrap.go#L174-L181

contracts/LockedTokens.cdc
```diff
@@ -166,6 +173,7 @@ pub contract LockedTokens {
         /// Called by the admin every time a vesting release happens
         pub fun increaseUnlockLimit(delta: UFix64) {  
             self.unlockLimit = self.unlockLimit + delta
+            emit UnlockLimitIncreased(address: self.owner!.address, increaseAmount: delta, newLimit: self.unlockLimit)
```
turbolent, line 176: Just checking: Is the resource not having an owner an unrecoverable error?
If not, consider using optional chaining instead.

The same applies to the two lines below which emit an event with the address of the resource owner

joshuahannan:
What is the status of this @psiemens? Is there anything else that needs to be done with this to get it merged?

psiemens:
@joshuahannan Thanks for checking in! I'm just working on updating the tests to use the new deployment transaction. I'll try to have it ready for review by tomorrow.

The custodians have been made aware of this change and know that it's non-breaking, so we can take a day to make sure it's extra solid.

psiemens:
@joshuahannan This PR should be good to go now. I updated the deployment of `LockedTokens` to use a separate account as the admin, and also added a line to link the capability for the `TokenAdminCollection` resource.

joshuahannan (commented):
These changes look good @psiemens! Can you approve it so we can merge?

psiemens (approved):

PR #39, opened 2020-10-14
Restructure Makefiles for easier testing

psiemens:
This PR adds a root `Makefile` with a `test` target. The Go tests can now be executed with `make test` directly from the root of the repository. 

I also removed the `sh_` prefix from the `StakingProxy` transactions since "staking helper" is no longer a term we use.


joshuahannan (approved):
thanks for cleaning this up!

PR #47, opened 2020-10-21
registration check transaction

joshuahannan:
Transactions for the token admin to check that an address is registered correctly

* One transaction checks the shared account directly
* The other gets the shared account from the main account and checks it


robmyers (approved):
This looks good.

"Is this account properly configured?" feels like something a script should be able to answer, but I understand why we need a tx with the current architecture.

PR #49, opened 2020-10-22
add delegator refund at end staking auction

joshuahannan:
I forgot to include refunding delegators if their node operator isn't included in the next epoch, so I added the refund and added tests for it


robmyers (approved):
Looks good!

I'm starting to look at loops with an eye for when gas fees are enabled and if they go up, but that's not relevant here yet.

psiemens (commented):
Sorry, didn't get this in before merge, but LGTM

PR #50, opened 2020-10-28
Transaction template manifest file

psiemens:
Apologies for not opening this PR sooner -- I cleaned up the transaction templates and created a JSON manifest file to use when putting these templates into the Ledger app.

I didn't change any of the transaction logic, just adjusted the style based on previous feedback (e.g. use `account` rather than `acct`). 

The `templates/manifest` package now contains a script that generates a portable JSON manifest file that acts as a single source of truth for all templates. This manifest is now being used by both the Ledger app and JS SDK, neither of which has the luxury of importing the Go packages directly.

The current configuration creates separate manifest files for testnet and mainnet:
- `manifest.testnet.json`
- `manifest.mainnet.json`


joshuahannan (approved):
This all looks good! We need to figure out how to generalize this to all the other contract repos. Maybe to pull out some of the generic stuff to generate the manifest and templates into their own repo so we can make a general API for this for and language

psiemens:
> Maybe to pull out some of the generic stuff to generate the manifest and templates into their own repo so we can make a general API for this for and language

@joshuahannan Yes! Something like that would be great. Thanks for the review :)

PR #53, opened 2020-10-30
Changes rewards calculation and test

joshuahannan:
* Changes the rewards calculation per user to (total rewards) * (amount staked by the user / total staked by all stakers)


robmyers (approved):
This is also looking good.

I've some thoughts about handling node types more generally but that's out of scope here and is probably over-generalizing things anyway. :-)

contracts/FlowIDTableStaking.cdc
```diff
@@ -719,7 +719,7 @@ pub contract FlowIDTableStaking {
                 if nodeRecord.tokensStaked.balance == 0.0 { continue }
 
                 /// Calculate the amount of tokens that this node operator receives
-                let rewardAmount = rewardsForNodeTypes[nodeRecord.role]! * (nodeRecord.tokensStaked.balance / FlowIDTableStaking.totalTokensStakedByNodeType[nodeRecord.role]!)
+                let rewardAmount = FlowIDTableStaking.epochTokenPayout * (nodeRecord.tokensStaked.balance / totalStaked)
```
psiemens, line 722: `nodeRecord.tokensStaked.balance` does _not_ include any delegated tokens, right?

psiemens (approved):

contracts/FlowIDTableStaking.cdc
joshuahannan: Yes, that is correct. 👌

PR #54, opened 2020-11-03
renamed transactions

joshuahannan:
Renamed all the staking transactions and scripts that didn't follow our existing naming conventions
Reorganized the idTableStaking templates file to make sure the different users had their own sections


psiemens (approved):
LGTM 👍

PR #55, opened 2020-11-03
Epoch Cluster Quorum Certificate Aggregator Contract

jordanschalm:
This PR adds a draft for the Cluster Quorum Certificate Aggregator contract, as well as a Cadence definition for `EpochSetup` and `EpochCommitted` service events. The goal is to hammer out the structure and organization of this contract enough to solidify an interface that clients of the contract will use. A subsequent issue (https://github.com/dapperlabs/flow-go/issues/4529) captures complete implementation and testing.

### DIFF
--- a/contracts/EpochClusterQCs.cdc
+++ b/contracts/EpochClusterQCs.cdc
@@ -179,14 +179,14 @@ pub contract EpochClusterQCs {
     // inserted by the Epoch contract.
     pub resource QCVoterStore: QCVoterReceiver {
         pub let nodeID: String
-        pub var voter: @EpochClusterQCs.QCVoter?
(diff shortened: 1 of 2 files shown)


joshuahannan:
This is really good, but I really think we need to have a big discussion about how all the epoch contracts are going to fit together and make user stories before we start writing much code for the individual contracts. I know from the QC and DKG contracts that I wrote that when we starting thinking about how they relate to the epoch schedule, the main epoch smart contract, and staking, things start getting a lot more complicated. I'm just trying to save you from doing extra work that will probably get changed in the future. I had to refactor my code for the staking contract over and over again because we didn't do proper planning and I would like to minimize that this time. Does that make sense?

PR #61, opened 2020-11-17
First draft of the cluster QC smart contract

joshuahannan:


contracts/FlowIDTableStaking.cdc
```diff
@@ -626,6 +657,51 @@ pub contract FlowIDTableStaking {
     /// and pay rewards to nodes at the end of an epoch
     pub resource Admin {
 
+        /// Any node can call this function to register a new Node
```
jordanschalm, line 660: Is this comment still accurate if this is in the `Admin` now?

contracts/epochs/EpochClusterQCs.cdc
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    // Vote represents a vote from one collection node. It simply contains a string with an
+    // encoded representation of the vote. Votes are aggregated to build quorum certificates;
+    // eventually we may want to do the aggregation and validate votes within the smart
+    // contract, but in the meantime the vote contents are opaque here.
+    pub struct Vote {
+        pub var nodeID: String
+        pub(set) var raw: String?
+        pub(set) var clusterIndex: UInt16
```
jordanschalm, line 70: I don't think this should be settable after instantiation. The `Voter` shouldn't be able to submit votes for a cluster they aren't a part of.

contracts/epochs/EpochClusterQCs.cdc
```diff
@@ -0,0 +1,201 @@
...
+    // ================================================================================
+
+    // Indicates whether votes are currently being collected.
+    // If false, no node operator will be able to submit votes
+    pub var inProgress: Bool
+
+    // The collection node clusters for the current epoch
+    access(account) var clusters: [Cluster]
+
+    // Votes that nodes claim at the beginning of each EpochSetup phase
```
jordanschalm, line 20: Could add a note here that this is node IDs as the keys in the mapping

contracts/epochs/EpochClusterQCs.cdc
```diff
@@ -0,0 +1,201 @@
...
+        // Returns whether this voter has successfully submitted a vote for this epoch.
+        pub fun voted(): Bool {
+            if EpochClusterQCs.generatedVotes[self.nodeID] == nil {
+                return true
+            } else {
+                return false
+            }
+        }
+
+        // Submits the given vote. Can be called only once.
```
jordanschalm, line 100:
```suggestion
        // Submits the given vote. Can be called only once per epoch.
```

contracts/epochs/EpochClusterQCs.cdc
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    // The Voter resource is generated for each collection node after they register.
+    // Each resource instance is good for all future potential epochs, but will
+    // only be valid if the node operator has been confirmed as a collector node for the next epoch.
+    pub resource Voter {
+
+        pub let nodeID: String
+
+        pub var currentVote: Vote?
```
jordanschalm, line 89: Doesn't seem like this field is used

contracts/epochs/EpochClusterQCs.cdc
```diff
@@ -0,0 +1,201 @@
...
+        pub let index: UInt16
+
+        pub fun size(): UInt16 {
+            return UInt16(self.nodeIDs.length) 
+        }
+
+        // Returns the minimum number of votes required in order to be able to generate a
+        // valid quorum certificate for this cluster.
+        pub fun voteThreshold(): UInt16 {
+            return self.size() * UInt16(2) / UInt16(3) + UInt16(1)
```
jordanschalm, line 54: Just a note to what alex wrote up on the precise way to do this and a link to the Go implementation, I forgot to add a TODO: https://www.notion.so/dapperlabs/Epoch-Smart-Contract-Construct-Cluster-QCs-b15a31ee48da4bee93295fe61d853cad#35f63e685c6741c5938b60b3e0257e8c

contracts/epochs/EpochClusterQCs.cdc
```diff
@@ -0,0 +1,201 @@
...
+    // The collection node clusters for the current epoch
+    access(account) var clusters: [Cluster]
+
+    // Votes that nodes claim at the beginning of each EpochSetup phase
+    access(account) var generatedVotes: {String: Vote}
+
+    // Votes submitted per cluster
+    access(account) var votesByCluster: {UInt16: [Vote]}
+
+    // Indicates if a voter resource has already been claimed by a node ID
```
jordanschalm, line 26: Adding a note that this design requires certain properties be true of node IDs, which aren't currently enforced. Need to document and enforce these on the protocol side and ideally have node IDs assigned by the smart contract during the staking process. 

I think we should aim for these properties:
* a node has the same node ID for the duration of their continual participation in the network (until they complete the un-staking process)
* node IDs are never re-used

jordanschalm (commented):
Looking good!

contracts/FlowIDTableStaking.cdc
joshuahannan: Nope, I'll change it

contracts/epochs/EpochClusterQCs.cdc
joshuahannan: yes, you're right. It should be a constant

contracts/epochs/EpochClusterQCs.cdc
joshuahannan: These are currently enforced by the identity table smart contract. 👍

joshuahannan:
Created an epoch feature branch and a new pull request with these changes https://github.com/onflow/flow-core-contracts/pull/66

closing this one

PR #62, opened 2020-11-19
Fix and test access node bugs

joshuahannan:
* Made it so access nodes don't receive any rewards
* Made it so access nodes can't be delegated to
* Made it so staked tokens have to be greater than zero in order to be considered a staked node, instead of greater than the minimum. The contract was mistakenly saying access nodes are currently staked, because their minimum is `0.0`


contracts/FlowIDTableStaking.cdc
```diff
@@ -924,7 +929,9 @@ pub contract FlowIDTableStaking {
         for nodeID in FlowIDTableStaking.getNodeIDs() {
             let nodeRecord = FlowIDTableStaking.borrowNodeRecord(nodeID)
 
-            if nodeRecord.tokensStaked.balance >= self.minimumStakeRequired[nodeRecord.role]!  {
+            // Access nodes do not have a minimum, so this makes sure they are included when they are staked,
+            // but not if they are equal to their minimum, which is zero
+            if nodeRecord.tokensStaked.balance > 0.0 {
```
zhangchiqing, line 934: but why it no longer needs to check against `minimumStakeRequired`?

contracts/FlowIDTableStaking.cdc
```diff
@@ -716,7 +716,7 @@ pub contract FlowIDTableStaking {
 
                 let nodeRecord = FlowIDTableStaking.borrowNodeRecord(nodeID)
 
-                if nodeRecord.tokensStaked.balance == 0.0 { continue }
+                if nodeRecord.tokensStaked.balance == 0.0 || nodeRecord.role == UInt8(5) { continue }
```
zhangchiqing, line 719: this is to say access node will not get reward even if they staked?

contracts/FlowIDTableStaking.cdc
```diff
@@ -868,6 +868,11 @@ pub contract FlowIDTableStaking {
 
         let nodeRecord = FlowIDTableStaking.borrowNodeRecord(nodeID)
 
+        assert (
+            nodeRecord.role != UInt8(5),
+            message: "Cannot register a delegator for an access node"
```
zhangchiqing, line 873: because access node don't get reward?

lib/go/test/flow_idtable_staking_test.go
```diff
@@ -1838,7 +1919,7 @@ func TestIDTable(t *testing.T) {
 
 		tx := flow.NewTransaction().
 			SetScript(templates.GenerateEndStakingScript(env)).
-			SetGasLimit(100).
+			SetGasLimit(10000).
```
zhangchiqing, line 1922: why increasing it?

How do I know what's the proper gas limit to use for a transaction?

contracts/FlowIDTableStaking.cdc
joshuahannan: Because before, it would consider access nodes a staked node even if their tokens staked were zero, because the minimum for access nodes is zero. Now, we can just say `> 0.0` because we know that if a node has any tokens staked, it will definitely be above the minimum, because the `endStaking` function enforces that.

contracts/FlowIDTableStaking.cdc
joshuahannan: Yes, I had forgotten to not allow them to get rewards. Thank you for pointing that out earlier!

contracts/FlowIDTableStaking.cdc
joshuahannan: Yes, I just want to prevent delegators from delegating to a place where they won't get rewards

lib/go/test/flow_idtable_staking_test.go
joshuahannan: I increased the gas limit because my tests had made it so 100 limit wasn't enough. I don't know what the actual amount it uses is, though. I just increased it to an arbitrary number

contracts/FlowIDTableStaking.cdc
robmyers: It's worth making a comment to that effect, as I had the same question.

Given that the idiom elsewhere is to check `nodeRecord.role == UInt8(5)` it might be clearer to do so here as well.

robmyers (approved):
Approved but do see clarity-related comment.

contracts/FlowIDTableStaking.cdc
joshuahannan: So the comment that I already put there isn't sufficient? That is why I included it

robmyers: Sorry, my brain stopped working briefly. Yes that's great.

joshuahannan: I make the check much more clear about what it is checking, and I also clarified it in the comment. Can you let me know if it looks good @robmyers @zhangchiqing ?

zhangchiqing: I wonder if we need to keep the check here consistent 
https://github.com/onflow/flow-core-contracts/pull/62/files#diff-65336be374bb3fc9ad7b822243e065a389d73e758c7c16223e52fc3181cea59bR914

If so, it worth to extract into a function

PR #63, opened 2020-11-19
FEATURE: Epoch contracts

joshuahannan:
This is the feature branch for the epoch smart contract. It will contain all the contracts related to the upcoming Epochs changes to the Flow protocol.

* Epoch Contract
* Refactored Staking Contract
* Cluster Quorum Certificate Contract
* Distributed Key Generation Contract
* Heartbeat contract


contracts/FlowIDTableStaking.cdc
```diff
@@ -56,12 +36,18 @@ pub contract FlowIDTableStaking {
...
+
     /// Holds the identity table for all the nodes in the network.
     /// Includes nodes that aren't actively participating
     /// key = node ID
-    /// value = the record of that node's info, tokens, and delegators
     access(contract) var nodes: @{String: NodeRecord}
 
+    access(contract) var stakingKeyClaimed: {String: Bool}
+    access(contract) var networkingKeyClaimed: {String: Bool}
+    access(contract) var networkingAddressClaimed: {String: Bool}
```
turbolent, line 49: I'd recommend not storing these dictionaries in the contract itself, as all this data will need to be decoded each time a transaction/script imports and uses the `FlowIDTableStaking` contract value.

For now it's better to store this data in the contract owner's account storage in separate storage paths, and maybe add some helper functions to access the data easily.

contracts/epochs/FlowDKG.cdc
```diff
@@ -0,0 +1,314 @@
...
+        // the existing one
+        for key in submission {
+
+            // if a key is different, stop checking this submission
+            // and return false
+            if key != existingSubmission[index] {
+                return false
+            }
+
+            index = index + 1
```
turbolent, line 223: Would be nice to get the index when iterating in the for-loop, e.g `for index, key in submission`. Feel free to open a feature request

contracts/FlowIDTableStaking.cdc
turbolent: We learned this the hard way in the Top Shot contract, which stores all plays and sets in the contract, so each time the Top Shot contract is used, all the data needs to be unnecessarily loaded from storage and decoded, which is expensive

contracts/epochs/FlowDKG.cdc
joshuahannan: Created! https://github.com/onflow/cadence/issues/781

turbolent: Awesome, thank you! 👍

joshuahannan:
@jordanschalm Will this version of the id table contract need the `dynamic weight` and `ejected` fields? Since we're moving some of the fields to use account storage instead of contract fields, I've realized we can upgrade the staking contract normally without having to do a manual state migration because we can just use the account storage for the new fields. Those two fields are a little more complicated, but we currently aren't using them at all, so it might make sense to save those for a future version so we don't have to worry about doing a state migration this time.

jordanschalm:
> Will this version of the id table contract need the dynamic weight and ejected fields?

Good point - To make use of these fields will anyway require a contract upgrade, so I'm happy to omit them from this version if it makes the upgrade simpler 👍

contracts/FlowIDTableStaking.cdc
```diff
@@ -276,6 +219,12 @@ pub contract FlowIDTableStaking {
         pub let tokensRequestedToUnstake: UFix64
         pub let initialWeight: UInt64
 
+        /// Derived Fields
+        pub let totalCommittedWithDelegators: UFix64
```
joshuahannan, line 223: @turbolent @SupunS Will adding and removing the fields from this struct definition in the Staking contract be a problem for upgrading? The reason I thought it might be ok is that as far as I know, nobody actually stores these structs anywhere because they only provide a snapshot at a specific moment in time, so upgrading this wouldn't corrupt them, but I am worried that it will still fail the upgrade checks

SupunS: Yeah, the validation would think adding a new field is a 'potential' problem and will fail. 

The problem we are facing right now is, there is no way to detect incoming dependencies to a contract/struct (e.g: `FlowIDTableStaking` ), and no way to know whether there are any stored data or not. So assuming the worst-case scenario, we are failing such scenarios for any contract/struct.

Removing a field is fine though, because:
  - It leaves data for the removed field unused at the storage, as it is no longer accessible.
  - However, it does not cause any runtime crashes, while trying to decode.

I am working on documenting all these criteria and rationals behind the update validations. It is about 80% completed. I would be able to push it out very soon.

joshuahannan:
@jordanschalm I resolved all the merge conflicts. Just need an approval and I can merge. 😃

jordanschalm (approved):
Approving based on the constituent PRs having been approved

PR #64, opened 2020-11-20
adds checks for balance before depositing

joshuahannan:
Added more checks to balance of tokens before depositing because we still haven't upgraded to remove the pre-condition in the FungibleToken interface on emulator, testnet, and mainnet: 
https://github.com/onflow/flow-ft/commit/8d76a13fb3eb258e9ecde373c79627322b2f084a#diff-f3172a01eb5e5fc357b4a912786b833c2d7908e7219b921506ec25390ed3bc37

Also, added the unstakeAll script and tests for it, because apparently I forgot to test it before. 🤦


robmyers (approved):
Looks good! 👍

PR #65, opened 2020-11-20
further checks for access nodes

joshuahannan:


contracts/FlowIDTableStaking.cdc
```diff
@@ -672,7 +672,8 @@ pub contract FlowIDTableStaking {
 
                 /// If the tokens that they have committed for the next epoch
                 /// do not meet the minimum requirements
-                if (totalTokensCommitted < FlowIDTableStaking.minimumStakeRequired[nodeRecord.role]!) || approvedNodeIDs[nodeID] == nil {
+                if !FlowIDTableStaking.isGreaterThanMinimumForRole(numTokens: totalTokensCommitted, role: nodeRecord.role) ||
+                   (approvedNodeIDs[nodeID] == nil) {
```
zhangchiqing, line 676: Consider doing the cheaper check first.
```suggestion
                if (approvedNodeIDs[nodeID] == nil || !FlowIDTableStaking.isGreaterThanMinimumForRole(numTokens: totalTokensCommitted, role: nodeRecord.role) 
                   ) {
```

Could you add some comments explain why we need the condition `approvedNodeIDs[nodeID] == nil`?

zhangchiqing (approved):
Looks good.

contracts/FlowIDTableStaking.cdc
joshuahannan: yep, I added a comment. That is where the protocol specifies nodes who are operating a node properly in order to prevent random users from registering a "node" without actually operating one. Currently, we just do this manually with our approved list of trusted operators

PR #67, opened 2020-11-21
Add lease account creation transaction

Kay-Zee:
## Context

Adds a template for creating an account that only takes 1 (1000 weight) key for the shared locked token account.


joshuahannan (approved):
Looks good!

PR #71, opened 2020-11-25
Add transaction to create lease account for existing account

boczeratul:
- Add transaction to create lease account for existing account.
- Add corresponding helper template function

// TODO
However, there isn't a test case associated with the transaction yet.


joshuahannan (approved):
Looks good to me! I trust that you've tested it. I'm including a PR soon with more tests, so I'll include a test for this in there

PR #72, opened 2020-11-25
Use open-sourced emulator, enable tests on CI

turbolent:


lib/go/test/flow_idtable_staking_test.go
```diff
@@ -76,16 +75,16 @@ func TestIDTable(t *testing.T) {
 		false,
 	)
 
-	var idTableAddress sdk.Address
+	var idTableAddress flow.Address
```
joshuahannan, line 78: What is the difference between using `sdk.Address` and `flow.Address`? I was using `flow.Address` before, but had to change to `sdk.Address` to get some of the event parsing to work. It is great if this works though!

joshuahannan (commented):
Did all the tests pass? Some of the deployment transactions weren't using the proper multiple contracts per account logic.

turbolent:
@joshuahannan Yep, I'm still updating them

lib/go/test/flow_idtable_staking_test.go
turbolent: There's none. There were two imports for the SDK, one plain (package name `flow`), and another one with alias `sdk`. I replace with one single import without alias, might avoid confusion. Can also easily revert!

joshuahannan: D'oh! No this is fine. I was just confused. Thanks for clarifying! 😅

turbolent:
@joshuahannan it's ready now

joshuahannan (approved):
Looks good! Thanks for making these updates. 😃

Issue #75, opened 2020-12-01
Create StorageFees smart contract


janezpodhostnik:
From the FLIP (https://github.com/onflow/flow/pull/99):

Each account will have a public `StorageCapacity.depositStorage` capability and a private `StorageCapacity` resource on predetermined paths.

The `StorageCapacity` resource defined on the `StorageFees` smart contract:
- has a `storageCapacity` field which defines how much storage can the account holding this resource use.
- has a `FlowVault` with Flow that was used for paying for this storage. This vault cannot be accessed by anyone except the `StorageFees` smart contract. The exception is the balance of this flow vault which should be readable by anyone.
- has an array of pairs of UInt64, UFix64 `purchases` used for keeping track of how much storage was purchased for how much flow. 
- has a `depositStorage` function that can only be used by the `StorageFees` smart contract (is this possible?):
    - it should increment the `storageCapacity` field deposit into the `FlowVault` and append to `purchases`

The `StorageFees` smart contract will also have the following responsibilities:

- Definitions:
    - `minimumStorageUnit`: define the minimum unit (chunk) size of storage in bytes. Storage can only be bought (or refunded) in a multiple of the minimum storage unit.
    - `minimumAccountStorage`: defines the minimum amount of `storage_capacity` an address can have and also the amount every new account has. `minimumAccountStorage` is a multiple of `minimumStorageUnit`.
    - `flowPerByte`: defines the cost of 1 byte of storage in FLOW tokens.
    - `flowPerAccountCreation`: define the cost of purchasing the initial minimum storage in FLOW tokens.
    - `refundingEnabled`: defines if accounts can refund storage
- `minimumAccountStorage`, `refundingEnabled`, `flowPerByte` and `flowPerAccountCreation` should be able to change in the future.
- Functions:
    - `setupStorageForAccount(paymentVault, authAccount)`:
        - The paymentVault should have funds for the minimum storage needed by the address.
        - Check if `StorageCapacity` is already present on the account (if so panic; account already bought some storage).
        - Create the `StorageCapacity` resource with `minimumAccountStorage` and put the `paymentVault` into it.
        - Put this `StorageCapacity` resource onto the account with a corresponding receiver.
    - `purchaseStorageForAddress(paymentVault, address, amount)`:
        - The payer pays for extra storage needed by the address.
        - Check if `StorageCapacity` is present on the address (if not so panic; account needs to buy initial storage first).
[...]

PR #76, opened 2020-12-01
Storage Fees smart contract

janezpodhostnik:
closes:https://github.com/onflow/flow-core-contracts/issues/75

superseded by https://github.com/onflow/flow-core-contracts/pull/96 due to a change in the definition of storage fees.


contracts/StorageFees.cdc
```diff
@@ -0,0 +1,180 @@
...
+    // Defines the minimum amount of `storage_capacity` an address can have and also the amount every new account has. `minimumAccountStorage` is a multiple of `minimumStorageUnit`.
+    pub var minimumAccountStorage: UInt64
+
+    // Defines the cost of 1 byte of storage in FLOW tokens.
+    pub var flowPerByte: UFix64
+
+    // Defines the cost of purchasing the initial minimum storage in FLOW tokens.
+    pub var flowPerAccountCreation: UFix64
+
+    // Defines the cost of 1 byte of storage in FLOW tokens.
```
dete, line 17: Copy/paste mistake! 😉

contracts/StorageFees.cdc
```diff
@@ -0,0 +1,180 @@
...
+    pub var refundingEnabled: Bool
+
+    access(contract) var idCounter: UInt64
```
dete, line 20: This variable's purpose isn't obvious to me, could use a comment here.

contracts/StorageFees.cdc
```diff
@@ -0,0 +1,180 @@
...
+        pub var storageCapacity: UInt64
+        access(contract) let address: Address
+        access(self) let vault: @FlowToken.Vault
+        access(self) let purchases: [StorageCapacityPurchase]
+
+        access(contract) fun addStorageCapacity(address: Address, amount: UInt64, paymentVault: @FlowToken.Vault){
+            pre{
+                self.address == address: "Unexpected address on storage capacity"
+            }
+            self.purchases.append(StorageCapacityPurchase(storageCapacity: amount, flowCost: paymentVault.balance))
```
dete, line 71: I'm really confused as to why we need to keep a record of the individual purchases. This seems to create a lot of bookkeeping overhead, and I'm not sure why that's valuable.

contracts/StorageFees.cdc
```diff
@@ -0,0 +1,180 @@
...
+        authAccount.link<&StorageCapacity{StorageCapacityCapability}>(
+            /public/storageCapacity,
+            target: /storage/storageCapacity
+        )
+    }
+
+    pub fun addStorageCapacity(to: Address, amount: UInt64, paymentVault: @FlowToken.Vault){
+        pre{
+            amount % StorageFees.minimumStorageUnit == UInt64(0): "Amount of storage capacity to add must be a multiple of the minimum storage unit"
+            paymentVault.balance != StorageFees.flowPerByte * UFix64(amount): "Adding ".concat(amount.toString()).concat(" storage capacity cost exactly ").concat(StorageCapacity.flowPerByte*amount).concat(" Flow tokens.")
```
dete, line 140: I think this should be `==` not `!=`...

contracts/StorageFees.cdc
```diff
@@ -0,0 +1,180 @@
...
+        pub fun setFlowPerByte(flowPerByte: UFix64) {
+            StorageFees.flowPerByte = flowPerByte
+        }
+
+        pub fun setFlowPerAccountCreation(flowPerAccountCreation: UFix64) {
+            StorageFees.flowPerAccountCreation = flowPerAccountCreation
+        }
+    }
+
+    pub resource interface StorageCapacityCapability{
```
dete, line 43: Let's not use the word "Capability" on resource objects. A "Capability" is a specific type in Cadence, and creating objects that have that name is sure to create confusion for people. StorageCapacityInterface, perhaps. StorageCapacityAccess, maybe.

contracts/StorageFees.cdc
```diff
@@ -0,0 +1,180 @@
...
+        }
+        let storageCapacityCapability = getAccount(to).getCapability<&StorageCapacity{StorageCapacityCapability}>(/public/storageCapacity)!.borrow()
+        if storageCapacityCapability == nil {
+            panic("Account needs to be setup first")
+        }
+
+        storageCapacityCapability?.addStorageCapacity(address: to, amount: amount, paymentVault: paymentVault)
+    }
+
+    pub fun refundStorageCapacity(from: Address, storageCapacityReference: &StorageCapacity, storageAmount: UInt64): @FlowToken.Vault {
```
dete, line 151: I'm not sure why we need an Address argument here. The StorageCapacity object has everything it needs to know which account it's managing. In fact, I'm not sure why this method exists at all, given that it just calls a method on one of the arguments!

dete (commented):
I think that we should have many, MANY more comments on contracts that represent core functionality of the protocol. Lots of people will want to be able to inspect these contracts, and see where and how their valuable assets are being handled. For example, I would think that all of the publicly accessible functions should have at least 3-5 lines of explanation. We had lots of discussions about how this should work, and there's lots of good thought that went into this implementation. Let's make that super clear!

contracts/StorageFees.cdc
janezpodhostnik: The idea was that even if the price of storage changes over time refunding purchased storage would give you back the same amount of flow as was originally used to purchase it.

I purchase 100 kb at 1 mFlow/kb then 50 kb at 2 mFlow/kb. The price now is 3 mFlow/kb. If I refund 100kb I would get back 100 mFlow (for the  50 kb at 2 mFlow/kb) + 50 (for the 100 kb at 1 mFlow/kb).

Is this not necessary? Can we just return at the current rate until there are funds  in the vault? This could lead to tokens being locked: 

I purchase 100 kb at 1 mFlow/kb; this is my only purchase.The price now is 0.1 mFlow/kb. I refund all my storage (down to the minimum) and I get 10 mFlow. There is still 90 mFlow in the vault I cant get to.

contracts/StorageFees.cdc
janezpodhostnik: One of the problems Im trying to prevent is storage swapping: two accounts swapping storage because one account bought it when it was cheap, and the other wants it. 

The process is as follows:
1. Account A has a lot of storage
2. in one transaction: A moves its StorageCapacity away from the standard location and calls `setupStorageForAccount` to get a new one so it has some capacity and the Tx doesn't fail 
3. it can now transfer its original storage to account B and B swaps it with its current storage 

I was trying to add an account field to StorageCapacity to prevent B using(or refunding) StorageCapacity created for A.

That is also what I was trying to do with the idCounter you commented above... I still need to think this scenario through...

janezpodhostnik:
I'm going to redesign StorageFees according to the discussion from the technical product meeting, so I'm moving this back to draft

contracts/StorageFees.cdc
```diff
@@ -0,0 +1,244 @@
...
+        pub fun setRefundingEnabled(_ enabled: Bool) {
+            if StorageFees.refundingEnabled == enabled {
+              return
+            }
+            StorageFees.refundingEnabled = enabled
+            emit RefundingEnabledChanged(enabled)
+        }
+
+        // Changes the amount storage capacity an account has per accounts' reserved storage.
+        pub fun setStorageBytesPerReservedFlow(_ storageBytesPerReservedFlow: UFix64) {
```
joshuahannan, line 60: Are there any pre-conditions on this that could check that the argument is not a unexpected value? I'm not sure if there are any disallowed values

contracts/StorageFees.cdc
```diff
@@ -0,0 +1,244 @@
...
+            if StorageFees.storageBytesPerReservedFlow == storageBytesPerReservedFlow {
+              return
+            }
+            StorageFees.storageBytesPerReservedFlow = storageBytesPerReservedFlow
+            emit StorageBytesPerReservedFlowChanged(storageBytesPerReservedFlow)
+        }
+
+        // Changes the minimum amount of Flow tokens an account has to have reserved.
+        pub fun setMinimumStorageReservation(_ minimumStorageReservation: UFix64) {
```
joshuahannan, line 69: Same comment as above about pre-conditions

contracts/StorageFees.cdc
```diff
@@ -0,0 +1,244 @@
...
+        access(contract) init(){}
+    }
+
+    // An interface for public access to accounts' storage reservation.
+    // If `StorageReservationReceiver` capability is not on the accounts' `storageReservationReceiverPath` path
+    // the account won't be able to receive additional storage capacity or refund current storage capacity.
+    pub resource interface StorageReservationReceiver {
+        pub fun deposit(from: @FlowToken.Vault) {
+            pre {
+                self.isInstance(Type<@StorageReservation>()): "The interface StorageReservationReceiver should only point to resource StorageReservation"
```
joshuahannan, line 86: Nice! Good use of `isInstance`! I haven't found a place to use it yet, but this is perfect

contracts/StorageFees.cdc
```diff
@@ -0,0 +1,244 @@
...
+            }
+        }
+    }
+
+    // A counter to uniquely identify all `StorageReservation` resources.
+    // This is needed to prevent an account refunding from a `StorageReservation` resources that is not on its own `storageReservationPath` path.
+    access(contract) var idCounter: UInt64
+
+    // The `StorageReservation` resource holds the amount of flow reserved for the accounts storage capacity. The amount of flow reserved in this resource is the accounts storage capacity.
```
joshuahannan, line 101:
```suggestion
    // The `StorageReservation` resource holds the amount of flow reserved for the accounts storage capacity. The amount of flow reserved in this resource determines the accounts storage capacity.
```

contracts/StorageFees.cdc
```diff
@@ -0,0 +1,244 @@
...
+
+        // Verify itself, that it is on the expected account on the expected path and of the expected type.
+        access(self) fun verify(): Bool {
+            let receiver = StorageFees.getStorageReservationReceiver(self.ownerAddress)
+            return receiver.verifyStorageReservation(ownerAddress: self.ownerAddress, storageReservationId: self.storageReservationId)
+        }
+
+        // This is called from the method above. The StorageReservation on this location should be the one that is expected.
+        access(contract) fun verifyStorageReservation(ownerAddress: Address, storageReservationId: UInt64): Bool {
+            if self.ownerAddress != ownerAddress || self.storageReservationId != self.storageReservationId  {
```
joshuahannan, line 166: Should there be a `self.owner.address!` somewhere here? It looks like this will never fail because the owner address that is passed to it from `verify` is always the same as `self.ownerAddress`

contracts/StorageFees.cdc
```diff
@@ -0,0 +1,244 @@
...
+
+        account.link<&StorageReservation{StorageReservationReceiver}>(
+            self.storageReservationReceiverPath,
+            target: self.storageReservationPath
+        )
+    }
+
+    // This function gets a reference to a `StorageReservationReceiver` from a address
+    pub fun getStorageReservationReceiver(_ address: Address): &StorageReservation{StorageReservationReceiver} {
+        return getAccount(address).getCapability<&StorageReservation{StorageReservationReceiver}>(self.storageReservationReceiverPath)!.borrow()!
```
joshuahannan, line 220:
```suggestion
        return getAccount(address).getCapability<&StorageReservation{StorageReservationReceiver}>(self.storageReservationReceiverPath)!.borrow()
        ?? panic("Could not borrow a reference to the StorageReservationReceiver")
```
Always good to have error messages.

joshuahannan (commented):
Great work! I think this will work great. Most of my comments were about naming conventions. Definitely push back if you disagree with any of them.

contracts/StorageFees.cdc
```diff
@@ -0,0 +1,244 @@
...
+// An account moving/deleting its `StorageFees.StorageReservation` resource will result in the transaction failing because the account will have no storage capacity.
+// Moving the `StorageFees.StorageReservationReceiver` will result in the account being unable to receive or withdraw reserved flow. This can be fixed by calling: 
+// ```
+// account.link<&StorageFees.StorageReservation{StorageFees.StorageReservationReceiver}>(
+//     StorageFees.storageReservationPath,
+//     target: StorageFees.storageReservationReceiverPath)
+// ```
+pub contract StorageFees {
+    // Emitted when storage capacity refunding is enabled or disabled.
+    pub event RefundingEnabledChanged(_ enabled: Bool)
```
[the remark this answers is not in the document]
janezpodhostnik: I though about this, And I cant really think of a solid argument either way. The only reason I went only one event is only to have less events defined.

contracts/StorageFees.cdc
```diff
@@ -0,0 +1,244 @@
...
+    
+    // Emitted when the amount of storage capacity an account has per reserved Flow token changes
+    pub event StorageBytesPerReservedFlowChanged(_ storageBytesPerReservedFlow: UFix64)
+
+    // Emitted when minimum amount of Flow
```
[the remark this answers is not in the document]
janezpodhostnik: Yep there was a comment mixup here! TY

contracts/StorageFees.cdc
```diff
@@ -0,0 +1,244 @@
...
+
+    // Defines the path where each account should have a `StorageReservationReceiver` capability
+    pub let storageReservationReceiverPath: PublicPath
+
+    // Defines the path where each account should have a `StorageReservation` capability
+    pub let storageReservationPath: StoragePath
+
+    // Defines how much storage capacity an account has per reserved Flow token.
+    // definition is written per unit of flow instead of the inverse, so there is no loss of precision calculating storage from flow, but there is loss of precision when calculating flow per storage.
+    pub var storageBytesPerReservedFlow: UFix64
```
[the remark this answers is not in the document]
janezpodhostnik: Yes this is better, because it specifies that these are tokens, and is shorter then `FlowTokens`.

contracts/StorageFees.cdc
janezpodhostnik: I'm not sure....
the only potential problem would be setting it to 0.0 but that transaction would definitely fail as the service account would not have sufficient storage anymore.

contracts/StorageFees.cdc
janezpodhostnik: No. As far as I understand the minimum can be anything.

contracts/StorageFees.cdc
janezpodhostnik: I will add more description to this function to try to make it cleared how it functions.

contracts/StorageFees.cdc
```diff
@@ -0,0 +1,244 @@
+import FlowToken from 0xFLOWTOKENADDRESS
+
+// The StorageFees smart contract
```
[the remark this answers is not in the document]
janezpodhostnik: 👍  on the block comments.

I was going with what I was used to from other languages when it comes to comment placements.  Usually when you comment a class the documentation is right above it. But i think it makes sense to put it on the top of the file here.

contracts/StorageFees.cdc
```diff
@@ -0,0 +1,244 @@
...
+// If any account fails this check the transaction wil fail.
+//
+// An account moving/deleting its `StorageFees.StorageReservation` resource will result in the transaction failing because the account will have no storage capacity.
+// Moving the `StorageFees.StorageReservationReceiver` will result in the account being unable to receive or withdraw reserved flow. This can be fixed by calling: 
+// ```
+// account.link<&StorageFees.StorageReservation{StorageFees.StorageReservationReceiver}>(
+//     StorageFees.storageReservationPath,
+//     target: StorageFees.storageReservationReceiverPath)
+// ```
+pub contract StorageFees {
```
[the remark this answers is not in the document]
janezpodhostnik: 👍

contracts/StorageFees.cdc
```diff
@@ -0,0 +1,244 @@
...
+    // Emitted when storage capacity refunding is enabled or disabled.
+    pub event RefundingEnabledChanged(_ enabled: Bool)
+    
+    // Emitted when the amount of storage capacity an account has per reserved Flow token changes
+    pub event StorageBytesPerReservedFlowChanged(_ storageBytesPerReservedFlow: UFix64)
```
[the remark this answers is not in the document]
janezpodhostnik: 👍

contracts/StorageFees.cdc
```diff
@@ -0,0 +1,244 @@
...
+
+    // Emitted when minimum amount of Flow 
+    pub event MinimumStorageReservationChanged(_ minimumStorageReservation: UFix64)
+
+    // Emitted when the minimum amount of Flow tokens that an account needs to have reserved for storage capacity changes.
```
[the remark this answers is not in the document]
janezpodhostnik: yup!

contracts/StorageFees.cdc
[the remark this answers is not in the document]
joshuahannan: yeah, I think it is probably fine to leave it this way. 👍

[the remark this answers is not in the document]
janezpodhostnik: ups! I already changed it :). Changing it back right away.

joshuahannan (approved):
Great work. 😃  I just have a few typo fixes

PR #77, opened 2020-12-02
Fixes minimum check for nodes and adds many nodes testing

joshuahannan:
I made a boneheaded mistake when adding the `getTotalCommittedBalance` utility function earlier and used it to check if the node was above the minimum stake, even though this one counted delegator commitment as well.
This PR:
* renames those functions
* adds a few more with more descriptive names
* Changes the minimum to use `getNodeCommittedBalanceWithoutDelegators`
* Tests unstaking below the minimum even if delegator total is above the minimum (FAILS)

I also added tests here for registering a bunch of nodes and delegators to test gas usage. The number of nodes is set to low numbers right now so the CI tests pass, but I'll still be testing them locally with higher numbers


contracts/FlowIDTableStaking.cdc
```diff
@@ -982,10 +986,26 @@ pub contract FlowIDTableStaking {
...
+        let nodeRecord = self.borrowNodeRecord(nodeID)
+
+        if (nodeRecord.tokensCommitted.balance + nodeRecord.tokensStaked.balance) < nodeRecord.tokensRequestedToUnstake {
+            return 0.0
+        } else {
+            var sum: UFix64 = 0.0
+
+            sum = nodeRecord.tokensCommitted.balance + nodeRecord.tokensStaked.balance - nodeRecord.tokensRequestedToUnstake
+
+            return sum
```
robmyers, line 1001: Could this be a one-liner?

contracts/FlowIDTableStaking.cdc
```diff
@@ -1004,6 +1024,21 @@ pub contract FlowIDTableStaking {
...
+    /// Gets the total amount of tokens that have been staked for a node. 
+    /// The sum from the node operator and all its delegators
+    pub fun getNodeStakedBalanceWithDelegators(_ nodeID: String): UFix64 {
+        let nodeRecord = self.borrowNodeRecord(nodeID)
+
+        var sum: UFix64 = nodeRecord.tokensStaked.balance
+
+        for delegator in nodeRecord.delegators.keys {
+            let delRecord = nodeRecord.borrowDelegatorRecord(delegator)
+            sum = sum + delRecord.tokensStaked.balance
```
robmyers, line 1036: It doesn't help here, but this is the kind of record keeping that in future we should update with each transaction from (in this example) a delegator.

lib/go/templates/idtable_staking_templates.go
```diff
@@ -338,9 +339,17 @@ func GenerateGetUnstakingRequestScript(env Environment) []byte {
 }
 
 // GenerateGetTotalCommitmentBalanceScript creates a script
-// that returns the balance of the total committed tokens of a node
+// that returns the balance of the total committed tokens of a node plus delegators
```
robmyers, line 342: 👍

transactions/idTableStaking/get_node_total_commitment_without_delegators.cdc
```diff
@@ -0,0 +1,7 @@
+import FlowIDTableStaking from 0xIDENTITYTABLEADDRESS
+
+// This script returns the balance of staked tokens of a node
+
+pub fun main(nodeID: String): UFix64 {
+    return FlowIDTableStaking.getNodeCommittedBalanceWithoutDelegators(nodeID)
+}
```
robmyers, line 7: I know it's a UNIX affectation but this and get_node_total_commitment.cdc don't have newlines at the end.

robmyers (approved):
Wow! flow_idtable_nodes_test.go is great!

It might be worth pulling the number of nodes form an environment variable and defaulting to 50 if it isn't present.

contracts/FlowIDTableStaking.cdc
joshuahannan: yes, I am learning a lot about how to write better cadence code from this testing

contracts/FlowIDTableStaking.cdc
joshuahannan: yes it could be simpler. I'll change it. Good suggestion!

contracts/FlowIDTableStaking.cdc
```diff
@@ -733,35 +735,41 @@ pub contract FlowIDTableStaking {
...
+            if totalStaked >= UFix64(100000000.0) {
+                // Maximum UFix64 divisor is 100M so we need to scale the numbers
+                // in order to not cause overflow
+                let div1000dividend = FlowIDTableStaking.epochTokenPayout / 1000.0
+                let div1000divisor = totalStaked / 1000.0
+                totalRewardScale = div1000dividend / div1000divisor
+            } else if totalStaked != 0.0 {
+                totalRewardScale = FlowIDTableStaking.epochTokenPayout / totalStaked
+            }
+
```
robmyers, line 749: This can still lose us accuracy in the fractional part of the number when scaling but it is 1000 times less likely to overflow. 👍

contracts/FlowIDTableStaking.cdc
```diff
@@ -473,18 +481,6 @@ pub contract FlowIDTableStaking {
...
-            for delegator in nodeRecord.delegators.keys {
-                let delRecord = nodeRecord.borrowDelegatorRecord(delegator)
-
-                if delRecord.tokensCommitted.balance > 0.0 {
-                    delRecord.tokensUnstaked.deposit(from: <-delRecord.tokensCommitted.withdraw(amount: delRecord.tokensCommitted.balance))
-                }
-
-                delRecord.tokensRequestedToUnstake = delRecord.tokensStaked.balance
-            }
-
```
robmyers, line 487: I'm not quite following what this is replaced by?

contracts/FlowIDTableStaking.cdc
```diff
@@ -733,35 +735,41 @@ pub contract FlowIDTableStaking {
...
+
+            if totalStaked >= UFix64(100000000.0) {
+                // Maximum UFix64 divisor is 100M so we need to scale the numbers
+                // in order to not cause overflow
+                let div1000dividend = FlowIDTableStaking.epochTokenPayout / 1000.0
+                let div1000divisor = totalStaked / 1000.0
+                totalRewardScale = div1000dividend / div1000divisor
+            } else if totalStaked != 0.0 {
+                totalRewardScale = FlowIDTableStaking.epochTokenPayout / totalStaked
+            }
```
robmyers, line 748: Each compoenent, and therefore the final calculation, can lose some tiny fractional amount of accuracy if it has many digits after the decimal point. But this is 1000 times less likely to overflow. :-) 👍

robmyers (approved):
With apologies, I have one probably silly question.

Otherwise 👍

contracts/FlowIDTableStaking.cdc
joshuahannan: I actually didn't have to add anything to replace this, because it was already replaced. In the `endStaking` function, we check if the node is committed below the minimum, and if they are, it marks all their delegators to unstake.

https://github.com/onflow/flow-core-contracts/blob/josh/many-nodes/contracts/FlowIDTableStaking.cdc#L682

contracts/FlowIDTableStaking.cdc
```diff
@@ -733,6 +733,8 @@ pub contract FlowIDTableStaking {
                 }
             }
 
+            log("found total!")
```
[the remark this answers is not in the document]
joshuahannan: thats from an old version of the contract. so it is removed now

contracts/FlowIDTableStaking.cdc
```diff
@@ -740,8 +740,20 @@ pub contract FlowIDTableStaking {
...
+                var rewardPortion: UFix64 = 0.0
+
                 /// Calculate the amount of tokens that this node operator receives
-                let rewardAmount = FlowIDTableStaking.epochTokenPayout * (nodeRecord.tokensStaked.balance / totalStaked)
+                if totalStaked >= 100000000.0 {
+                    // Maximum UFix64 divisor is 100M so we need to scale the numbers
+                    // in order to not cause overflow
+                    let div1000dividend = nodeRecord.tokensStaked.balance / 1000.0
+                    let div1000divisor = totalStaked / 1000.0
+                    rewardPortion = div1000dividend / div1000divisor
```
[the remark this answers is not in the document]
joshuahannan: This is the accepted solution for now, but will probably change in the future. Though I don't know if it would be better to use 1000 or 100 as the scaling factor. I posted a question in #flow-computation yesterday about it but didn't get a response.

contracts/FlowIDTableStaking.cdc
```diff
@@ -934,6 +959,26 @@ pub contract FlowIDTableStaking {
...
+
+    /// Utility Function that checks a delegator's overall committed balance from its borrowed record
+    access(contract) fun calculateDelegatorCommittedBalance(_ delegatorRecord: &DelegatorRecord): UFix64 {
+        
+        if (delegatorRecord.tokensCommitted.balance + delegatorRecord.tokensStaked.balance) < delegatorRecord.tokensRequestedToUnstake {
+            return 0.0
+        } else {
+            return delegatorRecord.tokensCommitted.balance + delegatorRecord.tokensStaked.balance - delegatorRecord.tokensRequestedToUnstake
+        }
+    }
```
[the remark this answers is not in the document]
joshuahannan: yeah that probably makes sense. I'll move it

transactions/idTableStaking/delegation/register_many_delegators.cdc
```diff
@@ -0,0 +1,23 @@
...
+
+transaction(nodeIDs: [String], paths: [Path]) {
+
+    prepare(acct: AuthAccount) {
+
+        var i = 0
+
+        for path in paths {
+            // Create a new delegator object for the node
+            let newDelegator <- FlowIDTableStaking.registerNewDelegator(nodeID: nodeIDs[0])
```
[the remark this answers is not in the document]
joshuahannan: That was supposed to be `i` there. Thanks for pointing that out!

joshuahannan:
@psiemens This branch is using your special version of the emulator that can set a custom max gas limit. Is it ok to merge this with that version of the emulator, or do we need to keep it separate from the rest of the tests?

psiemens:
@joshuahannan I'd prefer to merge the emulator PR and create a new release first. I can do that today.

Kay-Zee (approved):

PR #79, opened 2020-12-08
Add auto top-up logic to stake and delegate registration templates

psiemens:
This PR updates the "Register Node" and "Register Delegator" transactions to automatically withdraw from the unlocked account if the locked balance is less than the specified amount.


joshuahannan (approved):
Looks good! Thanks for putting these together!

PR #82, opened 2020-12-11
Add implementation of KeyManager contract

psiemens:
This PR implements the `KeyManager` contract used to programatically add a key to an otherwise frozen account.


contracts/TokenHolderKeyManager.cdc
```diff
@@ -0,0 +1,34 @@
+import KeyManager from 0xKEYMANAGERADDRESS
+
+// The TokenHolderKeyManager contract is an implementation of
+// the KeyManager interface intended for use by FLOW token holders.
+//
+// One instance is deployed to each token holder account.
+// Deployment is executed a with signature from the administrator,
```
robmyers, line 7:
```suggestion
// Deployment is executed a with signature from the administrator
// whose key has been added to the account by the account owner,
```

contracts/KeyManager.cdc
```diff
@@ -0,0 +1,15 @@
+// The KeyManager interface defines the functionality 
+// used to securely add public keys to a Flow account
+// through a Cadence contract.
+//
+// This interface is deployed once globally and implemented by 
+// all token holders who wish to allow their keys to be managed
+// by an administrator.
+pub contract interface KeyManager {
+
+  pub resource interface KeyAdder {
```
robmyers, line 10: I keep reading this as KeyAddr but that's probably just me.

transactions/keyManager/add_key.cdc
```diff
@@ -0,0 +1,23 @@
+import KeyManager from 0xKEYMANAGERADDRESS
+
+// This transaction adds a public key to a token holder account
+// using a KeyAdder resource.
+//
+// Each token holder provides a single KeyAdder resource to the administrator
+// stored at a unique path in the administrator account.
```
robmyers, line 7: It might be worth noting that, as no API is exposed for other parties to work with these resources, storing the resources directly in the admin account's storage with well-named paths is the best pattern to use here.

robmyers (approved):
I have some trivial comments but this is really solid. 👍

transactions/keyManager/add_key.cdc
joshuahannan: yeah I agree. Would it be better to define the paths in the contract?

joshuahannan (commented):
It looks fine to me. I have two general questions though
1. What is the immediate use case for this contract?
2. Does it make sense to put it in the core contracts repo, or should it be somewhere else?

PR #88, opened 2020-12-22
add cancel unstaking request

joshuahannan:
* Allows a node or delegator to cancel an unstaking request before it has been fulfilled and moved into the `tokensUnstaking` bucket
* If it is already in the `tokensUnstaking` bucket, it still cannot be canceled.


contracts/FlowIDTableStaking.cdc
```diff
@@ -408,14 +402,24 @@ pub contract FlowIDTableStaking {
...
+
+            if remainingAmount > 0.0 {
 
                 /// Add the removed tokens to tokens committed
-                nodeRecord.tokensCommitted.deposit(from: <-nodeRecord.tokensUnstaked.withdraw(amount: amount))
+                nodeRecord.tokensCommitted.deposit(from: <-nodeRecord.tokensUnstaked.withdraw(amount: remainingAmount))
 
-                emit TokensCommitted(nodeID: nodeRecord.id, amount: amount)
+                emit TokensCommitted(nodeID: nodeRecord.id, amount: remainingAmount)
             }
```
Kay-Zee, line 423: I guess we don't have any events related to unstaking, it's just seting the amount to 0.0

lib/go/test/flow_idtable_staking_test.go
```diff
@@ -2664,6 +2664,67 @@ func TestIDTable(t *testing.T) {
 
 	})
 
+	t.Run("Should be able cancel unstake request for delegator", func(t *testing.T) {
+
```
Kay-Zee, line 2668: This carries on the previous state, so at this point, there is 40k delegated, right?

one thing that could make these test additions easier to review is to have comments indicating the start state for the `Run`

Kay-Zee (approved):

contracts/FlowIDTableStaking.cdc
```diff
@@ -110,13 +110,10 @@ pub contract FlowIDTableStaking {
...
         pub var role: UInt8
 
-        /// The address used for networking
         pub(set) var networkingAddress: String
-
-        /// the public key for networking
+ 
         pub(set) var networkingKey: String
 
-        /// the public key for staking
```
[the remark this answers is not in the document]
joshuahannan: Removing them because the transaction for deploying the contract is too large, so I had to make it smaller

contracts/FlowIDTableStaking.cdc
joshuahannan: yeah the event for unstaking doesn't happen until the actual tokens are unstaked

lib/go/test/flow_idtable_staking_test.go
joshuahannan: yeah that is a good idea. I can start cleaning these up and adding those comments

PR #89, opened 2020-12-29
Add get info for locked accounts

Kay-Zee:
I've been using these scripts a lot, and wanted to make use of them in this issue: https://github.com/onflow/flow-cli/issues/20

originally from @psiemens and I believe is being used in Flow Port.


joshuahannan (approved):
Thanks for adding these!

PR #91, opened 2021-01-04
Adding scripts and interfaces

joshuahannan:
* Added interfaces to the staking contract to include node and delegator IDs

Added scripts for
* Getting node info from an address
* Getting delegator info from an address
* Changed the name of some locked token script getters to reflect they are for locked accounts

Changed transactions:
* Register node now creates a public capability
* Register delegator now creates a public capability
* Added transaction to change the payout and end the epoch at the same time
* Added transaction to add a public capability to an account with an existing `NodeStaker` object with no capability

Removed some comments and lines to keep the file smaller so it would fit in the transaction size


transactions/idTableStaking/register_node.cdc
```diff
@@ -28,6 +28,8 @@ transaction(id: String,
         
         if acct.borrow<&FlowIDTableStaking.NodeStaker>(from: FlowIDTableStaking.NodeStakerStoragePath) == nil {
             acct.save(<-nodeStaker, to: FlowIDTableStaking.NodeStakerStoragePath)
+
+            acct.link<&{FlowIDTableStaking.NodeStakerPublic}>(FlowIDTableStaking.NodeStakerPublicPath, target: FlowIDTableStaking.NodeStakerStoragePath)
```
psiemens, line 32: I noticed that `NodeStakerPublicPath` was already defined. Was it being used for something else before? Should there be a similar constant for `NodeDelegator` (`NodeDelegatorPublicPath`)?

joshuahannan: It was only being used for testing before, but now it is useful in real life! We can't add a `NodeDelegatorPublicPath` though because we can't add any state variables in an upgrade. We could add a getter function for it though that just returns the path we want. What do you think?

psiemens: Ah I see, didn't realize that was a restriction! I think the way you have it is fine

psiemens (approved):
LGTM

Kay-Zee (approved):

PR #94, opened 2021-01-11
Withdraw stake + delegate rewards directly to unlocked account

psiemens:
This PR adjusts the `withdraw_rewarded_tokens` scripts to move rewards to the unlocked account by default. I preserved the old transactions so that we still have a test for the case where the rewards move to the locked account.


joshuahannan (approved):
LGTM! Thanks for adding this. 👌

PR #103, opened 2021-01-19
Update to latest emulator, and fix and parallelize tests

turbolent:
- Update to Emulator v0.14.0-beta3, which now uses these very core contracts
- Parallelize the tests to speed them up
- Disable storage limits


joshuahannan (approved):
Looks good!

PR #104, opened 2021-01-19
Adds staking enabled field

joshuahannan:
* Adds a field in the contract to indicate if staking is currently enabled by the contract.
* The admin can toggle this field, and it is also set to false during `endStaking` and set to true during `moveTokens`
* If it is false, all staking operations will fail
* Also added tests and transactions
* Put the old staking contract in a separate file because I want to use it later to test upgrades on the emulator. There is no need to review the old one.


lib/go/test/flow_idtable_staking_test.go
```diff
@@ -2942,27 +3132,27 @@ func TestIDTableStaking(t *testing.T) {
 
 	})
 
-	// End staking auction and move tokens in the same transaction
-	t.Run("Should end staking auction and move tokens in the same transaction", func(t *testing.T) {
+	// // End staking auction and move tokens in the same transaction
```
turbolent, line 3135: Any reason this is commented out?

turbolent (approved):
Nice

contracts/FlowIDTableStaking.cdc
```diff
@@ -300,22 +280,11 @@ pub contract FlowIDTableStaking {
     /// Records the staking info associated with a delegator
     /// Stored in the node's NodeRecord
     pub resource DelegatorRecord {
-        /// Tokens this delegator has committed for the next epoch
```
turbolent, line 303: Any reason the documentation comments were removed?

joshuahannan: So I could fit the contract in the max transaction size. I'd like to have them in there, but I can't fit them unfortunately

lib/go/test/flow_idtable_staking_test.go
joshuahannan: because I updated the transaction after it to do the same thing. I'll remove the commented out lines.

contracts/FlowIDTableStaking.cdc
turbolent: @ramtinms Do we have a way for the service account to send larger transaction then the default?

PR #105, opened 2021-01-21
Transaction to add delegator capability

joshuahannan:
Adds a transaction to add a public delegator capability if an account doesn't already have one


transactions/idTableStaking/delegation/delegator_add_capability.cdc
```diff
@@ -0,0 +1,22 @@
...
+import FlowToken from 0xFLOWTOKENADDRESS
+
+// This transaction adds a public delegator capability to an account with
+// an existing NodeDelegator object
+
+transaction {
+
+    prepare(acct: AuthAccount) {
+
+        if acct.borrow<&FlowIDTableStaking.NodeDelegator>(from: FlowIDTableStaking.DelegatorStoragePath) == nil ||
```
turbolent, line 11: I've seen this "does storage contain something at this path with this type?" check a couple times before. 

Could you please open a feature request issue in the Cadence repo to add a "check" function, just like we have it for capabilities? 

This should be easy to implement and would be more idiomatic.

Thank you!

turbolent (approved):
Looks good!

transactions/idTableStaking/delegation/delegator_add_capability.cdc
```diff
@@ -0,0 +1,22 @@
...
+            acct.getCapability<&{FlowIDTableStaking.NodeDelegatorPublic}>(/public/flowStakingDelegator).check()
+        {
+            return
+        }
+
+        acct.link<&{FlowIDTableStaking.NodeDelegatorPublic}>(
+            /public/flowStakingDelegator,
```
robmyers, line 18: It's a shame this isn't a contract property like the StoragePath, but I know this can't be added in an upgrade and the contract is already pretty full. :-)

robmyers (approved):
Looks good. 👍

transactions/idTableStaking/delegation/delegator_add_capability.cdc
joshuahannan: Yeah I'll open an issue

PR #106, opened 2021-01-21
Adds metadata claimed fields to reduce gas use

joshuahannan:
Part of the Epochs feature. Will not be added to mainnet or testnet until epochs are enabled.

The old way the contract was checking that metadata was unique was by iterating through the whole table.
This PR adds contract fields that check if a piece of metadata is already in use automatically.
no new test cases were written because this case is already tested
Also had to remove some more comments to stay in the correct transaction size limit


contracts/FlowIDTableStaking.cdc
```diff
@@ -631,10 +615,13 @@ pub contract FlowIDTableStaking {
 
         /// Remove a node from the record
         pub fun removeNode(_ nodeID: String): @NodeRecord {
-            // Remove the node from the table
             let node <- FlowIDTableStaking.nodes.remove(key: nodeID)
                 ?? panic("Could not find a node with the specified ID")
 
+            FlowIDTableStaking.networkingAddressClaimed[node.networkingAddress] = nil
```
orodio, line 621: Would `FlowIDTableStaking.networkingAddressClaimed.remove(key: node.networkingAddress)` work here? might be better regarding storage?

orodio (approved):

contracts/FlowIDTableStaking.cdc
turbolent: Good catch! I also think calling remove makes it a bit more obvious that this removes an entry.
Semantics/performance wise the two variants are actually the same: Assigning nil calls remove internally: https://github.com/onflow/cadence/blob/13ee70e94308b3bb892c6907ec6c495ed3d85759/runtime/interpreter/value.go#L5885-L5894

turbolent (approved):

robmyers (approved):
This is a great optimization. 👍 

This struck me in your pull comment:

> Also had to remove some more comments to stay in the correct transaction size limit
:eyes:

If we're hitting this size limit already with our contracts it's worth capturing what the limits are and bearing them in mind when structuring future code.

contracts/FlowIDTableStaking.cdc
joshuahannan: yeah I can switch to that. I forgot about that when writing it

PR #108, opened 2021-01-22
Adds DKG contract and transactions

joshuahannan:
Adds an untested first draft for the DKG contract for feedback


contracts/epochs/FlowDKG.cdc
```diff
@@ -0,0 +1,251 @@
...
+    // only be valid if the node operator has been confirmed as a consensus node for the next epoch.
+    pub resource Participant {
+
+        pub let nodeID: String
+
+        // Submits a whiteboard message to the contract
+        pub fun sendMessage(phase: UInt8, _ content: String) {
+            pre {
+                FlowDKG.participantIsRegistered(self.nodeID): "Cannot send whiteboard message if not registered for the current epoch"
+                phase > FlowDKG.currentPhase && phase < UInt8(4): "Phase submission is invalid"
```
jordanschalm, line 83:
```suggestion
                phase >= FlowDKG.currentPhase && phase < UInt8(4): "Phase submission is invalid"
```

If we're in phase 1, we should be able to send phase 1 messages

contracts/epochs/FlowDKG.cdc
```diff
@@ -0,0 +1,251 @@
...
+        }
+
+        /// Resets all the fields for tracking the current DKG process
+        /// and sets the given node IDs as registered
+        pub fun startDKG(nodeIDs: [String]) {
+            FlowDKG.currentPhase = DKGPhase.Phase1
+
+            for id in nodeIDs {
+                FlowDKG.nodeHasSubmitted[id] = false
+            }
```
jordanschalm, line 188: I think we also need to reset existing values in this mapping. Otherwise DKG participants from last epoch that aren't DKG participants in this epoch would still be considered valid.
```suggestion
            FlowDKG.nodeHasSubmitted = {}
            for id in nodeIDs {
                FlowDKG.nodeHasSubmitted[id] = false
            }
```

contracts/epochs/FlowDKG.cdc
```diff
@@ -0,0 +1,251 @@
...
+        }
+
+        /// Sends the final key vector submission. 
+        /// Can only be called during the final phase and by consensus nodes that are registered.
+        pub fun sendFinalSubmission(_ submission: [String]) {
+            pre {
+                FlowDKG.participantIsRegistered(self.nodeID): "Cannot send final submission if not registered for the current epoch"
+                !FlowDKG.nodeHasSubmittedFinal(self.nodeID)!: "Cannot submit a final submission twice"
+                FlowDKG.currentPhase == DKGPhase.finalSubmission: "Can only send final submission in the final DKG phase"
+                // submission should be a certain length?
```
jordanschalm, line 103: It should be `n+1` where `n` is the number of participants in the current DKG instance -- I think `n` would be `FlowDKG.nodeHasSubmitted.keys.length`

contracts/epochs/FlowDKG.cdc
```diff
@@ -0,0 +1,251 @@
...
+                // move on to the next one
+                if submission.length != existingSubmission.length {
+                    return
+                }
+
+                // Check each key in the submiission to make sure that it matches
+                // the existing one
+                for key in submission {
+                    if key.length != 128 {
+                        // correct key length?
```
jordanschalm, line 124: I'll need to confirm, and this will depend on the encoding scheme we use when submitting the transaction. Maybe add a constant for this with a TODO for now

contracts/epochs/FlowDKG.cdc
```diff
@@ -0,0 +1,251 @@
...
+                var index = 0
+                // If the submission length is different than the one being compared to
+                // move on to the next one
+                if submission.length != existingSubmission.length {
+                    return
+                }
+
+                // Check each key in the submiission to make sure that it matches
+                // the existing one
+                for key in submission {
```
jordanschalm, line 122: Could have a helper here: `FlowDKG.submissionsEqual(existingSubmission, submission)`

contracts/epochs/FlowDKG.cdc
```diff
@@ -0,0 +1,251 @@
...
+    access(account) var nodeHasSubmitted: {String: Bool}
+
+    /// Array of unique final submissions from nodes
+    /// if a final submission is sent that matches one that already has been submitted
+    /// this array will not change at all
+    access(account) var uniqueFinalSubmissions: [[String]]
+
+    /// Tracks how many submissions have been sent
+    /// for each unique final submission
+    access(account) var uniqueFinalSubmissionCount: {Int: UInt64}
```
jordanschalm, line 46: I think eventually we will want to be able to determine which node submitted which key vector, since submitting an incorrect key vector is likely a slashable offense. More a note for the future though, not necessary for the first version.

jordanschalm (commented):
👏 👏 Nice work!

contracts/epochs/FlowDKG.cdc
joshuahannan: I'm still going to add it now just so we have it

contracts/epochs/FlowDKG.cdc
joshuahannan: Any update on the key length? @jordanschalm @tarakby

tarakby: what key are we considering here @joshuahannan ?

jordanschalm: [This is the list of public keys in the final submission](https://www.notion.so/dapperlabs/Smart-contract-based-DKG-29ad04d8996c4aae8019348ce49085a0#3da54491bbf54578bb0e8b384aa42c01)

contracts/epochs/FlowEpoch.cdc
```diff
@@ -71,7 +71,11 @@ pub contract FlowEpochs {
 
         // The source of randomness to seed the leader selection algorithm with 
         // for the upcoming epoch.
-        randomSource: String
+        randomSource: String,
+
+        DKGPhase1FinalView: UInt64,
```
jordanschalm, line 76:
```suggestion
        // The deadlines of each phase in the DKG protocol to be completed in the upcoming
        // EpochSetup phase. Deadlines are specified in terms of a consensus view number. 
        // When a DKG participant observes a finalized and sealed block with view greater 
        // than the given deadline, it can safely transition to the next phase. 
        DKGPhase1FinalView: UInt64,
```

transactions/dkg/admin/next_phase.cdc
```diff
@@ -0,0 +1,19 @@
+import FlowDKG from 0xDKGADDRESS
```
jordanschalm, line 1: can remove this

transactions/dkg/scripts/get_current_phase.cdc
```diff
@@ -0,0 +1,5 @@
+import FlowDKG from 0xDKGADDRESS
+
+pub fun main(): FlowDKG.DKGPhase {
```
jordanschalm, line 3: can remove

contracts/epochs/FlowDKG.cdc
tarakby: A single public key is 96 bytes. So following the document the total is `96 * (D+1)` bytes.
However, the keys would be all `nil` if the dkg node detected that the protocol failed.

contracts/epochs/FlowDKG.cdc
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    /// Returns the final set of keys if any one set of keys has more than 50% submissions
+    /// Returns nil if not found (incomplete)
+    pub fun dkgCompleted(): [String]? {
+
+        var index = 0
+
+        for submission in self.uniqueFinalSubmissions {
+            if self.uniqueFinalSubmissionCount[index]! > UInt64(self.nodeFinalSubmission.keys.length/2) {
```
jordanschalm, line 246: Looking back at the docs, this should be slightly different: `floor((n-1)/2)` rather than `floor(n/2)`

https://www.notion.so/dapperlabs/Smart-contract-based-DKG-29ad04d8996c4aae8019348ce49085a0#1be0b475e8994e398ff77baad91c8167

jordanschalm (approved):

contracts/epochs/FlowDKG.cdc
jordanschalm: 👍 Added this to the notion: https://www.notion.so/dapperlabs/Smart-contract-based-DKG-29ad04d8996c4aae8019348ce49085a0#e6e411af2cea452a94a55159de791a6f

> However, the keys would be all nil if the dkg node detected that the protocol failed.

We consider the DKG protocol failing to be a un-recoverable failure currently, so I'm OK to simply not handle this case in the contract for now

contracts/epochs/FlowDKG.cdc
```diff
@@ -0,0 +1,274 @@
...
+
+        }
+
+        /// Sends the final key vector submission. 
+        /// Can only be called by consensus nodes that are registered.
+        pub fun sendFinalSubmission(_ submission: [String]) {
+            pre {
+                FlowDKG.participantIsRegistered(self.nodeID): "Cannot send final submission if not registered for the current epoch"
+                FlowDKG.nodeHasSubmitted(self.nodeID)!.length == 0: "Cannot submit a final submission twice"
+                submission.length == FlowDKG.nodeFinalSubmission.keys.length + 1: "Submission must have number of elements equal to the number of nodes participating in the DKG"
```
tarakby, line 95:
```suggestion
                submission.length == FlowDKG.nodeFinalSubmission.keys.length + 1: "Submission must have number of elements equal to the number of nodes participating in the DKG plus one"
```

contracts/epochs/FlowDKG.cdc
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    /// Get the array of all the unique final submissions
+    pub fun getFinalSubmissions(): [[String]] {
+        return self.uniqueFinalSubmissions
+    }
+
+    /// Returns the final set of keys if any one set of keys has more than 50% submissions
+    /// Returns nil if not found (incomplete)
+    pub fun dkgCompleted(): [String]? {
```
tarakby, line 270: I'm wondering when this function is called. Is there a deadline for nodes to submit their submissions after which `dkgCompleted` is called?

tarakby (commented):
Thank you for implementing the logic Josh 👏 

I have left a few comments and questions below.

contracts/epochs/FlowDKG.cdc
```diff
@@ -0,0 +1,274 @@
...
+
+                // update the index counter
+                finalSubmissionIndex = finalSubmissionIndex + 1
+
+                // If no matches were found, add this submission as a new unique one
+                // and emit an event
+                if finalSubmissionIndex == FlowDKG.uniqueFinalSubmissions.length {
+                    FlowDKG.uniqueFinalSubmissionCount[finalSubmissionIndex] = 0
+                    FlowDKG.uniqueFinalSubmissions.append(submission)
+                    emit FinalSubmission(nodeID: self.nodeID, submission: submission)
```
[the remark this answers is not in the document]
joshuahannan: ok. I usually default to adding events anywhere where I think it might be useful to do it, but if this isn't necessary, I'll remove it

contracts/epochs/FlowDKG.cdc
```diff
@@ -0,0 +1,274 @@
...
+                }
+
+                // update the index counter
+                finalSubmissionIndex = finalSubmissionIndex + 1
+
+                // If no matches were found, add this submission as a new unique one
+                // and emit an event
+                if finalSubmissionIndex == FlowDKG.uniqueFinalSubmissions.length {
+                    FlowDKG.uniqueFinalSubmissionCount[finalSubmissionIndex] = 0
+                    FlowDKG.uniqueFinalSubmissions.append(submission)
```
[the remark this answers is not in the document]
joshuahannan: I discovered this bug while writing tests for the contract and it is fixed now. 👍

contracts/epochs/FlowDKG.cdc
```diff
@@ -0,0 +1,274 @@
...
+            return false
+        }
+
+        var index = 0
+
+        // Check each key in the submiission to make sure that it matches
+        // the existing one
+        for key in submission {
+            // If a key length is incorrect, it is an invalid submission
+            if key.length != FlowDKG.submissionKeyLength {
```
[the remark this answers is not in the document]
joshuahannan: yep, I moved it. 👍

contracts/epochs/FlowDKG.cdc
joshuahannan: I was intending that the main epoch contract would poll this function every block and when it and the qc is complete, it would be able to end the epoch setup phase. We know it is complete when this returns an array instead of `nil`. Does that make sense?

contracts/epochs/FlowDKG.cdc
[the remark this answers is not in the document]
jordanschalm: Some extra context @tarakby: this is a regular event (not a service event) and does not bubble up to the protocol software / network. It's just accessible for anyone who wants to query it.

contracts/epochs/FlowDKG.cdc
joshuahannan: `self.nodeFinalSubmission.keys.length` is actually equal to `n` here because it is a dictionary that has an entry reserved for every node. It might be good to change the name of it though, to avoid confusion. I'll try to think of something

contracts/epochs/FlowDKG.cdc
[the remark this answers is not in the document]
tarakby: It's good that is doesn't notify the protocol/network. But since the info won't be useful for any party, it's maybe worth it to save the little overhead.

contracts/epochs/FlowDKG.cdc
tarakby: Ah okay, I got confused with the length of a node's submission. Thanks for clarifying.

contracts/epochs/FlowDKG.cdc
tarakby: Yes that makes sense 👌

tarakby (approved):

PR #112, opened 2021-01-28
Adds FlowFreeze contract

joshuahannan:
Adds FlowFreeze contract, which maintains a list of accounts that are frozen that an admin, the service account, controls.

The admin can freeze and unfreeze accounts at will.
There is a public function to get the list of frozen accounts

Adds some tests to test freezing and unfreezing


contracts/temporary/FlowFreeze.cdc
```diff
@@ -0,0 +1,47 @@
...
+
+    FlowFreeze
+
+    This contract gives an admin the ability to freeze/unfreeze accounts
+    by adding them to a list stored in the contract.
+
+    If the account is on the frozen list, it will not be able to send transactions
+    and any transaction that accesses its account state will fail.
+
+    The execution envorionment will automatically query this contract as part of every transaction.
```
robmyers, line 11:
```suggestion
    The execution environment will automatically query this contract as part of every transaction.
```

contracts/temporary/FlowFreeze.cdc
```diff
@@ -0,0 +1,47 @@
...
+            FlowFreeze.freezeList[address] = true
+        }
+
+        pub fun unfreezeAccount(_ address: Address) {
+            FlowFreeze.freezeList.remove(key: address)
+        }
+    }
+
+    pub fun getFreezeList(): {Address: Bool} {
+        return FlowFreeze.freezeList
```
robmyers, line 37: Copy-on-write makes this efficient, I need to train myself out of being wary of returning composite types. :-)

robmyers (approved):
There's a typo in a comment but other than that this looks good technically. 👍

contracts/temporary/FlowFreeze.cdc
turbolent: Just to prevent misunderstandings: There's no copy-on-write optimization yet, but that shouldn't matter when writing contracts

robmyers: @turbolent Ah OK, thank you! Returning this map and then searching it in a script (as below) should still be almost as efficient as searching it in a function in the contract, though?

contracts/temporary/FlowFreeze.cdc
```diff
@@ -0,0 +1,47 @@
...
+pub contract FlowFreeze {
+
+    access(account) var freezeList: {Address: Bool}
+
+    pub let AdminStoragePath: StoragePath
+
+    pub resource Admin {
+
+        pub fun freezeAccount(_ address: Address) {
+            FlowFreeze.freezeList[address] = true
```
Kay-Zee, line 28: do we need to add: https://github.com/onflow/flow-go/pull/329/files#diff-e93a1f3993224e399ffe091fd8e6537b5cfcf0350249ea2a621096d4d01b4d59R61

```
setAccountFrozen(0x%s, true)
```

Kay-Zee:
is this still needed?

joshuahannan:
I was waiting for the freezing ability to be included in the emulator: https://github.com/onflow/flow-go/pull/329
before I moved forward with this, but I'm not sure if this is still the kind of contract we want. Do you know if we still need this?

Kay-Zee:
AFAIK, we can already freeze accounts with just a tx, without a contract, but maybe i'm missing something.

joshuahannan:
@Kay-Zee I think we thought it might be useful to have the frozen information easily accessible in a smart contract so it was more transparent and also emits events, but I don't know if those goals have changed? Who is in charge of this? Should we check with them?

joshuahannan:
@Kay-Zee @laynelafrance Do we still need this? This has been open for a long time and I'm not sure what we plan to do with it

justjoolz:
Who is the admin and under what circumstances can they Freeze a users
account?

Is this documented anywhere?

joshuahannan:
@justjoolz The admin is the service account, which already has been partially decentralized (4 different parties control keys for it other than dapper labs and the account needs two signatures to sign a transaction) This is only planned to be used if there is an account that is trying to do something obviously malicious to the network while we are still in the early non-BFT phases of Flow's development. Therefore, we don't expect to ever use it and really don't want to ever use it.

PR #117, opened 2021-02-12
Epoch Lifecycle contract

joshuahannan:
Adds the epoch lifecycle contract

Only one simple deployment test to make sure the contract checks and initializes properly, but it hasn't been tested at all yet besides that.


contracts/epochs/FlowEpoch.cdc
```diff
@@ -119,36 +131,458 @@ pub contract FlowEpochs {
...
+            array[randomIndex] = temp
+
+            i = i - 1
+        }
+
+        return array
+    } 
+
+
+    init (numViewsInEpoch: UInt64,
```
jordanschalm, line 492: The counter for the first epoch should be an argument here -- we need to be able to synchronize that between the protocol state and here when we're initializing.

We should also accept the `NodeInfo` list that was generated in the protocol bootstrapping, then we can fail if it doesn't match what we see in the staking contract.

More generally, we should accept each field of the `Setup` and `Commit` events here (even the ones that are usually computed like the DKG phase deadlines) and use those inputs as the source of truth for the first epoch. If the inputs are inconsistent (I think the only time this could happen is inputs related to the identity table which we need to check against the staking contract) then we should abort.

This is what I think the bootstrapping process should look like:
1. `Protocol Bootstrap Script` retrieves list of nodes for next epoch from staking contract
2. `Protocol Bootstrap Script` generates the root state for the spork using the node list from step 1 (this root state includes the `EpochSetup` and `EpochCommit` service events for the first post-spork epoch
3. `Protocol Bootstrap Script` initializes `FlowEpoch` contract, passing in the service events generated in step 2
4. `FlowEpoch` contracts initializes its internal state such that:
  * `currentEpochCounter` is the counter for the input service events
  * `currentEpochPhase` is `STAKING`
  * Epoch metadata for the current epoch is consistent with the input service events

contracts/epochs/FlowEpoch.cdc
```diff
@@ -119,36 +131,458 @@ pub contract FlowEpochs {
...
+        //                 DKGPhase1FinalView: stakingAuctionFinalView + self.numViewsInDKGPhase,
+        //                 DKGPhase2FinalView: stakingAuctionFinalView + (2 as UInt64 * self.numViewsInDKGPhase),
+        //                 DKGPhase3FinalView: stakingAuctionFinalView + (3 as UInt64 * self.numViewsInDKGPhase))
+
+
+        // emit EpochCommitted(counter: self.currentEpochCounter + 1 as UInt64,
+        //                     dkgPubKeys: dkgPubKeys,
+        //                     clusterQCs: clusterQCs)
+
+        let firstEpochMetadata = EpochMetadata(counter: self.currentEpochCounter,
```
jordanschalm, line 563: I'm not sure we need this -- is there some logic that assumes epoch metadata for the last epoch?

jordanschalm (commented):
Really awesome seeing this coming together, great work! 💯

jordanschalm:
Answering some of the outstanding questions here while they're fresh:

> Where can the random source come from?

It is safe to use the built-in `unsafeRandom` directly to generate the `RandomSource` for now.

> What node information can we share in the service events?

Ideally we only share the information necessary for the protocol software in service events. The purpose of service events is communication from the service account to the protocol software. Service events are included in full in execution receipts and ultimately in block payloads, so keeping the size down as much as possible is a priority as well.

If we can greatly simplify something by adding a small amount of not-strictly-necessary (to the protocol software) information to a service event, I'm OK with it. But in general we should think about these as network-internal communication.

> Where does the heartbeat function call come from?

Every block contains a system chunk which invokes a function in the service account. [This](https://github.com/onflow/flow-go/blob/1e0a30e7a9118aad96a706639e37ec955f562542/fvm/serviceaccount.go#L9-L17) is what that currently looks like (it just logs a message and doesn't do anything). This transaction will be updated to invoke the `Heartbeat` contract, which in turn does all the per-block housekeeping we need to do, including calling the epoch smart contract to check phase transitions. cc @ramtinms who knows more about this than I do 

> Initial bootstrapping

Hopefully I answered this in the review comments 😅

contracts/epochs/FlowEpoch.cdc
```diff
@@ -119,36 +131,458 @@ pub contract FlowEpochs {
...
+    pub resource Admin {
+        pub fun updateEpochLength(newEpochViews: UInt64) {
+            if FlowEpoch.currentEpochPhase != EpochPhase.STAKINGAUCTION {
+                return
+            }
+
+            FlowEpoch.numViewsInEpoch = newEpochViews
+        }
+
+        pub fun updateAuctionLength(newAuctionViews: UInt64) {
```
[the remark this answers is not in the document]
joshuahannan: Which function is it duplicating? `updateEpochLength` updates the length of an entire epoch, and `updateAuctionLength` only updates the length of the staking auction.

contracts/epochs/FlowEpoch.cdc
```diff
@@ -119,36 +131,458 @@ pub contract FlowEpochs {
...
+
+        let clusters = FlowEpochClusterQC.getClusters()
+
+        var clusterQCs: [String] = []
+
+        for cluster in clusters {
+            var votes: String = ""
+
+            for vote in cluster.votes {
+                if vote.raw! == cluster.votes[0].raw! {
```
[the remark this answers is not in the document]
joshuahannan: Good catch! I'm changing it to check for something else

contracts/epochs/FlowEpoch.cdc
```diff
@@ -55,19 +68,18 @@ pub contract FlowEpochs {
         //
         // Node IDs are hex-encoded 32-byte arrays. Public keys are encoded as
         // by the flow-go crypto library, then hex-encoded.
-        nodeIDs: [String],
-        nodeRoles: [UInt8],
-        nodeStakingPubKeys: [String],
-        nodeNetworkPubKeys: [String],
-        nodeNetworkAddresses: [String],
+        nodeInfo: [FlowIDTableStaking.NodeInfo]
```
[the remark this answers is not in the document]
joshuahannan: yeah I can change it to this also

contracts/epochs/FlowEpoch.cdc
```diff
@@ -119,36 +131,458 @@ pub contract FlowEpochs {
...
+    pub var numViewsInEpoch: UInt64
+
+    /// The number of views in the staking auction
+    pub var numViewsInStakingAuction: UInt64
+    
+    /// The number of views in each dkg phase
+    pub var numViewsInDKGPhase: UInt64
+
+    /// The number of collector clusters in each epoch
+    pub var numCollectorClusters: UInt16
```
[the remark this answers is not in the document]
joshuahannan: Good suggestions! I'll add all of this

contracts/epochs/FlowEpoch.cdc
```diff
@@ -119,36 +131,458 @@ pub contract FlowEpochs {
...
+
+    pub resource Heartbeat {
+        /// Function that is called every block to advance the epoch
+        pub fun advanceBlock(randomSource: String?, ) {
+
+            let currentBlock = getCurrentBlock()
+            let currentEpochMetadata = FlowEpoch.epochMetadata[FlowEpoch.currentEpochCounter]!
+
+            if FlowEpoch.currentEpochPhase == EpochPhase.STAKINGAUCTION {
+                let stakingAuctionFinalView = currentEpochMetadata.startView + FlowEpoch.numViewsInStakingAuction - 1 as UInt64
```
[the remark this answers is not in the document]
joshuahannan: What do you mean by a virtual field? Should it just be a field in `currentEpochMetadata`? Or are you saying it should be inferred? I kind of like having it as a clear enum so that it is easily queryable, but I understand that it might not be totally necessary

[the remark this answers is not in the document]
joshuahannan: Or do you mean `stakingAuctionFinalView`?

contracts/epochs/FlowEpoch.cdc
```diff
@@ -119,36 +131,458 @@ pub contract FlowEpochs {
...
+        // Start QC Voting with the supplied clusters
+        self.QCAdmin.startVoting(clusters: collectorClusters)
+
+        // Start DKG with the consensus nodes
+        self.DKGAdmin.startDKG(nodeIDs: consensusNodeIDs)
+
+        let proposedEpochMetadata = EpochMetadata(counter: self.currentEpochCounter + UInt64(1),
+                                                seed: randomSource,
+                                                startView: self.epochMetadata[self.currentEpochCounter]!.endView + UInt64(1),
+                                                endView: self.epochMetadata[self.currentEpochCounter]!.endView + UInt64(1) + self.numViewsInEpoch,
```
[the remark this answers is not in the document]
joshuahannan: Good catch! I'll remove the +1 from the end view

contracts/epochs/FlowEpoch.cdc
joshuahannan: no, I can get rid of it

contracts/epochs/FlowEpoch.cdc
joshuahannan: Do we actually need all the inputs for the service events in `init`? We decided that we aren't emitting the events in `init`, correct? So all we need is what is stored in `EpochMetadata`.

contracts/epochs/FlowEpoch.cdc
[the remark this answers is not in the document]
jordanschalm: Yeah, you're right. Sorry I don't know what I was thinking here 😓

contracts/epochs/FlowEpoch.cdc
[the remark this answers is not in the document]
jordanschalm: I was referring to the `stakingAuctionFinalView`. Maybe "virtual field" isn't the right term -- I mean making `stakingAuctionFinalView` a field/method that is computed only from fields in `EpochMetadata` and isn't dependent on mutable contract-global state.

More generally I think properties of a particular epoch should be computed from an immutable epoch-specific configuration rather than the current global configuration, since we allow the global configuration to be changed over time.

Concretely, instead of doing:
```
let stakingAuctionFinalView = currentEpochMetadata.startView + FlowEpoch.numViewsInStakingAuction - 1 as UInt64
```

we would do something like:
```
pub struct EpochMetadata {
    pub fun stakingAuctionFinalView() UInt64 {
        return self.startView + self.config.numViewsInStakingAuction - 1 as UInt64
    }
// ...
}
```

contracts/epochs/FlowEpoch.cdc
jordanschalm: Yes good point. We only need what is stored in `EpochMetadata`

contracts/epochs/FlowEpoch.cdc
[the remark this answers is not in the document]
joshuahannan: oh yeah that makes sense. I can change it to something like that. We could just make it another field in the epoch metadata struct. Would that make more sense, so it would be immutable?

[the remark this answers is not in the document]
jordanschalm: Yeah, a literal field that's computed when we construct the `EpochMetadata` instance, or a computed property based on immutable fields of `EpochMetadata`, either one works for me 👍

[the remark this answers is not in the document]
joshuahannan: ok. I made it a immutable field of epochMetadata

contracts/epochs/FlowEpoch.cdc
```diff
@@ -55,19 +68,18 @@ pub contract FlowEpochs {
...
+        firstView: UInt64,
 
         // The last view (inclusive) of the upcoming epoch.
         finalView: UInt64,
 
         // The cluster assignment for the upcoming epoch. Each element in the list
         // represents one cluster and contains all the node IDs assigned to that
         // cluster, separated by commas.
-        collectorClusters: [String],
+        collectorClusters: [FlowEpochClusterQC.Cluster],
```
bjartek, line 82: maybe adjust the comment here? This is no longer a string csv right?

contracts/epochs/FlowEpoch.cdc
bjartek, line 71: Adjust the comment here, this now contains more then just ids.

contracts/epochs/FlowEpoch.cdc
```diff
@@ -94,13 +106,13 @@ pub contract FlowEpochs {
...
         // TODO: define ordering
-        // TODO: which is group public key
+        // Group public key is the last element
         dkgPubKeys: [String],
 
         // The result of the QC aggregation process. Each element contains all the votes
         // received for a particular cluster, comma-separated.
         // TODO: define ordering
-        clusterQCs: [String]
+        clusterQCs: [FlowEpochClusterQC.ClusterQC]
```
bjartek, line 115: adjust the comment here, what is the difference between collectorsClusters  and clusterQCs

contracts/epochs/FlowEpoch.cdc
```diff
@@ -109,46 +121,511 @@ pub contract FlowEpochs {
...
+        pub let stakingEndView: UInt64
+
         /// The organization of collector node IDs into clusters
         /// determined by a round robin sorting algorithm
-        pub let collectorClusters: [EpochClusterQCs.Cluster]
+        pub let collectorClusters: [FlowEpochClusterQC.Cluster]
 
         /// The Quorum Certificates from the ClusterQC contract
-        pub var clusterQCs: Int
+        pub var clusterQCs: [FlowEpochClusterQC.ClusterQC]
```
bjartek, line 140: adjust comment?

contracts/epochs/FlowEpoch.cdc
```diff
@@ -109,46 +121,511 @@ pub contract FlowEpochs {
...
+            pre {
+                FlowEpoch.currentEpochPhase == EpochPhase.STAKINGAUCTION: "Can only update fields during the staking auction"
+                newEpochViews > FlowEpoch.configurableMetadata.numViewsInStakingAuction: "Epoch Views must be greater than staking views"
+                newEpochViews > FlowEpoch.configurableMetadata.numViewsInDKGPhase: "Epoch Views must be greater than DKG phase views"
+            }
+
+            FlowEpoch.configurableMetadata.numViewsInEpoch = newEpochViews
+        }
+
+        pub fun updateAuctionViews(_ newAuctionViews: UInt64) {
```
bjartek, line 235: I really like the way you have structured these methods, It makes it easy to reason about the state machine.

contracts/epochs/FlowEpoch.cdc
```diff
@@ -109,46 +121,511 @@ pub contract FlowEpochs {
...
+
+    pub resource Heartbeat {
+
+        /// Function that is called every block to advance the epoch
+        pub fun advanceBlock() {
+
+            let currentBlock = getCurrentBlock()
+            let currentEpochMetadata = FlowEpoch.epochMetadata[FlowEpoch.currentEpochCounter]!
+
+            if FlowEpoch.currentEpochPhase == EpochPhase.STAKINGAUCTION {
```
bjartek, line 272: This is a method with lots of complexity. At first I wondered why the ifs where not colleapsed into one, why there where two levels, but then I got the nueance of it. 

Have you considered returning and not using else ifs for the other states? switch/case statement on enums would have made this code a lot easier to read.

contracts/epochs/FlowEpoch.cdc
```diff
@@ -109,46 +121,511 @@ pub contract FlowEpochs {
...
+
+        // Get nodeinfo for all the nodes
+        // get all the collector and consensus nodes
+        // to initialize the QC and DKG
+        for id in ids {
+            let nodeInfo = FlowIDTableStaking.NodeInfo(nodeID: id)
+
+            nodeInfoArray.append(nodeInfo)
+
+            if nodeInfo.role == 1 as UInt8 {
```
bjartek, line 373: Have you considered using enums for the role and not just a number? Or at least assign the number as a contant and use that as a variable here. For me this flags as 'magical number'

contracts/epochs/FlowEpoch.cdc
```diff
@@ -109,46 +121,511 @@ pub contract FlowEpochs {
...
+        while clusterIndex < self.configurableMetadata.numCollectorClusters {
+            clusters.append(FlowEpochClusterQC.Cluster(index: clusterIndex, nodeWeights: nodeWeightsDictionary[clusterIndex]!))
+        }
+
+        return clusters
+
+    }
+  
+    /// A function to generate a random permutation of arr[] 
+    /// using the fisher yates shuffling algorithm
```
bjartek, line 510: love that you reference the algorithm here.

bjartek (commented):
You asked for reviews in the flow discord so I took a stab at it. As you said testing this will be hard. You almost have to set up a new chain just to test this IMHO. 

Have you looked at the posibility of setting up entire new networks using IaC principles? Spin up a bunch of nodes on a provider and have a short lived real cluster?

contracts/epochs/FlowEpoch.cdc
joshuahannan: yes, it used to be a string, but we updated it. I'll fix the comment

contracts/epochs/FlowEpoch.cdc
joshuahannan: Yeah, I'll probably use `switch` instead. I keep forgetting that exists!

contracts/epochs/FlowEpoch.cdc
joshuahannan: The roles are defined in the staking contract, which was written before enums we possible in Cadence, and upgrading state in the contract is very unsafe, so were stuck with `UInt8` for now. Do you think it would be useful to define an enum in this contract for it and just convert for everything in here?

bjartek: I think so yes, it makes the code here easier to understand. And it is already a handful :)

contracts/epochs/FlowEpoch.cdc
bjartek: Oh Bastian implemented that, I did not know :) Wonderful.

contracts/epochs/FlowEpoch.cdc
```diff
@@ -109,46 +121,511 @@ pub contract FlowEpochs {
...
+            FlowEpoch.configurableMetadata.numCollectorClusters = newNumClusters
+        }
+    }
+
+    pub resource Heartbeat {
+
+        /// Function that is called every block to advance the epoch
+        pub fun advanceBlock() {
+
+            let currentBlock = getCurrentBlock()
```
jordanschalm, line 269: We might want to re-structure this slightly so we can inject the current block in tests, unless the emulator makes it easy enough to do that.

contracts/epochs/FlowEpoch.cdc
joshuahannan: ok I'll do that! 👍

contracts/epochs/FlowEpoch.cdc
```diff
@@ -166,24 +171,42 @@ pub contract FlowEpoch {
...
+    access(contract) let configurableMetadata: Config
+
+    /// Metadata that is managed by the smart contract
+    /// and cannot be changed by the Admin
+
     /// The counter, or ID, of the current epoch
     pub var currentEpochCounter: UInt64
 
+    /// The counter (ID) for the next epoch
+    pub var proposedEpochCounter: UInt64
```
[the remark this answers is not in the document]
joshuahannan: Good point. Yeah that should be a function

contracts/epochs/FlowEpoch.cdc
```diff
@@ -109,46 +128,551 @@ pub contract FlowEpochs {
...
+
+        emit EpochSetup(counter: proposedEpochMetadata.counter,
+                        nodeInfo: nodeInfoArray, 
+                        firstView: proposedEpochMetadata.startView,
+                        finalView: proposedEpochMetadata.endView,
+                        collectorClusters: collectorClusters,
+                        randomSource: randomSource,
+                        DKGPhase1FinalView: proposedEpochMetadata.startView + self.configurableMetadata.numViewsInStakingAuction + self.configurableMetadata.numViewsInDKGPhase,
+                        DKGPhase2FinalView: proposedEpochMetadata.startView + self.configurableMetadata.numViewsInStakingAuction + (2 as UInt64 * self.configurableMetadata.numViewsInDKGPhase),
+                        DKGPhase3FinalView: proposedEpochMetadata.startView + self.configurableMetadata.numViewsInStakingAuction + (3 as UInt64 * self.configurableMetadata.numViewsInDKGPhase))
```
arrivets, line 451: I believe these are off by 1
```suggestion
                        DKGPhase1FinalView: proposedEpochMetadata.startView + self.configurableMetadata.numViewsInStakingAuction + self.configurableMetadata.numViewsInDKGPhase -1 as UInt64,
                        DKGPhase2FinalView: proposedEpochMetadata.startView + self.configurableMetadata.numViewsInStakingAuction + (2 as UInt64 * self.configurableMetadata.numViewsInDKGPhase) - 1 as UInt64,
                        DKGPhase3FinalView: proposedEpochMetadata.startView + self.configurableMetadata.numViewsInStakingAuction + (3 as UInt64 * self.configurableMetadata.numViewsInDKGPhase) -1 as UInt64)
```

joshuahannan: yes, I believe you are correct. Good catch!

joshuahannan:
@jordanschalm I changed the order of the `EpochCommitted` event fields so that it follows the order of the rest of the contract, (QC first, DKG second)
```
    pub event EpochCommitted(
        counter: UInt64,
        clusterQCs: [FlowEpochClusterQC.ClusterQC]
        dkgPubKeys: [String],
    )
```
I just wanted to flag that in case there was any node software that needs to be updated.

lib/go/test/flow_epoch_test.go
```diff
@@ -0,0 +1,795 @@
...
+
+	// create new user accounts, mint tokens for them, and register them for staking
+	addresses, signers := registerAndMintManyAccounts(t, b, accountKeys, numEpochAccounts)
+	ids, _, _ := generateNodeIDs(numEpochAccounts)
+	registerNodesForStaking(t, b, env,
+		addresses,
+		signers,
+		ids)
+
+	t.Run("Should not be able to register a QC voter or DKG participant if the staking auction isn't over", func(t *testing.T) {
```
jordanschalm, line 430: I'm not sure this will work with the operator onboarding flow we discussed 🤔. We wanted to:
* submit the staking request
* create the node operator's machine account
* retrieve the DKGParticipant or QCVoter resource and insert it to the machine account

all at the same time (ideally in one transaction) through Flow Port. So we would need to allow retrieving the QC/DKG resources during staking auction, right?

jordanschalm (approved):

contracts/epochs/FlowEpoch.cdc
```diff
@@ -109,46 +128,542 @@ pub contract FlowEpochs {
...
         }
     }
 
-    init () {
-        self.currentEpochCounter = 0
+    /// Pays rewards, moves staking tokens between buckets,
+    /// and starts the new epoch staking auction
+    access(account) fun startNewEpoch() {
+
+        self.stakingAdmin.payRewards()
```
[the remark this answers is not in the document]
joshuahannan: Good point. We'll probably need to put the total rewards calculation in this contract actually.

lib/go/test/flow_epoch_test.go
joshuahannan: We can't allow retrieving the resources during the staking auction because those can only be retrieved by approved nodes which doesn't happen until the end of the staking auction. As long as the node operator controls both accounts, then we should be able to do it.

joshuahannan:
@jordanschalm I am just going to merge this to the feature branch now and create new branches with my changes for rewards, registrations, and approvals

PR #118, opened 2021-03-02
add settable object fields to locked tokens contract

joshuahannan:


contracts/LockedTokens.cdc
```diff
@@ -99,11 +99,11 @@ pub contract LockedTokens {
 
         /// Optional NodeStaker resource. Will only be filled if the user
         /// signs up to be a node operator
-        pub var nodeStaker: @FlowIDTableStaking.NodeStaker?
+        pub(set) var nodeStaker: @FlowIDTableStaking.NodeStaker?
 
         /// Optional NodeDelegator resource. Will only be filled if the user
         /// signs up to be a delegator
-        pub var nodeDelegator: @FlowIDTableStaking.NodeDelegator?
+        pub(set) var nodeDelegator: @FlowIDTableStaking.NodeDelegator?
```
turbolent, line 106: Why make these fields publicly writeable? Isn't that a security problem?

contracts/LockedTokens.cdc
```diff
@@ -373,10 +373,19 @@ pub contract LockedTokens {
...
 
+        access(self) fun nodeObjectExists(_ managerRef: &LockedTokenManager): Bool {
+            return managerRef.nodeStaker != nil
+        }
+
         /// Stakes new locked tokens
         pub fun stakeNewTokens(amount: UFix64) {
             let tokenManagerRef = self.tokenManager.borrow()!
 
+            assert (
```
turbolent, line 384: Just my OCD: Is there a reason for putting a space inbetween? `assert` is just a function like any other

contracts/LockedTokens.cdc
joshuahannan: This object is only in the locked account, so it can only be accessed via a transaction co-signed by the user and the token admin. We need it right now to fix some accounts that were compromised by an operation we performed recently

contracts/LockedTokens.cdc
joshuahannan: I guess because I kind of view it as a function body instead of an argument list, so I treat it like a curly brace. but you're right that it should just be right after

psiemens (approved):
Logic looks sound to me.

Did you consider adding a method like `LockedTokenManager.removeDelegator`? Or does that cause breaking changes during the upgrade?

Kay-Zee (approved):

joshuahannan:
@psiemens That wouldn't cause problems during the upgrade, but the functionality wouldn't be any different. Do you think it would be better?

contracts/LockedTokens.cdc
turbolent: Sounds good, thank you for the explanation and context 👍

turbolent:
I quite like @psiemens's idea! Is setting  delegator and staker to `nil` all that is required right now? Then maybe adding remove functions for delegator and staker is better than allowing the two fields to be set to anything

turbolent (approved):

joshuahannan:
Good suggestion @psiemens! I just pushed some changes with that method instead of the way I was doing it. 😃

robmyers (approved):
👍  This looks good.

contracts/LockedTokens.cdc
```diff
@@ -224,6 +224,18 @@ pub contract LockedTokens {
 
             emit LockedAccountRegisteredAsDelegator(address: self.owner!.address, nodeID: nodeID)
         }
+
+        pub fun removeNode(): @FlowIDTableStaking.NodeStaker {
+            let node <- self.nodeStaker <- nil
+
+            return <-node!
```
turbolent, line 231: Maybe just make this function return an optional, and not force-unwrap here? 
Then in case the field is already nil, the function is just a "no-op" instead of aborting.

turbolent (approved):

PR #119, opened 2021-03-04
FEATURE: Staking Collection contract

joshuahannan:
## Summary

A Staking Collection is a resource that allows its owner to manage multiple staking
objects in a single account via a single storage path, and perform staking and delegation actions using both locked and unlocked Flow.

Before the staking collection, accounts could use the instructions in [the unlocked staking guide](https://docs.onflow.org/staking/unlocked-staking-guide/)
to stake with tokens. This was a bit restrictive, because that guide (and the corresponding transactions) only supports one node and one delegator object
per account. If a user wanted to have more than one per account, they would either have to use custom transactions with custom storage paths for each object,
or they would have had to use multiple accounts, which comes with many hassles of its own.

The same applies to the [locked tokens staking guide](https://docs.onflow.org/staking/locked-staking-guide/).
We only built in support for one node and one delegator per account.

The staking collection is a solution to both of these deficiencies. When an account is set up to use a staking collection,
the staking collection recognizes the existing locked account capabilities (if they exist) and unlocked account staking objects,
and incorporates their functionality so any user can stake for a node or delegator through a single common interface, regardless of
if they have a brand new account, or have been staking through the locked account or unlocked account before.

## Do I have to upgrade?

For those of you that are service providers who provide a staking service, this improvement is **completely optional**.  You can continue managing staking exactly the same as you are currently doing it and it will be totally fine and behave the same as before.

The Flow team will be performing this upgrade for Flow Port though, so if you provide a staking service for ledger or blocto users, you might need to upgrade to this if you want to give them access to the entire functionality of their account if they are also using Flow Port. If their entire interaction with staking is through Flow Port, then all the changes are handled for them and there is nothing for you to worry about.

## Technical features

I would recommend familiarizing yourself with the [staking contract](https://docs.onflow.org/staking) and [locked tokens contract](https://docs.onflow.org/flow-token/locked-account-setup/), because knowledge of how those work will help you understand these upgrades.

* The staking collection contract stores [a dictionary of staking objects](https://github.com/onflow/flow-core-contracts/blob/josh/proxycollection/contracts/FlowStakingCollection.cdc#L68) from the staking contract that are used to manage the stakers tokens. Since they are dictionaries, there can be as many node or delegator objects per account as the user wants. 
* The resource only has one set of staking methods, which route the call to the correct staking object based on the arguments that the caller specifies. (nodeID, delegatorID)
* The contract also stores an [optional capability to the locked token vault](https://github.com/onflow/flow-core-contracts/blob/josh/proxycollection/contracts/FlowStakingCollection.cdc#L63) and [locked tokens `TokenHolder` resource](https://github.com/onflow/flow-core-contracts/blob/josh/proxycollection/contracts/FlowStakingCollection.cdc#L73). This is only used if the user already has a locked account. The staking collection does not change the locked account setup at all, it only has access to it and to the locked vault.
* The collection makes the staking objects and vault capability fields private, because since it has access to the locked tokens, it needs to mediate access to the staking objects so users cannot withdraw tokens that are still locked from the sale. The resource has fields `unlockedTokensUsed` and `lockedTokensUsed`, to keep track of how many locked and unlocked tokens are being used for staking in order to allow the user to withdraw the correct amount when they choose to.
* The staking collection contract is a brand new contract that will be deployed to the same account as the existing locked tokens contract. A few of the fields on the `LockedTokens` contract have been updated to have `access(account)` visibility instead of `access(self)` because the staking collection contract needs to be able to access to them in order to work properly.
* We also included a public interface and getters in the contract so you can easily query it with an address to get node or delegator information from a collection. 

Looking for feedback on design decisions, implementation details, any events that would be useful to include in the contract, and whatever you feel is important!

We intend for this to be the method that all Flow Port users (ledger, blocto, etc) use for the forseeable future. When we enable it in Flow Port, we will ask every user to run a transaction to set up their account to use the staking collection from then on.


JeffreyDoyle (approved):

contracts/FlowStakingCollection.cdc
```diff
@@ -0,0 +1,844 @@
...
+        /// Operations to register new staking objects
+
+        /// Function to register a new Staking Record to the Staking Collection
+        pub fun registerNode(id: String, role: UInt8, networkingAddress: String, networkingKey: String, stakingKey: String, amount: UFix64) {
+
+            let tokens <- self.getTokens(amount: amount)
+
+            let nodeStaker <- FlowIDTableStaking.addNodeRecord(id: id, role: role, networkingAddress: networkingAddress, networkingKey: networkingKey, stakingKey: stakingKey, tokensCommitted: <-tokens)
+
+            //emit NewNodeCreated(nodeID: nodeStaker.id, role: nodeStaker.id, amountCommitted: amount)
```
turbolent, line 335: 1. Line is commented out
2. `role: nodeStaker.id` looks like a typo

contracts/FlowStakingCollection.cdc
```diff
@@ -0,0 +1,844 @@
...
+            pre {
+                self.doesStakeExist(nodeID: nodeID, delegatorID: nil): "Specified node does not exist"
+                self.lockedTokensUsed == UFix64(0.0): "Cannot remove node if locked tokens are used"
+            }
+            
+            if self.nodeStakers[nodeID] != nil {
+                let stakingInfo = FlowIDTableStaking.NodeInfo(nodeID: nodeID)
+                let totalStaked = stakingInfo.tokensStaked + stakingInfo.tokensCommitted + stakingInfo.tokensUnstaking + stakingInfo.tokensUnstaked
+
+                /// Since the NodeStaker object is being removed, the total number of unlocked tokens staked to it is deducted from the counter.
```
turbolent, line 279: Only use doc comments (`///`) for declaration documentation, i.e. switch to normal comment (`//`)

contracts/FlowStakingCollection.cdc
```diff
@@ -0,0 +1,844 @@
...
+                }
+
+                return self.borrowNode(nodeID) != nil
+            }
+        }
+
+        /// Function to add an existing NodeStaker object
+        pub fun addNodeObject(_ node: @FlowIDTableStaking.NodeStaker) {
+            let stakingInfo = FlowIDTableStaking.NodeInfo(nodeID: node.id)
+            let totalStaked = stakingInfo.tokensStaked + stakingInfo.tokensCommitted + stakingInfo.tokensUnstaking + stakingInfo.tokensUnstaked
```
turbolent, line 254: This summing (`tokensStaked + tokensCommitted + tokensUnstaking + tokensUnstaked`) is repeated many times in the contract. Consider adding a function and using it here and in the other places

contracts/FlowStakingCollection.cdc
```diff
@@ -0,0 +1,844 @@
...
+                if let delegator = self.borrowDelegator(nodeID, _delegatorID) {
+                    // We update the unlocked tokens used field before withdrawing because 
+                    // rewards are newly minted and not immediately reflected in the tokens used fields
+                    self.unlockedTokensUsed = self.unlockedTokensUsed + amount
+
+                    let tokens <- delegator.withdrawRewardedTokens(amount: amount)
+
+                    self.depositTokens(from: <-tokens)
+                } else {
+                    let delegator = self.tokenHolder!.borrow()!.borrowDelegator()
```
turbolent, line 558: Maybe refactor the repeated part `self.tokenHolder!.borrow()!` out into a variable

contracts/FlowStakingCollection.cdc
```diff
@@ -0,0 +1,844 @@
...
+        // Closers
+
+        /// Closes an existing stake or delegation, moving all withdrawable tokens back to the users account and removing the stake
+        /// or delegator object from the StakingCollection.
+        pub fun closeStake(nodeID: String, delegatorID: UInt32?) {
+            pre {
+                self.doesStakeExist(nodeID: nodeID, delegatorID: delegatorID): "Specified stake does not exist"
+            }
+
+            if let _delegatorID = delegatorID {
```
turbolent, line 594: Maybe just shadow the optional variable with the non-optional one, i.e. name it the same:

```suggestion
            if let delegatorID = delegatorID {
```

contracts/FlowStakingCollection.cdc
```diff
@@ -0,0 +1,844 @@
...
+                }
+            } else {
+                let stakeInfo = FlowIDTableStaking.NodeInfo(nodeID: nodeID)
+
+                assert(
+                    stakeInfo.tokensStaked + stakeInfo.tokensCommitted + stakeInfo.tokensUnstaking == 0.0,
+                    message: "Cannot close a stake until all tokens have been withdrawn, or moved to a withdrawable state."
+                )
+
+                if (stakeInfo.tokensUnstaked > 0.0) {
```
turbolent, line 630: Style: no need for parentheses (here and below):

```suggestion
                if stakeInfo.tokensUnstaked > 0.0 {
```

contracts/FlowStakingCollection.cdc
```diff
@@ -0,0 +1,844 @@
...
+                    destroy staker
+                } else {
+                    if let tokenHolderCapability = self.tokenHolder {
+                        let tokenManager = tokenHolderCapability.borrow()!.borrowTokenManager()
+                        let staker <- tokenManager.removeNode()
+                        destroy staker
+                    } else {
+                        panic("Token Holder capability needed and not found.")
+                    }
+                }
```
turbolent, line 649: Consider merging the nested `if` in the `else` branch, i.e. use an `else if`: 

```suggestion
                } else if let tokenHolderCapability = self.tokenHolder {
                    let tokenManager = tokenHolderCapability.borrow()!.borrowTokenManager()
                    let staker <- tokenManager.removeNode()
                    destroy staker
                } else {
                    panic("Token Holder capability needed and not found.")
                }
```

contracts/FlowStakingCollection.cdc
```diff
@@ -0,0 +1,844 @@
...
+            }
+        }
+
+        // Getters
+
+        /// Function to get all node ids for all Staking records in the StakingCollection
+        pub fun getNodeIDs(): [String] {
+            let nodeIDs: [String] = self.nodeStakers.keys
```
turbolent, line 657: No need for type annotation (here and below):

```suggestion
            let nodeIDs = self.nodeStakers.keys
```

contracts/FlowStakingCollection.cdc
```diff
@@ -0,0 +1,844 @@
...
+        access(self) fun borrowNode(_ nodeID: String): &FlowIDTableStaking.NodeStaker? {
+            if self.nodeStakers[nodeID] != nil {
+                return &self.nodeStakers[nodeID] as? &FlowIDTableStaking.NodeStaker
+            } else {
+                return nil
+            }
+        }
+
+        /// Borrows a reference to a delegator in the collection
+        access(self) fun borrowDelegator(_ nodeID: String, _ delegatorID: UInt32): &FlowIDTableStaking.NodeDelegator? {
```
turbolent, line 368: I'd recommend using argument labels here, this is exactly what they are for: Here, two IDs are passed, but in a function call arguments get mixed up:
```suggestion
        access(self) fun borrowDelegator(nodeID: String, delegatorID: UInt32): &FlowIDTableStaking.NodeDelegator? {
```

I'd also add them for consistency, `removeDelegator` has the argument labels

contracts/FlowStakingCollection.cdc
```diff
@@ -0,0 +1,844 @@
...
+                if idInfo.delegatorNodeID == nodeID { panic("Cannot register a delegator for a node that is already being delegated to") }
+            }
+            
+            let tokens <- self.getTokens(amount: amount)
+
+            let nodeDelegator <- FlowIDTableStaking.registerNewDelegator(nodeID: nodeID)
+
+            nodeDelegator.delegateNewTokens(from: <- tokens)
+
+            //emit NewDelegatorCreated(nodeID: nodeDelegator.nodeID, delegatorID: nodeDelegator.id, amountCommitted: amount)
```
turbolent, line 353: Commented out code

contracts/FlowStakingCollection.cdc
```diff
@@ -0,0 +1,844 @@
...
+            //emit NewNodeCreated(nodeID: nodeStaker.id, role: nodeStaker.id, amountCommitted: amount)
+
+            self.nodeStakers[id] <-! nodeStaker
+        }
+
+        /// Function to register a new Delegator Record to the Staking Collection
+        pub fun registerDelegator(nodeID: String, amount: UFix64) {
+            let delegatorIDs = self.getDelegatorIDs()
+            for idInfo in delegatorIDs {
+                if idInfo.delegatorNodeID == nodeID { panic("Cannot register a delegator for a node that is already being delegated to") }
```
turbolent, line 344: Long line, consider wrapping:
```suggestion
                if idInfo.delegatorNodeID == nodeID {
                	panic("Cannot register a delegator for a node that is already being delegated to")
                }
```

contracts/FlowStakingCollection.cdc
```diff
@@ -0,0 +1,844 @@
...
+                
+                // If not all can be removed from locked, remove what can be, then remove the rest from unlocked
+                } else {
+
+                    // update locked tokens used record by adding the rest of the locked balance
+                    self.lockedTokensUsed = self.lockedTokensUsed + lockedBalance
+
+                    // Update the unlocked tokens used record by adding the amount requested
+                    // minus whatever was used from the locked tokens
+                    self.unlockedTokensUsed = self.unlockedTokensUsed + (amount - lockedBalance)
```
turbolent, line 157: Repeated calculation of `amount - lockedBalance` here and below, consider declaring a variable and using it in both places

contracts/FlowStakingCollection.cdc
```diff
@@ -0,0 +1,844 @@
...
+
+        /// Called when committing tokens for staking. Gets tokens from either or both vaults
+        /// Uses locked tokens first, then unlocked if any more are still needed
+        access(self) fun getTokens(amount: UFix64): @FungibleToken.Vault {
+
+            // If there is a locked account, use the locked vault first
+            if self.lockedVault != nil {
+
+                var lockedBalance: UFix64 = self.lockedVault!.borrow()!.balance - FlowStorageFees.minimumStorageReservation
+                var unlockedBalance: UFix64 = self.unlockedVault.borrow()!.balance - FlowStorageFees.minimumStorageReservation
```
turbolent, line 136: Consider declaring a variable for the repeated `self.lockedVault!.borrow()!` and `self.unlockedVault.borrow()!` here and below (for the withdrawal).

Also, the variable declarations neither need to be mutable nor need a type annotation:

```suggestion
				let lockedVault = self.lockedVault!.borrow()!
				let unlockedVault = self.unlockedVault.borrow()!

                let lockedBalance = lockedVault.balance - FlowStorageFees.minimumStorageReservation
                let unlockedBalance = unlockedVault.balance - FlowStorageFees.minimumStorageReservation
```

contracts/FlowStakingCollection.cdc
```diff
@@ -0,0 +1,844 @@
...
+
+                    // Deposit them into the same vault
+                    tokens.deposit(from: <-lockedPortion)
+                    tokens.deposit(from: <-unlockedPortion)
+
+                    return <-tokens
+                }
+            } else {
+                // Since there is no locked account, all tokens have to come from the normal unlocked balance
+                var unlockedBalance: UFix64 = self.unlockedVault.borrow()!.balance - FlowStorageFees.minimumStorageReservation
```
turbolent, line 173:
```suggestion
                let unlockedBalance = unlockedVault.balance - FlowStorageFees.minimumStorageReservation
```

contracts/FlowStakingCollection.cdc
```diff
@@ -0,0 +1,844 @@
...
+
+                assert(
+                    amount <= unlockedBalance,
+                    message: "Insufficient total Flow balance"
+                )
+
+                self.unlockedTokensUsed = self.unlockedTokensUsed + amount
+
+                return <-self.unlockedVault.borrow()!.withdraw(amount: amount)
```
turbolent, line 182:
```suggestion
                return <-unlockedVault.withdraw(amount: amount)
```

contracts/FlowStakingCollection.cdc
```diff
@@ -0,0 +1,844 @@
...
+
+                    // Update the unlocked tokens used record by adding the amount requested
+                    // minus whatever was used from the locked tokens
+                    self.unlockedTokensUsed = self.unlockedTokensUsed + (amount - lockedBalance)
+
+                    let tokens <- FlowToken.createEmptyVault()
+
+                    // Get the actual tokens from each vault
+                    let lockedPortion <- self.lockedVault!.borrow()!.withdraw(amount: lockedBalance)
+                    let unlockedPortion <- self.unlockedVault.borrow()!.withdraw(amount: amount - lockedBalance)
```
turbolent, line 163:
```suggestion
                    let lockedPortion <- lockedVault.withdraw(amount: lockedBalance)
                    let unlockedPortion <- unlockedVault.withdraw(amount: amount - lockedBalance)
```

contracts/FlowStakingCollection.cdc
```diff
@@ -0,0 +1,844 @@
...
+            var tokenHolderNodeID: String? = nil
+            var tokenHolderDelegatorNodeID: String? = nil
+            var tokenHolderDelegatorID: UInt32?  = nil
+
+            // If there is a locked account, get the staking info from that account
+            if self.tokenHolder != nil {
+                if let _tokenHolder = self.tokenHolder!.borrow() {
+                    tokenHolderNodeID = _tokenHolder!.getNodeID()
+                    tokenHolderDelegatorNodeID = _tokenHolder!.getDelegatorNodeID()
+                    tokenHolderDelegatorID = _tokenHolder!.getDelegatorID()
```
turbolent, line 230: Due to the optional binding (if-let), `_tokenHolder` should be non-optional here:

```suggestion
                    tokenHolderNodeID = _tokenHolder.getNodeID()
                    tokenHolderDelegatorNodeID = _tokenHolder.getDelegatorNodeID()
                    tokenHolderDelegatorID = _tokenHolder.getDelegatorID()
```

contracts/FlowStakingCollection.cdc
```diff
@@ -0,0 +1,844 @@
...
+                }
+            }
+
+            // If the request is for a delegator, check all possible delegators for possible matches
+            if let _delegatorID = delegatorID {
+                if (tokenHolderDelegatorNodeID != nil && tokenHolderDelegatorID != nil && tokenHolderDelegatorNodeID! == nodeID && tokenHolderDelegatorID! == _delegatorID) {
```
turbolent, line 236: Long line, consider wrapping this for each operator:
```suggestion
                if tokenHolderDelegatorNodeID != nil 
                	&& tokenHolderDelegatorID != nil 
                	&& tokenHolderDelegatorNodeID! == nodeID 
                	&& tokenHolderDelegatorID! == _delegatorID
                {
```

turbolent (commented):
I had a look through and added some recommendations for the Cadence code. I don't know anything regarding staking and the business logic.

Please feel free to request another review if you want it

joshuahannan:
@turbolent Thank you so much for the review! I've addressed your comments in https://github.com/onflow/flow-core-contracts/pull/188

bluesign:
Maybe I am missing something but is addNodeObject and addDelegatorObject being in the StakingCollectionPublic interface correct?

joshuahannan:
@bluesign That was intentional because we wanted to allow users to transfer their staking objects to another account if they wanted. Do you see a potential problem with that?

bluesign:
What if I overwrite someone staking object? 

PS: I don't have much information about staking contract tbh

PR #132, opened 2021-03-23
Add reward APY and separate rewards payment

### LINKED ISSUE #130: EPOCHS: Separate the pay rewards function from start new epoch
### Issue To Be Solved
Rewards might need to be paid during a different phase, so it shouldn't be part of the start new epoch function.

### Suggest A Solution

* Create new `payRewards` function for epochs that is separate
* Allow this function to be called by the epoch admin directly 
* Test transaction

### LINKED ISSUE #131: EPOCHS: Calculate Rewards when paying them and allow admin to control APY
### Issue To Be Solved
Need to be able to calculate new rewards payout every epoch and save as as part of epoch metadata

### Suggest A Solution
* Work off the `feature/epochs` branch
* Add a field to epoch metadata for the reward amount
* Add a field to configurable metadata for reward APY
* Admin can update APY field
* When paying rewards, calculate the new APY

joshuahannan:
Closes #130 
Closes #131 

* Adds a configurable field for the annual percentage yield for rewards, i.e. the annualized percentage of the total supply of FLOW that gets minted every epoch for rewards.
I have the actual calculation commented out for now because it will be manual at first

* Also separates rewards payment into a function that the admin calls manually, providing a new payout value when it is called for the next epoch

* Adds `totalRewards` field to epoch metadata to show the amount of rewards that are paid for that epoch

* Also added an index field to the `clusterQC` struct to show which cluster index it is for


contracts/epochs/FlowEpoch.cdc
```diff
@@ -192,11 +197,15 @@ pub contract FlowEpoch {
         /// The number of collector clusters in each epoch
         pub(set) var numCollectorClusters: UInt16
 
-        init(numViewsInEpoch: UInt64, numViewsInStakingAuction: UInt64, numViewsInDKGPhase: UInt64, numCollectorClusters: UInt16) {
+        /// Tracks the percentage of FLOW total supply that is minted as rewards at the end of an epoch
```
jordanschalm, line 200: To me the comment implies that this value is the per-epoch percentage, but [here](https://github.com/onflow/flow-core-contracts/pull/132/files#diff-1da2c6b9c85ee2833c92c917dd4f57655a9f505ff07e2d8088977b6bcdb1969bR395) we divide by 52. Can make it clearer in the comment that this is the annual inflation rate

lib/go/test/flow_epoch_test.go
```diff
@@ -196,6 +203,26 @@ func TestEpochPhaseMetadataChange(t *testing.T) {
...
+		tx = createTxWithTemplateAndAuthorizer(b, templates.GenerateUpdateRewardPercentageScript(env), idTableAddress)
+		_ = tx.AddArgument(CadenceUFix64("2.04"))
+		signAndSubmit(
+			t, b, tx,
+			[]flow.Address{b.ServiceKey().Address, idTableAddress},
+			[]crypto.Signer{b.ServiceKey().Signer(), IDTableSigner},
+			true,
+		)
+
+		// Should succeed because it is > 1
```
jordanschalm, line 216:
```suggestion
		// Should succeed because it is < 1
```

contracts/epochs/FlowEpoch.cdc
```diff
@@ -361,12 +384,25 @@ pub contract FlowEpoch {
         }
     }
 
-    /// Pays rewards, moves staking tokens between buckets,
-    /// and starts the new epoch staking auction
-    access(account) fun startNewEpoch() {
+    /// Pays rewards to staked nodes and delegators,
+    /// Calculates a new token payout, and sets the new payout
+    access(account) fun payRewards(_ newPayout: UFix64?) {
```
jordanschalm, line 389: With this change, it seems like we now provide the reward amount for _next_ epoch when paying out rewards for the _current_ epoch. Is that right? I'm a bit worried about the indirection and the change (I believe) from the current functionality for a manual process.

contracts/epochs/FlowEpoch.cdc
```diff
@@ -429,6 +465,7 @@ pub contract FlowEpoch {
                                                 startView: self.epochMetadata[self.currentEpochCounter]!.endView + UInt64(1),
                                                 endView: self.epochMetadata[self.currentEpochCounter]!.endView + self.configurableMetadata.numViewsInEpoch,
                                                 stakingEndView: self.epochMetadata[self.currentEpochCounter]!.endView + self.configurableMetadata.numViewsInStakingAuction,
+                                                totalRewards: 0.0 as UFix64,
```
jordanschalm, line 468: Should this be 0?

contracts/epochs/FlowEpoch.cdc
joshuahannan: yes, you are correct. The current functionality would stay the same until we decide to change it, because the line is commented out. How else would you recommend we do it when it is fully automated? To me this is the only logical place do calculate the rewards, because the rewards depend on how many tokens are actually rewarded in the current epoch, which could change because of arithmetic precision loss when calculating it for each staker. And we shouldn't calculate it  after the epoch starts, because the stakers should know, at the beginning of the epoch, the number of tokens that are going to get paid.

contracts/epochs/FlowEpoch.cdc
joshuahannan: Since the value can't be fully accurately calculated until the end of the epoch, I just set it as zero as a placeholder value and update it at the end of the epoch. Do you think we should do a rough calculation here also? Or set it to `nil` instead of `0.0`?

jordanschalm: Right that makes sense. The part l'm missing is where it's updated at the end of an epoch. I see we set it when [creating the first epoch](https://github.com/onflow/flow-core-contracts/pull/132/files#diff-1da2c6b9c85ee2833c92c917dd4f57655a9f505ff07e2d8088977b6bcdb1969bR701) and when [resetting an epoch](https://github.com/onflow/flow-core-contracts/pull/132/files#diff-1da2c6b9c85ee2833c92c917dd4f57655a9f505ff07e2d8088977b6bcdb1969bR378), but not sure when we'd set `totalRewards` to a non-zero value for epochs started with `startNewEpoch`

contracts/epochs/FlowEpoch.cdc
jordanschalm: > The current functionality would stay the same until we decide to change it

This might be the part I'm missing. I assumed we would would use this function for the manual payout as well. But I guess we would do something like:

```
stakingAdmin.setEpochTokenPayout(currentEpochPayout)
stakingAdmin.payRewards()
```

joshuahannan: We would use this function. This function behaves in exactly the same way that we do it now. We pay rewards, then we set the new payout for the next epoch. It just happens in the function in stead of in separate function calls

contracts/epochs/FlowEpoch.cdc
joshuahannan: You're right, that should have been included in the payRewards function. I'll add that and add a test for it

joshuahannan:
@jordanschalm How does this look now? I addressed all your comments in aea53b5

jordanschalm (approved):
Looks good 👌

PR #133, opened 2021-03-26
Un-restrict QC and DKG registration

joshuahannan:
Removes the requirement that QC and DKG registration has to be during EpochSetup.
Now it can happen at any time
Adds a transaction to register for staking and QC/DKG at the same time
Tests these changes


lib/go/test/epoch_test_helpers.go
```diff
@@ -235,14 +236,64 @@ func advanceView(
...
+	)
+
+	if !shouldFail {
+		newTokensCommitted = tokensCommitted.Plus(amount).(interpreter.UFix64Value)
+	}
+
+	return
+}
+
+/// Resters the specified number of nodes for staking and qc/dkg in the same transaction
```
jordanschalm, line 287:
```suggestion
/// Registers the specified number of nodes for staking and qc/dkg in the same transaction
```

lib/go/test/flow_dkg_test.go
```diff
@@ -63,27 +66,27 @@ func TestDKG(t *testing.T) {
...
 
-	t.Run("Should not be able to register a voter if the node hasn't been registered as a consensus node", func(t *testing.T) {
+	t.Run("Should be able to register a voter before the dkg is enabled", func(t *testing.T) {
 
-		tx := createTxWithTemplateAndAuthorizer(b, templates.GenerateCreateDKGParticipantScript(env), DKGAddress)
+		tx := createTxWithTemplateAndAuthorizer(b, templates.GenerateCreateDKGParticipantScript(env), jordanAddress)
 
 		_ = tx.AddArgument(cadence.NewAddress(DKGAddress))
-		_ = tx.AddArgument(cadence.NewString(adminID))
+		_ = tx.AddArgument(cadence.NewString(accessID))
```
jordanschalm, line 74: just checking this wasn't a bad find+replace -- not seeing where the variable definition changed

transactions/epoch/node/register_node.cdc
```diff
@@ -0,0 +1,79 @@
...
+
+        let nodeRef = acct.borrow<&FlowIDTableStaking.NodeStaker>(from: FlowIDTableStaking.NodeStakerStoragePath)
+            ?? panic("Could not borrow node reference from storage path")
+
+        let nodeInfo = FlowIDTableStaking.NodeInfo(nodeID: nodeRef.id)
+
+        // If the node is a collector or consensus node, create a secondary account for their specific objects
+        if nodeInfo.role == 1 as UInt8 {
+
+            let newAcct = AuthAccount(payer: acct)
```
jordanschalm, line 56:
```suggestion
            let machineAcct = AuthAccount(payer: acct)
```

To match terminology

jordanschalm (approved):
🎸  LGTM

lib/go/test/flow_dkg_test.go
joshuahannan: Since the registration succeeds now, I have to use a different ID than before because the registration later that uses it would fail if I didn't

transactions/epoch/node/register_node.cdc
joshuahannan: Good suggestion! I'll change

PR #143, opened 2021-04-07
Adds admin function approve nodes separate from the end of staking auction

joshuahannan:


contracts/FlowIDTableStaking.cdc
```diff
@@ -645,17 +645,25 @@ pub contract FlowIDTableStaking {
             FlowIDTableStaking.stakingEnabled = true
         }
 
+        /// Ends the staking Auction by removing any unapproved nodes
+        /// and setting stakingEnabled to false
+        pub fun endStakingAuction(approvedNodeIDs: {String: Bool}) {
```
jordanschalm, line 650: Will this be called by the `EpochLifecycle` contract?

jordanschalm (approved):

contracts/FlowIDTableStaking.cdc
joshuahannan: yeah we can call it from the lifecycle contract

PR #145, opened 2021-04-08
Removes new staking contract fields and use storage instead

joshuahannan:
In order to make the upgrade easier, I've removed all the new fields from the staking contract and used account storage instead. This will allow us to upgrade the contract via our normal process instead of having to do a manual state migration to add new fields.

I also added a method to the admin resource that will need to be executed in the same transaction when we perform the upgrade to initialize the claimed fields in storage.

I had to remove comments to get it to deploy properly, but I'll add them back when we are able to have a larger contract.


contracts/FlowIDTableStaking.cdc
```diff
@@ -622,64 +593,68 @@ pub contract FlowIDTableStaking {
...
 
         pub fun startStakingAuction() {
             pre {
-                !FlowIDTableStaking.stakingEnabled: "Cannot start staking auction if it is already in progress"
+                !FlowIDTableStaking.stakingEnabled(): "Cannot start staking auction if it is already in progress"
             }
-            FlowIDTableStaking.stakingEnabled = true
+            let stakingEnabled = FlowIDTableStaking.account.load<Bool>(from: /storage/stakingEnabled)
+
+            FlowIDTableStaking.account.save(true, to: /storage/stakingEnabled)
```
turbolent, line 616: Maybe refactor this and the similar code below in `endStakingAuction` into a function

contracts/FlowIDTableStaking.cdc
```diff
@@ -927,15 +911,35 @@ pub contract FlowIDTableStaking {
...
+        self.account.save(claimedDictionary, to: path)
+    }
+
+    pub fun stakingEnabled(): Bool {
+        if let stakingEnabled = self.account.load<Bool>(from: /storage/stakingEnabled) {
+            self.account.save(stakingEnabled, to: /storage/stakingEnabled)
+            return stakingEnabled
+        } else {
+            return false
+        }
```
turbolent, line 941: Don't need to `load` (remove) and re-`save` just to read: Bool is a struct, you can use `copy`:

```suggestion
        return self.account.copy<Bool>(from: /storage/stakingEnabled) ?? false
```

contracts/FlowIDTableStaking.cdc
```diff
@@ -878,13 +841,35 @@ pub contract FlowIDTableStaking {
...
+
+            for nodeID in FlowIDTableStaking.nodes.keys {
+                claimedNetAddressDictionary[FlowIDTableStaking.nodes[nodeID]?.networkingAddress!]
+                claimedNetKeyDictionary[FlowIDTableStaking.nodes[nodeID]?.networkingKey!]
+                claimedStakingKeysDictionary[FlowIDTableStaking.nodes[nodeID]?.stakingKey!]
+            }
+
+            FlowIDTableStaking.account.save(claimedNetAddressDictionary, to: /storage/networkingAddressesClaimed)
+            FlowIDTableStaking.account.save(claimedNetKeyDictionary, to: /storage/networkingKeysClaimed)
+            FlowIDTableStaking.account.save(claimedStakingKeysDictionary, to: /storage/stakingKeysClaimed)
```
turbolent, line 864: Maybe use `updateClaimed` here?

contracts/FlowIDTableStaking.cdc
```diff
@@ -993,7 +990,31 @@ pub contract FlowIDTableStaking {
...
+            ?? panic("Invalid path for dictionary")
+
+        var claimed = false
+        if claimedDictionary[key] != nil {
+            claimed = true
+        } else {
+            claimed = false
+        }
+        self.account.save(claimedDictionary, to: path)
+        return claimed
```
turbolent, line 1016: No need to `load` and re-`save` after the read, you can either just `copy` or `borrow`, and then shorten the if-statement to nil-coaelescing to provide a default if there is no entry:

```suggestion
		let claimedDictionary = self.account.borrow<&{String: Bool}>(from: path)
            ?? panic("Invalid path for dictionary")
        return claimedDictionary[key] ?? false
```

turbolent (commented):
Yep, looks good, this is exactly what I had meant with replacing the fields with storage! 

It's unfortunate that the code becomes more complicated, and I wish we would have implemented contract fields to just do this under the hood, but at least this is more efficient and doesn't require a state migration

contracts/FlowIDTableStaking.cdc
```diff
@@ -622,64 +593,68 @@ pub contract FlowIDTableStaking {
...
+        /// and setting stakingEnabled to false
         pub fun endStakingAuction(approvedNodeIDs: {String: Bool}) {
             pre {
-                FlowIDTableStaking.stakingEnabled: "Cannot end the staking auction if it is not currently in progress"
+                FlowIDTableStaking.stakingEnabled(): "Cannot end the staking auction if it is not currently in progress"
             }
 
+            self.removeUnapprovedNodes(approvedNodeIDs: approvedNodeIDs)
+
+            let stakingEnabled = FlowIDTableStaking.account.load<Bool>(from: /storage/stakingEnabled)
```
jordanschalm, line 628: seems like we don't use this value after loading it

contracts/FlowIDTableStaking.cdc
```diff
@@ -878,13 +841,35 @@ pub contract FlowIDTableStaking {
...
+                ?? panic("Invalid path for networking address dictionary")
+
+            let claimedNetKeyDictionary = FlowIDTableStaking.account.load<{String: Bool}>(from: /storage/networkingKeysClaimed)
+                ?? panic("Invalid path for networking key dictionary")
+
+            let claimedStakingKeysDictionary = FlowIDTableStaking.account.load<{String: Bool}>(from: /storage/stakingKeysClaimed)
+                ?? panic("Invalid path for staking key dictionary")
+
+            for nodeID in FlowIDTableStaking.nodes.keys {
+                claimedNetAddressDictionary[FlowIDTableStaking.nodes[nodeID]?.networkingAddress!]
```
jordanschalm, line 857:
```suggestion
                claimedNetAddressDictionary[FlowIDTableStaking.nodes[nodeID]?.networkingAddress!] = true
```

Looks like we're not writing anything to the dictionary but might be my Cadence knowledge limitations

jordanschalm (approved):

jordanschalm:
Just to confirm: does this mean we don't need the manual state upgrade any more? https://github.com/dapperlabs/flow-go/issues/5424

contracts/FlowIDTableStaking.cdc
joshuahannan: oh yeah I forgot this existed! Thanks for reminding me! :)

contracts/FlowIDTableStaking.cdc
joshuahannan: I thought about using that, but this seemed like a more efficient way to do it, though I guess we'll only run it once so it might not be an issue

contracts/FlowIDTableStaking.cdc
joshuahannan: Good suggestion!

contracts/FlowIDTableStaking.cdc
joshuahannan: yeah, I removed the variable since we don't use it

contracts/FlowIDTableStaking.cdc
joshuahannan: yeah you're right. I was waiting to add a test until I got the first review, but I'll add one for it now

joshuahannan:
@jordanschalm Yes, this should mean that we can do it without the manual upgrade, but I'll need to double check one more time once I merge this that everything should be ok

PR #148, opened 2021-04-12
changes fields to functions

joshuahannan:
We can't add fields to some of the staking info structs, so I had to change it to use functions instead so the upgrade is easier


jordanschalm (approved):
LGTM

PR #149, opened 2021-04-12
General upgrades to staking contract

joshuahannan:
This PR introduces some changes to the staking contract to better organize the code, avoid excessive gas usage, provide new utilities to developers and users, and prepare for the staking collection and epoch contracts.
We should be able to do this as a normal upgrade without a complicated manual state migration, by upgrading the contract, then running the `FlowIDTableStaking.Admin.setClaimed()` function, then running the `FlowIDTableStaking.Admin.startStakingAuction()` function.

* Adds new utility functions that track if a particular node metadata is in use. Before, we would just iterate through the whole identity table to make sure a certain id, networking address, networking key, or staking key wasn't already in use, which could take more and more gas as more nodes are added to ID table, but now we use "fields" to track that. These are dictionaries that we store in the storage of the contract's account. When we need to read it, we use a utility function to copy the dictionary and read the needed value to check if a piece of metadata is in use. When we need to update it, we load it from storage, update the required field, then store it back it storage. When we perform the contract upgrade, we'll run the `setClaimed` function to initialize these fields with the existing node information.
* Adds another "field" that tracks if staking is enabled or not. Staking operations are only allowed to be performed during the staking auction phase, which currently is the entire epoch, but in the future, the staking auction will be shorter, so the admin or central epoch contract needs the ability to control whether or not staking is enabled.
* Separates approving nodes from ending the staking auction. This way, if the protocol needs to enforce that certain nodes should not be staked, it can submit transactions to remove nodes at any time without having to end the staking auction.
* Moves node and delegator info functions to their respective `NodeInfo` and `DelegatorInfo` objects so they can be grouped in a shared struct with all the other node and delegator info instead of being an awkward single function for each
* Removes the `totalTokensStaked` field from `NodeInfo` because the naming wasn't clear. It is replaced by `totalStakedWithDelegators`
* Removes some `balance > 0` checks that were only there before to prevent a panic when a zero balance vault was deposited. This restriction was removed from the `FungibleToken` interface so some of these checks are not needed any more.

Some of these changes, primarily the changes to the node info and delegator info getters, will be breaking for certain apps that are querying the staking contract via the old getters. I think this is a better way to organize it, so I think it is worth changing for now, but I am open to other ideas if we want to avoid breaking changes


rheaplex (approved):
I like the strategy of using storage.

Can you use borrow rather than load/save, at least for composite values?

joshuahannan:
We don't do load any composite values here, unless I am missing something, so I don't think that would work

contracts/FlowIDTableStaking.cdc
```diff
@@ -1041,8 +1019,17 @@ pub contract FlowIDTableStaking {
...
+        self.account.save(true, to: /storage/stakingEnabled)
+
         self.nodes <- {}
 
+        let stakingKeyClaimed: {String: Bool} = {}
+        self.account.save(stakingKeyClaimed, to: /storage/stakingKeysClaimed)
+        let networkingKeyClaimed: {String: Bool} = {}
+        self.account.save(networkingKeyClaimed, to: /storage/networkingKeysClaimed)
+        let networkingAddressClaimed: {String: Bool} = {}
+        self.account.save(networkingAddressClaimed, to: /storage/networkingAddressesClaimed)
```
turbolent, line 1031: Maybe simplify to

```suggestion
        self.account.save({}, to: /storage/stakingKeysClaimed)
        self.account.save({}, to: /storage/networkingKeysClaimed)
        self.account.save({}, to: /storage/networkingAddressesClaimed)
```

contracts/LockedTokens.cdc
```diff
@@ -591,6 +588,51 @@ pub contract LockedTokens {
...
+        access(account) fun getVaultBalance(): UFix64 {
+            let vaultRef = self.lockedVault!.borrow()!
+
+            return vaultRef.balance
+        }
+
+        // A contract in the same account can call these functions when given a LockedVaultHolder,
+        // but nobody else can
+        
+        access(account) fun addVault(lockedVault: Capability<&FlowToken.Vault>) {
```
turbolent, line 611: Maybe this should be called `setVault`? "Add" sounds non-destructive, whereas this function replaces the potentially existing locked vault capability

transactions/idTableStaking/admin/remove_unapproved_nodes.cdc
```diff
@@ -0,0 +1,25 @@
...
+        // borrow a reference to the admin object
+        self.adminRef = acct.borrow<&FlowIDTableStaking.Admin>(from: FlowIDTableStaking.StakingAdminStoragePath)
+            ?? panic("Could not borrow reference to staking admin")
+    }
+
+    execute {
+        let approvedIDs: {String: Bool} = {}
+        for id in ids {
+            approvedIDs[id] = true
+        }
```
turbolent, line 21: Maybe just make the transaction parameter `{String: Bool}` instead of constructing it in the transaction

transactions/lockedTokens/user/get_total_balance.cdc
```diff
@@ -54,12 +49,8 @@ pub fun main(address: Address): UFix64 {
             nodeID: delegatorRef.nodeID,
             delegatorID: delegatorRef.id
         )
-        sum = sum 
-              + delegatorInfo.tokensStaked
-              + delegatorInfo.tokensCommitted
-              + delegatorInfo.tokensUnstaking
-              + delegatorInfo.tokensUnstaked
-              + delegatorInfo.tokensRewarded
+        sum = sum + delegatorInfo.totalTokensInRecord()
```
turbolent, line 52: 👍

contracts/FlowIDTableStaking.cdc
joshuahannan: Should that be possible? When I try to do that, I get this error:
```
expected `Storable`, got `{Never: Never}`
```
I think I still need to do it the way I have it because I need to specify the type. Is there a better way to do it?

PR #154, opened 2021-04-15
Add Method to storage fees to get usable balance

janezpodhostnik:
ref: https://github.com/dapperlabs/flow-internal/issues/1499

Added a convenience method to get usable balance of an account.


joshuahannan (approved):
Looks good! Thanks for adding. 👍

PR #159, opened 2021-04-16
Available balance fix and comments

janezpodhostnik:
ref: https://github.com/dapperlabs/flow-go/issues/5204

There was a bug in https://github.com/onflow/flow-core-contracts/pull/154 when storage used was less than the minimum account balance, minimum account balance should have been deducted from the balance instead the used storage was deducted.

I also took this opportunity to add more comments.


joshuahannan (approved):
Good catch! And thanks for adding more comments. 👍

PR #162, opened 2021-04-19
Remove panic from get_latest_whiteboard_messages.cdc

arrivets:
We shouldn't panic when someone requests broadcast messages with an out-of-range
start index; just return an empty array

ex: 
the contract contains 1 broadcast message
client requests messages with fromIndex=1
the contract should not panic, just return an empty array of messages


joshuahannan (approved):
looks good! Thanks for making the change. 👍

PR #164, opened 2021-04-20
Adds support for nil submissions to the DKG and ends DKG and QC at the end of epoch

### LINKED ISSUE #152: DKG: Handle nil key vector submissions
## Context

When the DKG fails, it returns a key vector containing nil keys. Despite the failure, these key vectors should still be submitted to the smart contract. When deciding whether the DKG has completed, we should consider a key vector (final submission) containing any nil keys to be invalid, even if a majority of participants have submitted that key vector. 

## Definition of Done

* Decide how to serialize a nil key (either change smart contract to accept `Array<Optional<String>>` and encode nil keys as nil or define a canonical serialization for a nil key (eg. 0x00...) coordinate with https://github.com/dapperlabs/flow-go/issues/5452
* Update key vector submission / storage to use updated encoding, if necessary
* Modify logic determining when DKG is done to consider case where majority key vector contains nil keys as the DKG not being done

cc @tarakby @danuio

### LINKED ISSUE #155: Allow DKG key vector submissions after DKG is finished
## Context

The current behaviour of the `FlowDKG` and `EpochLifecycle` contracts is to stop the DKG for a given epoch once the threshold (50%) of key vector submissions is met. We can still consider the `FlowDKG` complete once this threshold is reached, and emit the `EpochCommit` event at this point, but we would like to continue allowing key vector submissions after the threshold has been met. 

This allows us to have a better sense for the health of the DKG by seeing how many nodes submitted valid key vectors (and therefore can be reasonably expected to participate in consensus in the next epoch).

## Definition of Done

* `EpochLifecycle` - end `DKG` immediately before transitioning to next epoch rather than when the `DKG` reports being "done"

joshuahannan:
Closes #152 
Closes #155 

# Description
Changes the type of DKG final submissions to `[String?]` so that nodes can submit `nil` keys.
`nil` Keys are allowed as a valid submission, but if the majority submission has any `nil` keys in it, the dkg is not considered to be completed

Also moves the disabling of the qc and dkg contracts to the end of the epoch so all nodes still have time to get their submissions in before the epoch ends.


contracts/epochs/FlowDKG.cdc
```diff
@@ -216,8 +219,20 @@ pub contract FlowDKG {
 
             // if a key is different, stop checking this submission
             // and return false
-            if key != existingSubmission[index] {
-                return false
+            if let keyUnwrap = key {
```
jordanschalm, line 222: I would expect the way you had it previously would still work. Cadence seems to handle equality of optional strings the way you would expect based on a simple [example in the Playground](https://play.onflow.org/aadabdd1-4a25-46ce-86ca-865c2178361d?type=script&id=9d8b8c96-7e4c-450b-95a5-cb7e6ec0c694).

lib/go/test/flow_dkg_test.go
```diff
@@ -547,3 +547,116 @@ func TestDKG(t *testing.T) {
...
+
+	DKGAddress, err := b.CreateAccount([]*flow.AccountKey{DKGAccountKey}, []sdktemplates.Contract{
+		{
+			Name:   "FlowDKG",
+			Source: string(DKGCode),
+		},
+	})
+	if !assert.NoError(t, err) {
+		t.Log(err.Error())
+	}
```
jordanschalm, line 573:
```suggestion
	assert.NoError(t, err)
```
`NoError` will log the error itself if non-nil

jordanschalm (approved):
Looks good 🎸

contracts/epochs/FlowDKG.cdc
joshuahannan: good point! I'll change it back

PR #170, opened 2021-04-25
replaced adminAccount from init with self.account

### LINKED ISSUE #169: Error: GraphQL error: failed to deploy account code: Execution failed: error: invalid argument count, too few arguments:
### Problem

```
Error
Error: GraphQL error: failed to deploy account code: Execution failed: error: invalid argument count, too few arguments: expected 1, got 0, next missing argument: `AuthAccount` --> 03b8fe83a39fbe146dc47a1c8b69373cba34a97b29a9453c87bd498cfc154a63:4:2 | 4 | signer.contracts.add(name: name, code: code.decodeHex()) | 
```

### Steps to Reproduce 

In running the fungible token example in cadence playground v0.12.5, after deploying the ```FungibleToken``` contract interface to account 0x01, the error appears when attempting to deploy the ```FlowToken``` contract from this repo.

### Acceptance Criteria

The ```FlowToken``` example code should run out of the box. I believe it used to run without this error; perhaps something has changed?
  
### Context

Just working through the starter playground project:
https://play.onflow.org/a10239fd-6cbe-49ca-a7c1-87324815a9ff?type=account&id=0

Is this error caused by the playground interpreter? or do I need to add ```AuthAccount``` as a parameter somewhere? 

I assume this is on the backend as no contract code is referenced in the error.

Jovonni:
This Fixes #169, 

[Key Assumption]
Admin account is always the actual owning address. If self.account isn't the intended admin account, then this would be too restrictive. For clarity, how else would we pass in the AuthAccount to the init function if the contract won't even deploy?

Thanks in advance

### DIFF
--- a/contracts/FlowToken.cdc
+++ b/contracts/FlowToken.cdc
@@ -165,18 +165,18 @@ pub contract FlowToken: FungibleToken {
         }
     }
 
-    init(adminAccount: AuthAccount) {
+    init() {
         self.totalSupply = 0.0
 
         // Create the Vault with the total supply of tokens and save it in storage
         //
         let vault <- create Vault(balance: self.totalSupply)
-        adminAccount.save(<-vault, to: /storage/flowTokenVault)
+        self.account.save(<-vault, to: /storage/flowTokenVault)
 
         // Create a public capability to the stored Vault that only exposes
         // the `deposit` method through the `Receiver` interface
         //
-        adminAccount.link<&FlowToken.Vault{FungibleToken.Receiver}>(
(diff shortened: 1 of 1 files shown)


Jovonni:
Also let me know if this PR and issue are in the wrong repo as well as this is specific to the playground

joshuahannan:
I appreciate the PR, but we need to leave the contract like this. This contract is imported by the flow-go and emulator repos for bootstrapping the emulator and localnet, and we have to provide the service account as the admin, so this should stay the way it is. Sorry for the inconvenience!

PR #171, opened 2021-04-28
Decouples rewards payment from the end of the epoch

joshuahannan:
Updated `FlowIDTableStaking` to:
* Calculate rewards separately from paying them
* Removed some comments to fit in transaction size limit

Updated `FlowEpoch` to:
* Add a field to epoch metadata to store the calculated rewards from the epoch
* Call the calculate rewards function when ending an epoch and store the list
* Admin can call pay rewards at any time during the next epoch


contracts/epochs/FlowEpoch.cdc
```diff
@@ -385,16 +397,17 @@ pub contract FlowEpoch {
...
-            FlowEpoch.epochMetadata[FlowEpoch.currentEpochCounter] = firstEpochMetadata
+            FlowEpoch.epochMetadata[FlowEpoch.currentEpochCounter] = newEpochMetadata
         }
     }
 
     /// Pays rewards to staked nodes and delegators,
     /// Calculates a new token payout for the next epoch
     /// and sets the new payout
-    access(account) fun payAndSetRewards(_ newPayout: UFix64?) {
+    access(account) fun calculateAndSetRewards(_ newPayout: UFix64?) {
```
jordanschalm, line 407: Does this function need to be called manually each epoch? 

From the PR description it sounded to me like we wanted to do this automatically when ending the epoch, but I don't see where that happens.

> Call the calculate rewards function when ending an epoch and store the list

contracts/epochs/FlowEpoch.cdc
```diff
@@ -406,10 +419,14 @@ pub contract FlowEpoch {
         }
     }
 
+    access(account) fun payRewards() {
```
jordanschalm, line 422: Should we add some protection against paying out multiple times, since this will be exposed as callable from a transaction?

jordanschalm (commented):
🎸  Nice work - added a few questions.

contracts/epochs/FlowEpoch.cdc
joshuahannan: yeah, we could potentially just set the rewards breakdown field in epoch metadata to `nil` and have pay rewards not work if the field is `nil`. Do you think that is ok? Or should we just have a `rewardsPaid: Bool` field?

contracts/epochs/FlowEpoch.cdc
joshuahannan: I forgot to include the automatic call to it in the heartbeat method, but I've included it in my latest changes

contracts/epochs/FlowEpoch.cdc
jordanschalm: Both work but the `rewardsPaid` field seems cleaner to me

jordanschalm (approved):

PR #175, opened 2021-05-11
Adds functions to track nodes' secondary machine accounts

joshuahannan:
The FlowEpoch contract now tracks where collector and consensus nodes' qc and dkg objects are stored in the contract.
This dictionary is stored in the contract storage in order to make loading the contract object more efficient.

I also picked up the latest changes from master in this PR


contracts/epochs/FlowEpoch.cdc
```diff
@@ -647,6 +648,22 @@ pub contract FlowEpoch {
         return array
     }
 
+    /// Gets the the address of the secondary account that is associated
```
jordanschalm, line 651:
```suggestion
    /// Gets the address of the secondary account that is associated
```

contracts/epochs/FlowEpoch.cdc
```diff
@@ -647,6 +648,22 @@ pub contract FlowEpoch {
...
+    /// with collector and consensus nodes
+    pub fun updateMachineAccountForNode(qcVoterReference: &FlowEpochClusterQC.Voter?, dkgParticipantReference: &FlowDKG.Participant?) {
+        let accountDictionary = self.account.load<{String: Address}>(from: self.machineAccountsStoragePath)
+            ?? panic("Invalid path for machine account dictionary")
+
+        if let voterRef = qcVoterReference {
+            accountDictionary[voterRef.nodeID] = voterRef.owner!.address
```
jordanschalm, line 658: Is `owner` a builtin field for reference types?

transactions/idTableStaking/admin/capability_end_epoch.cdc
```diff
@@ -0,0 +1,41 @@
...
+// and moves tokens between buckets
+// It also sets a new token payout for the next epoch
+
+transaction(ids: [String], newPayout: UFix64) {
+
+    // Local variable for a reference to the ID Table Admin object
+    let adminRef: &FlowIDTableStaking.Admin
+
+    prepare(acct: AuthAccount) {
+        let adminCapability = acct.copy<Capability>(from: FlowIDTableStaking.StakingAdminStoragePath)
```
jordanschalm, line 17: Is a capability not a resource type?

contracts/epochs/FlowEpoch.cdc
```diff
@@ -647,6 +648,22 @@ pub contract FlowEpoch {
         return array
     }
 
+    /// Gets the the address of the secondary account that is associated
+    /// with collector and consensus nodes
+    pub fun updateMachineAccountForNode(qcVoterReference: &FlowEpochClusterQC.Voter?, dkgParticipantReference: &FlowDKG.Participant?) {
```
jordanschalm, line 653: This approach feels like we are mixing concerns a bit to me. I wonder if it would make more sense to do the machine account house-keeping in the user's own staking account (does this fit into how the staking collection works?).

The notion of a machine account is outside the scope of what the epoch smart contract needs to be aware of in my opinion. The epoch smart contract exposes the DKG and QC participant resources, but I don't think it should make assumptions about how those resources are stored or used. Flow Port put the resources into machine accounts with a 1-1 relation between node IDs and resources as the default case, but ultimately they are owned by node operators who can deal with them however they like.

joshuahannan: That is a good point. Though I thought we said that we wanted the machine account thing to be something that the protocol expects all node operators to do?

joshuahannan: Though I guess we can't enforce that since we are setting it in a transaction instead of a function of the contract, so probably shouldn't have it here

jordanschalm: > we said that we wanted the machine account thing to be something that the protocol expects all node operators to do

All node operators will eventually require a machine account that contains resources, but how those accounts and resources are set up can't really (and shouldn't) be enforced in this contract like you mentioned.

joshuahannan:
closing this to make the changes to the staking collection instead

PR #184, opened 2021-05-21
Adds machine account functionality to staking collection

joshuahannan:
Closes https://github.com/dapperlabs/flow-internal/issues/1510

This PR includes the ability to track node's secondary machine accounts in their staking collection. 
If you are not familiar with the staking collection, I recommend [reading this summary to get up to speed.](https://github.com/onflow/flow-core-contracts/pull/119)

The machine account record lives in the staking collection for all nodes who use staking collections. It is not mandatory for nodes to use this construct, but will be if the node operator wants to view their node information properly via Flow Port. This is why the record is in the staking collection contract and not the epoch contracts.
The record is a dictionary mapping Node IDs (String) to `MachineAccountInfo` structs. When a user registers a node, the staking collection creates a new account for them, adds the qc or dkg object to the new account, gets the vault capability from the new account, and returns the new account to the caller. The caller must then add public keys to the new account in the transaction.

For existing nodes, the user will simply call the `createMachineAccountForExistingNode` and it will do the same thing for them.

Let me know if you have any questions!


joshuahannan:
@jordanschalm @JeffreyDoyle I've added you as reviewers to this PR. Let me know if you have any questions. I haven't added tests yet because I want to see your feedback first, but I'm going to start writing some basic ones now

lib/go/test/lockedtokens_helpers.go
```diff
@@ -454,4 +481,35 @@ func verifyStakingCollectionInfo(
...
+// Queries the machine account address of a recently registered Node
+func getMachineAccountFromEvent(
+	t *testing.T,
+	b *emulator.Blockchain,
+	env templates.Environment,
+	result *types.TransactionResult,
+) flow.Address {
+
+	for _, event := range result.Events {
+		fmt.Println(event.Type)
```
jordanschalm, line 506: debug log - might want to remove

transactions/stakingCollection/register_node.cdc
```diff
@@ -8,22 +8,27 @@ transaction(id: String,
             networkingAddress: String,
             networkingKey: String,
             stakingKey: String,
-            amount: UFix64) {
+            amount: UFix64,
+            publicKeys: [[UInt8]]) {
```
jordanschalm, line 12:
```suggestion
            machineAccountPublicKeys: [[UInt8]]) {
```

contracts/FlowStakingCollection.cdc
```diff
@@ -749,8 +855,46 @@ pub contract FlowStakingCollection {
             return delegatorInfo
         }
 
+        /// Function to get the list of all of the machine accounts for a node in the collection
+        /// Key: Node ID
+        /// Value: Account address for the account that stores the QC or dkg object
+        pub fun getMachineAccounts(): {String: Address} {
+            let accountsRecord = FlowStakingCollection.account.copy<{String: Address}>(from: /storage/epochMachineAccounts)
```
jordanschalm, line 862: Should this path be a contract constant?

jordanschalm, line 862: If I'm understanding correctly, the machine accounts are stored in the storage of the account where the `FlowStakingCollection` contract is deployed (presumably the service account) rather than in the user's own account. I'm curious why it's stored there rather than in the user's own account. Seems like a potential abuse vector if we provide a mechanism to store data where the storer doesn't have to pay for it 🤔 .

contracts/epochs/FlowEpoch.cdc
```diff
@@ -716,20 +725,7 @@ pub contract FlowEpoch {
...
-
-        let DKGAdmin <- self.account.load<@FlowDKG.Admin>(from: FlowDKG.AdminStoragePath)
-            ?? panic("Could not load DKG Admin from storage")
-        self.DKGAdmin <- DKGAdmin
-
-        let stakingAdmin <- self.account.load<@FlowIDTableStaking.Admin>(from: FlowIDTableStaking.StakingAdminStoragePath)
-            ?? panic("Could not load staking Admin from storage")
-        self.stakingAdmin <- stakingAdmin
-        self.stakingAdmin.startStakingAuction()
+        self.borrowStakingAdmin().startStakingAuction()
```
jordanschalm, line 728: 👍  Nice

transactions/stakingCollection/register_node.cdc
jordanschalm: Maybe this should be optional, since we don't create a machine account for all node roles?

transactions/stakingCollection/register_node.cdc
```diff
@@ -8,22 +8,27 @@ transaction(id: String,
...
             networkingAddress: networkingAddress,
             networkingKey: networkingKey,
             stakingKey: stakingKey,
-            amount: amount)
+            amount: amount,
+            payer: account) 
+        {
+            for key in publicKeys {
+                machineAccount.addPublicKey(key)
+            }
```
jordanschalm, line 31: May also want to add a sanity check that there is at least one key here - otherwise the machine account will be unusable.

lib/go/test/lockedtokens_helpers.go
joshuahannan: oh yes, good catch. Thanks!

transactions/stakingCollection/register_node.cdc
joshuahannan: I was just going to have the `nil` case be an empty array but it is probably better and more explicit to just make it an optional. 👍

contracts/FlowStakingCollection.cdc
joshuahannan: It should be, but this is going to be an upgrade to the existing staking collection contract, so we can't add fields in an upgrade. Should I include the constant in the first deployment anyway, even though we won't use it until the upgrade?

joshuahannan: Like with my other comment, this is an upgrade to the original staking collection in this PR: https://github.com/onflow/flow-core-contracts/pull/119

So we can't create that field in this version because adding fields in upgrades isn't allowed. Should I include it in the first version anyway even though we won't use it?

transactions/stakingCollection/register_node.cdc
joshuahannan: Good idea. I'll add that

joshuahannan:
@jordanschalm based on your comments, now I am thinking that it probably would be a good idea include those machine account fields in the first staking collection PR. I'll port some of these changes to that PR.

jordanschalm (approved):
Looks good to me. I reviewed the [diff since my previous review](https://github.com/onflow/flow-core-contracts/pull/184/files/c8b047b3d344d666c7796e3caf3663c8b435bb29..6c7ac0033add3f90d80d6ba4efcdcfa1b51e679c). 

One thing that occurred to me while reviewing this is how we will need to handle (specifically) a lost machine account key and (generally) any lost operator key. With epochs deployed, the keys people specify to the staking contract are used directly in the protocol state and are not changeable without un-staking and re-staking. Asked about this [in Slack](https://axiomzen.slack.com/archives/C015G65HR2P/p1622491252128900).

transactions/stakingCollection/register_node.cdc
```diff
@@ -26,7 +26,10 @@ transaction(id: String,
             amount: amount,
             payer: account) 
         {
-            for key in publicKeys {
+            if publicKeys == nil || publicKeys!.length == 0 {
+                panic("Cannot provide zero keys for the machine account")
+            }
+            for key in publicKeys! {
                 machineAccount.addPublicKey(key)
```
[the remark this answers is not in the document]
joshuahannan: The weight of the key is determined when creating it with before passing it to the transaction, so this doesn't assume the keys are full weight, it just adds whatever was passed as an argument

joshuahannan:
I am inclined to say that if someone loses that machine account key, it is their problem and they need to create a new node. Do you think this is something we need to address in this PR, or can we merge this and deal with it later?

jordanschalm:
> Do you think this is something we need to address in this PR

Definitely not. I agree it is the operator's responsibility. I was mainly wanting to see whether by deploying the epochs changes we are changing the implications of key loss from what it is currently. From the Slack discussion it sounds like we are not, and that operators already are required to unstake and restake, so I don't think we need to make any other changes.

PR #187, opened 2021-05-24
Move isRestrictedAccountCreationEnabled inside FlowServiceAccount

janezpodhostnik:
Creating an account  in the FVM checks a setting *isRestrictedAccountCreationEnabled* before checking `FlowServiceAccount.isAccountCreator`. This setting needs to be moved to from the FVM to the `FlowServiceAccount`, so that checking this setting can go within the `setupNewAccount` function. This will allow the call to `setupNewAccount` to be a direct function call instead of a meta-transaction (which would be faster). 

This relates to: https://github.com/dapperlabs/flow-internal/issues/1513


contracts/FlowServiceAccount.cdc
```diff
@@ -92,9 +97,19 @@ pub contract FlowServiceAccount {
 
     /// Returns true if the given address is permitted to create accounts, false otherwise
     pub fun isAccountCreator(_ address: Address): Bool {
+        if !self.isRestrictedAccountCreationEnabled() {
+            return true
+        }
         return self.accountCreators[address] ?? false
     }
```
joshuahannan, line 105: Can you put a comment above this describing what it does?

contracts/FlowServiceAccount.cdc
```diff
@@ -92,9 +97,19 @@ pub contract FlowServiceAccount {
 
     /// Returns true if the given address is permitted to create accounts, false otherwise
     pub fun isAccountCreator(_ address: Address): Bool {
+        if !self.isRestrictedAccountCreationEnabled() {
```
joshuahannan, line 100: Would it be possible to rename this to something like `accountCreationRestricted` or something like that? The current name is kind of unclear what it is describing

joshuahannan (commented):
Left a few comments. Are you able to write a test in this repo for this also? That would be good to have. 👍

janezpodhostnik:
Added tests. Though this make the PR a lot larger :)

joshuahannan (approved):
Looks good! Thanks for adding tests. 👍

contracts/FlowServiceAccount.cdc
```diff
@@ -92,9 +97,21 @@ pub contract FlowServiceAccount {
...
+            return true
+        }
         return self.accountCreators[address] ?? false
     }
 
+    /// Is true if new acconts can only be created by approved accounts `self.accountCreators`
+    pub fun isAccountCreationRestricted(): Bool {
+        let value = self.account.load<Bool>(from: /storage/isAccountCreationRestricted) ?? false
+        self.account.save<Bool>(value ,to: /storage/isAccountCreationRestricted)
+        return value
```
turbolent, line 111: No need to remove the value from storage and write it back:

```suggestion
        return self.account.copy<Bool>(from: /storage/isAccountCreationRestricted) ?? false
```

contracts/FlowServiceAccount.cdc
```diff
@@ -126,6 +143,11 @@ pub contract FlowServiceAccount {
             FlowServiceAccount.accountCreators.remove(key: accountCreator)
             emit AccountCreatorRemoved(accountCreator: accountCreator)
         }
+
+         pub fun setIsAccountCreationRestricted(_ enabled: Bool) {
+            FlowServiceAccount.account.load<Bool>(from: /storage/isAccountCreationRestricted)
+            FlowServiceAccount.account.save<Bool>(enabled ,to: /storage/isAccountCreationRestricted)
```
turbolent, line 149:
```suggestion
			let path = /storage/isAccountCreationRestricted
            FlowServiceAccount.account.load<Bool>(from: path)
            FlowServiceAccount.account.save<Bool>(enabled, to: path)
```

turbolent (approved):

contracts/FlowServiceAccount.cdc
```diff
@@ -126,6 +143,11 @@ pub contract FlowServiceAccount {
             FlowServiceAccount.accountCreators.remove(key: accountCreator)
             emit AccountCreatorRemoved(accountCreator: accountCreator)
         }
+
+         pub fun setIsAccountCreationRestricted(_ enabled: Bool) {
```
turbolent, line 147: Should this maybe emit an event?

janezpodhostnik: I did not know adding an event type was legal. I will add this.

contracts/FlowServiceAccount.cdc
janezpodhostnik: Ah! Did not know about the `copy` method. Must have missed it in the docs. Thank you for this.

joshuahannan: @turbolent Saving the value to storage if it was copied causes the save fail, at least when I have tried, so I have to load it during my testing.  Is this not how it is supposed to behave?

turbolent: Yes, this is the behaviour of save, i.e. it aborts if there is already a stored value.
Here, there is no need to first remove it from storage and then put it back, the function only reads it and the read value is a boolean, so it can just be copied. Note that the suggestion replaces the load + save with just a single copy

joshuahannan: oh yeah you're right. I didn't look close enough. My bad!

contracts/FlowServiceAccount.cdc
janezpodhostnik: Added the emit event.

contracts/FlowServiceAccount.cdc
janezpodhostnik: added event emit.

joshuahannan:
@janezpodhostnik Looks good to me! Can we merge this?

PR #188, opened 2021-05-26
Addresses PR Comments and adds some events

joshuahannan:
I made changes to address all the PR comments in https://github.com/onflow/flow-core-contracts/pull/119.

I also fixed the existing events and added a couple more


contracts/FlowStakingCollection.cdc
```diff
@@ -388,37 +408,37 @@ pub contract FlowStakingCollection {
...
-                if let delegator = self.borrowDelegator(nodeID, _delegatorID) {
+                if let delegator = self.borrowDelegator(nodeID: nodeID, delegatorID: delegatorID) {
                     delegator.delegateNewTokens(from: <- self.getTokens(amount: amount))
                 } else {
                     // Get any needed unlocked tokens, and deposit them to the locked vault.
                     let lockedBalance = self.lockedVault!.borrow()!.balance - FlowStorageFees.minimumStorageReservation
                     if (amount > lockedBalance) {
-                        self.tokenHolder!.borrow()!.deposit(from: <- self.unlockedVault.borrow()!.withdraw(amount: amount - lockedBalance))
+                        let numUnlockedTokensToUse = amount - lockedBalance
+                        self.tokenHolder!.borrow()!.deposit(from: <- self.unlockedVault.borrow()!.withdraw(amount: numUnlockedTokensToUse))
```
turbolent, line 420: Maybe also refactor those borrows out to the top (`self.tokenHolder!.borrow()!` and `self.unlockedVault.borrow()!`), as they're done in both if-branches. It would make the code here more readable

turbolent (commented):
Nice refactor! I only looked at the Cadence code, I don't know the details about staking logic, so someone else who does should maybe check that

PR #189, opened 2021-05-26
Adds struct and fields for machine account records

joshuahannan:
In preparation for epochs, I am including the machine account information in a struct and field in the staking collection. These won't actually get used in the first version of the staking collection, but once epochs are deployed, we will upgrade the staking collection to utilize it.


joshuahannan:
@jordanschalm Do these look like they are ok for tracking the machine account information? I'll update the other PR to use these instead of the stuff in account storage from before

jordanschalm (approved):
Nice, looks good! 👍

PR #191, opened 2021-05-27
Adds check to FlowStorageFees to avoid division by zero

joshuahannan:
When storage fees are disabled, if the `storageCapacityToFlow` function is called, it will fail with a division by zero error. I added a condition in it to return zero if the conversion is zero. I also added a test for it

Also updated some go.mod files and imports to make the project build properly


janezpodhostnik (approved):

janezpodhostnik:
Thank you for adding this!

Issue #195, opened 2021-06-08
EPOCHS: Change Epoch metadata dictionary to storage


joshuahannan:
### Issue To Be Solved
As the epoch metadata field gets larger, loading the epoch contract will get more and more time-consuming. Need to make it more efficient

### Suggest A Solution

* Add a metadata storage path
* Change the epochMetadata field to a function that loads and saves to storage

PR #196, opened 2021-06-08
Epoch Metadata to account storage

joshuahannan:
Closes https://github.com/onflow/flow-core-contracts/issues/195

The epoch metadata dictionary was a field in the contract, but I realized that since it will increase in size as we complete more and more epochs, it would be safer to store it in account storage instead so that it doesn't affect the performance down the line.


contracts/epochs/FlowEpoch.cdc
```diff
@@ -755,6 +791,10 @@ pub contract FlowEpoch {
         self.currentEpochPhase = EpochPhase.STAKINGAUCTION
         self.adminStoragePath = /storage/flowEpochAdmin
         self.heartbeatStoragePath = /storage/flowEpochHeartbeat
+        self.metadataStoragePath = /storage/flowEpochMetadata
+
+        let epochMetadata: {UInt64: EpochMetadata} = {}
+        self.account.save<{UInt64: EpochMetadata}>(epochMetadata, to: self.metadataStoragePath)
```
turbolent, line 798: Type can be inferred from the first argument:

```suggestion
        self.account.save(epochMetadata, to: self.metadataStoragePath)
```

contracts/epochs/FlowEpoch.cdc
```diff
@@ -232,22 +232,55 @@ pub contract FlowEpoch {
...
+            return metadataDictionary[epochCounter]
+        }
+        return nil
+    }
+
+    /// Saves a modified EpochMetadata struct to the metadata in account storage
+    /// 
+    access(contract) fun saveEpochMetadata(_ newMetadata: EpochMetadata) {
+        pre {
+            self.currentEpochCounter == (0 as UInt64) ||
```
turbolent, line 253: Just FYI: With https://github.com/onflow/cadence/pull/957 (to be released, i.e. coming in Cadence v0.17) all those static casts shouldn't be necessary anymore, so this can become more readable

PR #199, opened 2021-06-15
change epochcommitted to epochcommit

joshuahannan:
Change all occurrences of epoch committed to epoch commit to align with what is in flow-go


jordanschalm (approved):
🙏

PR #202, opened 2021-06-23
adds create machine account transaction

joshuahannan:
Adds a transaction that tests the function to create a machine account for a node that already exists in the staking collection


joshuahannan:
Hey @jordanschalm, @JeffreyDoyle told me that it would be easier for Flow Port to accept the public keys for the machine account as a hex-encoded string instead of a byte array. I know we have been using byte arrays in all of our tests and in the node software, but do you have a preference of what format the keys are in?

jordanschalm:
Yes I think we should definitely use a hex-encoded string when passing the public key between the `Bootstrapping Script` and `Flow Port`. This matches what we already do for inputting the other public keys afaik ([example](https://github.com/onflow/flow-core-contracts/blob/feature/epochs/transactions/idTableStaking/node/register_node.cdc#L11-L12))

joshuahannan:
ok, I'll update all these transactions for machine accounts to use a string instead of a byte array

JeffreyDoyle (approved):

PR #205, opened 2021-06-30
Adds QC signature validation and vote counting

joshuahannan:
Uses the crypto API to validate signatures for QC votes
Changes the QC contract to check if quorum has been reached by summing the weight for identical votes instead of all votes

Also changes the name of the QC contract from `FlowEpochClusterQC` to `FlowClusterQC`
Still adding more tests for this


contracts/epochs/FlowEpoch.cdc
```diff
@@ -606,10 +606,11 @@ pub contract FlowEpoch {
 
         // iterate through all the clusters and create their certificate arrays
         for cluster in clusters {
-            var certificate: FlowEpochClusterQC.ClusterQC = FlowEpochClusterQC.ClusterQC(index: cluster.index, votes: [], voterIDs: [])
+            var certificate: FlowEpochClusterQC.ClusterQC = FlowEpochClusterQC.ClusterQC(index: cluster.index, signatures: [], messages: [], voterIDs: [])
 
             for vote in cluster.votes {
```
jordanschalm, line 611: ⚠️  This is including _all_ votes in the output, which will result in an invalid QC if any votes have a different `voteMessage` than the quorum. ⚠️ 

We should only include valid votes in the `ClusterQC` that gets included in the `EpochCommit` event. There are two properties required for a vote to be valid:
1. The `voteSignature` is valid with respect to the `voteMessage` 
2. The `voteMessage` is the same as submitted by a quorum of voters 

We're doing 1 currently, but not fully doing 2. We don't consider a cluster complete until we receive a quorum of votes with consistent `voteMessage`s but we still include votes with inconsistent vote messages in the resulting QC.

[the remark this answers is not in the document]
joshuahannan: Can you elaborate? I don't see what you are referring to.

joshuahannan: ok, I didn't realize we only wanted to include valid votes in the QC also. I'll make that change.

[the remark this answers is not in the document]
jordanschalm: I'm seeing `votes` being removed:
* https://github.com/onflow/flow-core-contracts/pull/205/files#diff-ee29ea76539bcca0d0d00ec4d3d45b9401a3931a8e9162832732e1284b508b88R72-R79
* https://github.com/onflow/flow-core-contracts/blob/qcCounting/contracts/epochs/FlowEpochClusterQC.cdc#L60-L138

Screen Shot 2021-06-30 at 2 27 11 PM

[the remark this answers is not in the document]
joshuahannan: ok I see. yeah I hadn't tested it with the epoch contract yet, but I updated it in my latest commit

joshuahannan:
@jordanschalm I fixed the contract based on your comments if you want to take one more look. I still need to add more tests though so if you'd rather wait to review until you get back from your vacay and all the tests are there, that is fine

tarakby (approved):
👌

contracts/epochs/FlowEpochClusterQC.cdc
```diff
@@ -104,16 +121,20 @@ pub contract FlowEpochClusterQC {
...
-            self.index = index
-            self.nodeWeights = nodeWeights
-
-            var totalWeight: UInt64 = 0
-            for weight in nodeWeights.values {
-                totalWeight = totalWeight + weight
+        /// Returns the status of this cluster's QC process
+        /// If there is a number of identical votes exceeding the `voteThreshold`,
+        /// Then this cluster's QC generation is considered complete and this method returns 
+        /// the vote that reached quorum
```
jordanschalm:
```suggestion
        /// the vote message that reached quorum
```

contracts/epochs/FlowEpochClusterQC.cdc
```diff
@@ -104,16 +121,20 @@ pub contract FlowEpochClusterQC {
...
+        /// If no vote is found to reach quorum, then `nil` is returned
+        pub fun isComplete(): String? {
+            var i = 0
```
jordanschalm:
```suggestion
        pub fun isComplete(): String? {
```

lib/go/test/flow_qc_test.go
```diff
@@ -326,4 +429,193 @@ func TestQuorumCertificate(t *testing.T) {
...
+	numberOfNodesPerCluster := 3
+
+	// Create new user accounts
+	addresses, _, signers := registerAndMintManyAccounts(t, b, accountKeys, numberOfClusters*numberOfNodesPerCluster)
+
+	clusterNodeIDStrings := make([][]string, numberOfClusters*numberOfNodesPerCluster)
+
+	stakingPrivateKeys, stakingPublicKeys, _, _ := generateManyNodeKeys(t, numberOfClusters*numberOfNodesPerCluster)
+
+	clusters := initClusters(clusterNodeIDStrings, numberOfClusters, numberOfNodesPerCluster)
```
jordanschalm, line 501:
```suggestion
	startVotingArguments := initClusters(clusterNodeIDStrings, numberOfClusters, numberOfNodesPerCluster)
```

lib/go/test/flow_qc_test.go
```diff
@@ -326,4 +429,193 @@ func TestQuorumCertificate(t *testing.T) {
...
+				false,
+			)
+
+			result := executeScriptAndCheck(t, b, templates.GenerateGetVotingCompletedScript(env), nil)
+			assert.Equal(t, cadence.NewBool(false), result)
+		}
+
+	})
+
+	t.Run("Should register incomplete if a cluster has different vote messages", func(t *testing.T) {
```
jordanschalm, line 573: I think we can add more context on exactly what is being tested here. A cluster having different vote messages is not sufficient on its own to cause the QC aggregation to not be complete. 
```suggestion
	// If a cluster has received votes with weight exceeding the quorum threshold, but those
	// votes are spread across different vote messages such that no set of votes corresponding
	// to any vote message represents weight exceeding the quorum threshold, the cluster voting
	// should be considered incomplete.
	t.Run("Should register incomplete if a cluster has different vote messages", func(t *testing.T) {
```

jordanschalm, line 573: I would add a test case for the positive version of this test: If a cluster has quorum for a vote message, and some number of votes for a different vote message, that cluster should be considered complete

contracts/epochs/FlowClusterQC.cdc
```diff
@@ -0,0 +1,414 @@
...
+    pub struct ClusterQC {
+
+        /// The index of the qc in the cluster record
+        pub let index: UInt16
+
+        /// The vote signatures from all the nodes in the cluster
+        pub var voteSignatures: [String]
+
+        /// The vote messages from all the nodes in the cluster
+        pub var voteMessages: [String]
```
jordanschalm, line 211: All the messages here must be the same so this could be a `String`

joshuahannan:
@jordanschalm Thanks for the thorough review! I've made the fixes from all your comments and added the additional test cases in 6088ca8. Does it look good to you?

jordanschalm (approved):
Thanks for the updates. LGTM 🎸

PR #216, opened 2021-07-07
Fixing comments from Quantstamp Audit

joshuahannan:
Fixes a few of the comments that Quantstamp made in their audit
* Only emits events when a field actually changes
* Fix a comment


janezpodhostnik (approved):
LGTM

PR #219, opened 2021-07-16
`resetEpoch` updates

jordanschalm:


jordanschalm:
This is not necessary for the same reason as https://github.com/onflow/flow-go/pull/1007#issuecomment-884462482

PR #221, opened 2021-07-22
Deduct entire balance if the balance of an account is less then the transaction fees

janezpodhostnik:
references: https://github.com/dapperlabs/flow-internal/issues/1449

If an account is at (or below) minimum balance and it is a payer for a failing transaction, the transaction fees will still be deducted from the account, even though that puts it (further) below the minimum account balance.

If this continues the account will eventually reach a situation where it has less balance then what it needs to pay for the transaction. If this is the case, deduct the remaining balance from the account.

In the future this condition will not be hit, as transactions where a payer is below the account minimum balance will not be included in a collection in the first place.


joshuahannan (approved):

contracts/FlowServiceAccount.cdc
```diff
@@ -67,8 +67,12 @@ pub contract FlowServiceAccount {
         }
 
         let tokenVault = self.defaultTokenVault(acct)
-        let feeVault <- tokenVault.withdraw(amount: self.transactionFee)
-
+        var feeAmount = self.transactionFee
+        if self.transactionFee > tokenVault.balance {
+            feeAmount = tokenVault.balance
+        }
```
turbolent, line 73: I guess it would be nice to have `min` and `max` functions for all number types

turbolent (approved):

PR #223, opened 2021-07-27
removing added `address` field from `MachineAccountInfo`

joshuahannan:
We need to be able to upgrade the contract, so we can't include the address field because it wasn't in the original version.

Removed the part of the tests that compared the address field since it doesn't exist any more


jordanschalm (commented):
@joshuahannan If there isn't one already, could you include in this PR a mechanism to read the machine account address (from a script) given the staking account address or node ID? 

This change is breaking for @kc1116's [script](https://github.com/dapperlabs/flow-mgmt/pull/57#discussion_r682171290) without a new way to get that info.

joshuahannan:
Yep, I've included on in my latest changes

jordanschalm (approved):
🎸

Issue #222, opened 2021-07-22
BUG: Ability to register nodes with duplicate node IDs


jordanschalm:
### Problem

Nodes are identified by a 32-byte identifier, provided as input by the node operator at the time of registration. In `flow-go` it is represented as a `[32]byte`, in `flow-core-contracts` it is represented as a hex-encoded `String`. 

Node IDs must be unique. The smart contract enforces uniqueness by using the node ID as the key in a dictionary, however this results in uniqueness being defined by the string representation, which means two different operators could input (eg.) `Abcd` and `abcd` and both would be accepted, despite both referring to the same identifier.

### Acceptance Criteria

* Node ID uniqueness should be enforced

joshuahannan:
How can we enforce this in the smart contract, since we take the arguments as strings already?

jordanschalm:
> How can we enforce this in the smart contract, since we take the arguments as strings already?

One way would be to enforce an unambiguous representation, for example requiring all inputs be lowercase or converting them to lowercase.

joshuahannan:
So would I just iterate through each character and make sure it is a number or lowercase hex character?
```
0,1,2,3,4,5,6,7,8,9,a,b,c,d,e,f
```

Is there a better way to do it than that?

PR #224, opened 2021-08-02
Adds approved list and node id hex restriction

joshuahannan:
Contains two upgrades to the core contracts

## Separate approve list for node IDs
* closes https://github.com/onflow/flow-core-contracts/issues/225
* Affects `FlowIDTableStaking` and `FlowEpoch`
* The staking contract now contains a separate list of node IDs that the admin has approved, instead of requiring the admin to send the list whenever they end the staking auction
* The list stays constant between epochs, and the admin only needs to submit a new list when they intend to change it
* The `endStakingAuction` methods in the staking and epoch contract no longer require id list parameters

## Restrict node IDs to only be number or lowercase hex characters
* Closes https://github.com/onflow/flow-core-contracts/issues/222
* Make it so node operators cannot submit duplicate node IDs because of the hex-encoding


contracts/FlowIDTableStaking.cdc
```diff
@@ -728,6 +729,15 @@ pub contract FlowIDTableStaking {
             return <-node
         }
 
+        /// Sets a list of approved node IDs for the next epoch
+        /// Nodes not on this list will be unstaked at the end of the staking auction
+        /// and not considered to be a proposed/staked node
+        pub fun setApprovedList(_ nodeIDs: [String]) {
+            let list = FlowIDTableStaking.account.load<[String]>(from: /storage/idTableApproveList)
+
+            FlowIDTableStaking.account.save<[String]>(nodeIDs, to: /storage/idTableApproveList)
```
jordanschalm, line 738: Is there a reason we need to load it first?

transactions/epoch/admin/advance_view.cdc
```diff
@@ -17,7 +17,7 @@ transaction(phase: String) {
                 // for now we just set approved to true for all
                 approvedIDs[id] = true
             }
```
jordanschalm, line 19: Can we remove remove lines 10-19?

jordanschalm (approved):

contracts/FlowIDTableStaking.cdc
joshuahannan: If you try to save a value to storage when there is already something there, the save will fail

transactions/epoch/admin/advance_view.cdc
joshuahannan: yep, removed

PR #227, opened 2021-08-04
Allows reset Epoch during staking auction and adds function to update networking address

joshuahannan:
# Reset Epoch
* Epochs can now be reset during the staking auction in addition to epoch setup or epoch commit
* If the epoch is reset during the staking auction, the staking auction is ended and invalid nodes are removed, but the epoch setup phase is not started and the epoch setup event is not emitted since the epoch is being reset anyway

# Update Networking Address
* With this change, node operators can update their networking address at any time during the staking auction
* They cannot update it during epoch setup and epoch commit
* Added the wrapper functions and transactions for this to `LockedTokens` and `FlowStakingCollection` also
* Added tests for each contract


contracts/FlowIDTableStaking.cdc
```diff
@@ -417,6 +418,22 @@ pub contract FlowIDTableStaking {
             self.id = id
         }
 
+        /// Change the node's networking address to a new one
+        pub fun updateNetworkingAddress(_ newAddress: String) {
+            pre {
+                FlowIDTableStaking.stakingEnabled(): "Cannot update networking address if the staking auction isn't in progress"
+                newAddress.length > 0 && newAddress.length <= 510: "The networkingAddress must be less than 255 bytes (510 hex characters)"
```
jordanschalm, line 425: The address isn't hex characters, it's a URL 🤔

jordanschalm (approved):

Issue #233, opened 2021-08-20
Update Staking Contract to Use Transaction Fees for Staking Rewards


joshuahannan:
### Issue To Be Solved

Flow rewards are meant to include transaction fees, so we need a way to move the transaction fees from the `FlowFees` contract to the staking contract so they can be used instead of minting even more tokens and burning the fees. It allows for a more direct transfer.

### Suggest A Solution

* Figure out where the `FlowFees.Administrator` resource is
* Organize for the `FlowFees.Administrator` to be moved to the staking contract account
* In the payRewards function, using the Fees admin resource to withdraw all the fees that have accumulated in the fees vault
* Ensure that all of the fees tokens are used before any more tokens are minted
* Maybe mint all the extra tokens in one batch instead of individually for each staker

cc @laynelafrance @Kay-Zee

laynelafrance:
@joshuahannan in addition, @Kay-Zee pointed out that this should include the automation of rewards as it's currently hard coded. Would you like me to make a separate issue for that? It would include calculating rewards rather than taking an input and then reducing the mint by the amount fees contribute to the payout pool for that epoch.

joshuahannan:
We could include that also, but I was told that for this version of epochs, we weren't going to automate rewards calculation and payment yet. Has that changed?

PR #235, opened 2021-08-24
Uses Transaction Fees to pay rewards instead of minting new tokens

joshuahannan:
Closes https://github.com/onflow/flow-core-contracts/issues/233

Adds a change to use fees from the `FlowFees` smart contract to pay rewards instead of minting new tokens
The rewards payments take from the fees vault first, and once that has run out, it will start minting new tokens.
If there are still fees left after all rewards have been paid, those are burned.

Adds an event documenting the details of the reward payments


lib/go/contracts/contracts.go
```diff
@@ -315,3 +316,33 @@ func TESTFlowStakingCollection(
 
 	return []byte(code)
 }
+
+func TESTFlowFees(fungibleTokenAddress, flowTokenAddress string) []byte {
```
janezpodhostnik, line 320:
```suggestion
func TestFlowFees(fungibleTokenAddress, flowTokenAddress string) []byte {
```

contracts/FlowIDTableStaking.cdc
```diff
@@ -836,22 +838,58 @@ pub contract FlowIDTableStaking {
...
+                }
+                
+                emit RewardsPaid(nodeID: rewardBreakdown.nodeID, amount: nodeReward)
 
                 for delegator in rewardBreakdown.delegatorRewards.keys {
                     let delRecord = nodeRecord.borrowDelegatorRecord(delegator)
-                    delRecord.tokensRewarded.deposit(from: <-flowTokenMinter.mintTokens(amount: rewardBreakdown.delegatorRewards[delegator]!))
-                    emit DelegatorRewardsPaid(nodeID: rewardBreakdown.nodeID, delegatorID: delegator, amount: rewardBreakdown.delegatorRewards[delegator]!)
+                    let delegatorReward = rewardBreakdown.delegatorRewards[delegator]!
+                    if delegatorReward <= feeVault.balance {
```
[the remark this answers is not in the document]
joshuahannan: I don't think I really want to make it a separate function because it just makes it less clear what it is doing. I did simplify the logic in my latest commit though

janezpodhostnik (approved):

contracts/FlowIDTableStaking.cdc
```diff
@@ -836,22 +838,54 @@ pub contract FlowIDTableStaking {
...
             let allNodeIDs = FlowIDTableStaking.getNodeIDs()
 
             let flowTokenMinter = FlowIDTableStaking.account.borrow<&FlowToken.Minter>(from: /storage/flowTokenMinter)
                 ?? panic("Could not borrow minter reference")
 
             for rewardBreakdown in rewardsBreakdownArray {
                 let nodeRecord = FlowIDTableStaking.borrowNodeRecord(rewardBreakdown.nodeID)
-                nodeRecord.tokensRewarded.deposit(from: <-flowTokenMinter.mintTokens(amount: rewardBreakdown.nodeRewards))
-                emit RewardsPaid(nodeID: rewardBreakdown.nodeID, amount: rewardBreakdown.nodeRewards)
+                let nodeReward = rewardBreakdown.nodeRewards
```
jordanschalm, line 855: I wonder if it would be better to sum up the the tokens we need to mint first, then mint them at once and combine them with the fees vault.

It looks like we're minting for every node and delegator, and every mint will emit an event -- we could get rid of a lot of unneeded events. We could also then take the conditional out of these this codepath and have the payout part be a lot simpler.

jordanschalm (approved):
LGTM

contracts/FlowIDTableStaking.cdc
joshuahannan: I didn't do it that way because then we'd have to iterate through everything twice, which might be more excessive than emitting a minted event for every staker. What do you think?

I guess we could just check what the total payout is and subtract the fee vault balance from that and mint the result.

Issue #241, opened 2021-09-09
QC Generation assumes all cluster participants have voted


jordanschalm:
## Context

When generating a QC, the `FlowClusterQC` contract adds all votes which match the majority-selected message to a `ClusterQC` object, which is then used as the root QC for bootstrapping cluster consensus in the subsequent epoch. In this process, the logic makes an unsafe assumption that all members of a cluster have submitted any vote (by force-unwrapping the vote's contents [here](https://github.com/onflow/flow-core-contracts/blob/master/contracts/epochs/FlowClusterQC.cdc#L151-L155)). If any cluster member does not submit a vote, this causes the `generateQuorumCertificate` function to panic, which permanently blocks the system chunk.

Error on Execution Node:
```
{"level":"error","node_role":"execution","node_id":"5eb78ef718eb193d38a533d279d5853e7e80191acf1ebf4c2598acc19a098b30","engine":"computation","component":"block_computer","error_message":"[Error Code: 1101] cadence runtime error Execution failed:\nerror: unexpectedly found nil while forcing an Optional value\n   --> f8d6e0586b0a20c7.FlowClusterQC:152:35\n","block_id":"ad158f44dff3a74210845adde37c58795490c5bb15e815371d2ece04fc035c22","system_chunk_error":true,"critical_error":true,"time":"2021-09-09T01:43:18Z","message":"error executing system chunk transaction"}
```

## Definition of Done

* When building a QC, ignore votes which have not been submitted
* Add a test for the case where 1 node has not submitted a vote

PR #244, opened 2021-09-09
Don't create machine accounts for non-consensus or collection nodes

joshuahannan:
Closes https://github.com/onflow/flow-core-contracts/issues/241

* Ignores votes that have not been submitted for qc and adds a test case for this

* Only call create machine account function if the node is collection or consensus so the account that is registering doesn't have to pay account creation fees


JeffreyDoyle (approved):

joshuahannan:
@jordanschalm Can you check out this PR instead? This includes the QC fix as well as a staking collection update for a machine account bug

jordanschalm (approved):

PR #245, opened 2021-09-15
update dependencies

joshuahannan:
This PR updates the dependencies to the latest cadence, flow-go-sdk, emulator, and flow-go

The cadence update required the cadence string function to be changed to support an error return value
The flow-go update required that we do not use the service account for any of the tests because it has conflicting storage paths with the tests

Also added an upgrade to the `LockedTokens` contract to allow extra `TokenHolder` objects to be created after the first and a  transaction to create an extra one for a new account, along with a test


contracts/LockedTokens.cdc
```diff
@@ -285,8 +285,12 @@ pub contract LockedTokens {
...
             self.address = lockedAddress
             self.tokenManager = tokenManager
-            self.nodeStakerProxy = nil
-            self.nodeDelegatorProxy = nil
+
+            // Create a new staker proxy that can be accessed in transactions
+            self.nodeStakerProxy = LockedNodeStakerProxy(tokenManager: self.tokenManager)
+
+            // create a new delegator proxy that can be accessed in transactions
+            self.nodeDelegatorProxy = LockedNodeDelegatorProxy(tokenManager: self.tokenManager)
```
Kay-Zee, line 293: So in the case of someone calling `createNodeStaker` after this upgrade, this will just get over-written, right?

joshuahannan: yeah, but I can actually just remove the lines in `createNodeStaker` because they aren't needed anymore

Kay-Zee (approved):
```
            // Create a new staker proxy that can be accessed in transactions
            self.nodeStakerProxy = LockedNodeStakerProxy(tokenManager: self.tokenManager)

            // create a new delegator proxy that can be accessed in transactions
            self.nodeDelegatorProxy = LockedNodeDelegatorProxy(tokenManager: self.tokenManager)
```

Is the major change that i'm concerned with, and looks good to me.

Not sure about the dependency updates though

PR #247, opened 2021-09-23
automatic rewards calculations and payout

joshuahannan:
Updates the epoch contract to calculate and pay rewards automatically every epoch

It would probably be worth upgrading this on testnet and canary soon so we don't have to pay rewards manually for people who are testing


joshuahannan:
Closing this PR in favor of https://github.com/onflow/flow-core-contracts/pull/267
I addressed the comments here though

Issue #248, opened 2021-09-30
Include Ability to withhold rewards from inactive nodes


joshuahannan:
### Issue To Be Solved
Some nodes are inactive for large parts of epochs, or do not update when a spork happens. 
Need to be able to withhold rewards from these nodes.

### Suggest A Solution

* Admin can submit a list of node IDs and how much their rewards should be decreased at the end of the epoch
* This list is cleared at the end of the epoch
* Delegators to a punished node are also penalized
* Withheld rewards are redistributed among the rest of the stakers in the network

PR #249, opened 2021-10-01
Adds non-operational rewards withholding

joshuahannan:
Closes https://github.com/onflow/flow-core-contracts/issues/248

Some nodes do not upgrade quickly enough for sporks or other network upgrades
and some nodes spend a lot of time down

This PR introduces a mechanism for the staking admin to mark certain nodes as inactive and marks how much their rewards should be decreased by

Nodes that are marked as inactive will have their rewards decreased by the amount specified by the admin when rewards are calculated. The delegators to an inactive node will also have their rewards decreased. 

The withheld rewards are then distributed among the rest of the non-offending nodes and delegators.

After rewards are calculated, the inactive nodes list is cleared so the node has the opportunity to fix their misbehavior before the next rewards payment.

This process will be controlled manually first, but is a small work towards future automated slashing in the future.


bluesign:
hi @joshuahannan , first of all this is a great idea. 

I have a question about `The delegators to an inactive node will also have their rewards decreased.` part.  For delegator rewards, there is a fixed percentage fee node taking for staking on behalf of delegators, wouldn't be more fair to remove the penalty first from that bucket, then apply it to delegators? Otherwise it is a bit of unnecessary heat on node operators. ( If it will be really implemented with uptime in the future instead of admin involvement )

joshuahannan:
@bluesign Just to clarify, you are suggesting that if the amount of rewards withheld is somewhere greater than zero but less than 100%, we take it from the node's delegator cuts first before taking from the delegators? Or are you suggesting something else? I'm a little confused

bluesign:
@joshuahannan , sorry I could state more clear. 

Now there is a 8% cut for code operator. 92% is going to the delegator. If there would be 2% amount to withheld. I think first this should go from cut, so node operator should get 6%, delegator should still get 92%. 

So if it is 8% -> 8% withheld 0% cut 92% delegators. 

Only after 8% of penalty, effecting delegators revenue can be beneficial to the ecosystem. Users don't have much effect about node operators uptime, also this allows node operators to be more motivated to keep uptime high.

contracts/FlowIDTableStaking.cdc
```diff
@@ -886,26 +902,90 @@ pub contract FlowIDTableStaking {
...
+            var withheldRewardsScale = sumRewardsWithheld / totalStaked
+
             /// iterate through all the nodes to pay
             for nodeID in allNodeIDs {
+                if nonOperationalNodes[nodeID] != nil { continue }
+
                 let nodeRecord = FlowIDTableStaking.borrowNodeRecord(nodeID)
 
-                var nodeRewardAmount = nodeRecord.tokensStaked.balance * totalRewardScale
+                var nodeRewardAmount = nodeRecord.tokensStaked.balance * (totalRewardScale + withheldRewardsScale)
```
rheaplex, line 988: `(totalRewardScale + withheldRewardsScale)` is calculated again on line 1000 below. Is it worth calculating it once and assigning it to a local variable, for clarity and ease of future change?

transactions/idTableStaking/admin/set_non_operational.cdc
```diff
@@ -0,0 +1,25 @@
...
+    prepare(acct: AuthAccount) {
+        // borrow a reference to the admin object
+        self.adminRef = acct.borrow<&FlowIDTableStaking.Admin>(from: FlowIDTableStaking.StakingAdminStoragePath)
+            ?? panic("Could not borrow reference to staking admin")
+    }
+
+    execute {
+        let nodeList: {String: UFix64} = {}
+        for id in ids {
+            nodeList[id] = 0.0
```
rheaplex, line 20: This sets the rewards to zero?

rheaplex (approved):
This is good stuff! A few trivial suggestions but that's it.

👍

transactions/idTableStaking/admin/set_non_operational.cdc
joshuahannan: yeah, first version is only going to be setting the rewards to zero. I'll be adding more tests for other amounts soon, but just wanted to get this version in ASAP

contracts/FlowIDTableStaking.cdc
joshuahannan: good suggestion! I'm gonna change that

contracts/FlowIDTableStaking.cdc
```diff
@@ -886,26 +902,91 @@ pub contract FlowIDTableStaking {
...
+                // Each node's rewards can be decreased by a different percentage
+                // Its delegator's rewards are also decreased by the same percentage
+                let rewardDecreasePercentage = nonOperationalNodes[nodeID]!
+
+                // Calculate the normal reward amount, then the rewards left after the decrease
+                var nodeRewardAmount = nodeRecord.tokensStaked.balance * totalRewardScale
+                var nodeRewardsLeft = nodeRewardAmount * rewardDecreasePercentage
+
+                // Add the remaining to the total number of rewards withheld
+                sumRewardsWithheld = sumRewardsWithheld + (nodeRewardAmount - nodeRewardsLeft)
```
jordanschalm, line 945: ⚠️  This is the reverse of what I would expect, based on the comments and variable naming. 

`rewardDecreasePercentage` seems intended to mean the percentage of a node's rewards that will be withheld. For example, `0.2` would result in the node getting 80 tokens and 20 tokens being withheld.

`nodeRewardsLeft` is the reward value the node receives after the withholding. But we are computing it like `nodeRewardAmount * rewardDecreasePercentage`. Which means if we say we want to decrease node A's reward by 20%, we will actually decrease it by 80% (and only pay them 20%).

contracts/FlowIDTableStaking.cdc
```diff
@@ -886,26 +902,91 @@ pub contract FlowIDTableStaking {
...
+            for nodeID in nonOperationalNodes.keys {
+                let nodeRecord = FlowIDTableStaking.borrowNodeRecord(nodeID)
+
+                // Each node's rewards can be decreased by a different percentage
+                // Its delegator's rewards are also decreased by the same percentage
+                let rewardDecreasePercentage = nonOperationalNodes[nodeID]!
+
+                // Calculate the normal reward amount, then the rewards left after the decrease
+                var nodeRewardAmount = nodeRecord.tokensStaked.balance * totalRewardScale
+                var nodeRewardsLeft = nodeRewardAmount * rewardDecreasePercentage
```
jordanschalm, line 942:
```suggestion
                var nodeRewardsAfterWithholding = nodeRewardAmount * rewardDecreasePercentage
```

contracts/FlowIDTableStaking.cdc
```diff
@@ -886,26 +902,91 @@ pub contract FlowIDTableStaking {
...
+
+                // Iterate through all the withheld node's delegators
+                // and calculate their decreased rewards as well
+                for delegator in nodeRecord.delegators.keys {
+                    let delRecord = nodeRecord.borrowDelegatorRecord(delegator)
+
+                    // Calculate the amount of tokens that this delegator receives
+                    // minus the decrease from the non-operational node
+                    var delegatorRewardAmount = delRecord.tokensStaked.balance * totalRewardScale
+                    var delegatorRewardsLeft = delegatorRewardAmount * rewardDecreasePercentage
```
jordanschalm, line 957:
```suggestion
                    var delegatorRewardsAfterWithholding = delegatorRewardAmount * rewardDecreasePercentage
```

lib/go/test/flow_idtable_staking_test.go
```diff
@@ -2303,3 +2303,246 @@ func TestIDTableStaking(t *testing.T) {
...
+	var cutPercentage interpreter.UFix64Value = 8000000       // 08.0 %
+
+	// Create new keys for the ID table account
+	IDTableAccountKey, IDTableSigner := accountKeys.NewWithSigner()
+	idTableAddress, feesAddr := deployStakingContract(t, b, IDTableAccountKey, IDTableSigner, env, true)
+
+	env.IDTableAddress = idTableAddress.Hex()
+	env.FlowFeesAddress = feesAddr.Hex()
+
+	var totalStaked interpreter.UFix64Value = 510000000000000
```
jordanschalm, line 2330: a comment here for what this value is (like above) would be great, since `UFix64` values are offset

lib/go/test/flow_idtable_staking_test.go
```diff
@@ -2303,3 +2303,246 @@ func TestIDTableStaking(t *testing.T) {
...
+			[]crypto.Signer{b.ServiceKey().Signer(), IDTableSigner},
+			false,
+		)
+
+		result := executeScriptAndCheck(t, b, templates.GenerateGetNonOperationalListScript(env), nil)
+
+		idArray := result.(cadence.Array).Values
+		assert.Equal(t, len(idArray), 2)
+		assert.Equal(t, idArray[0], CadenceString(ids[0]))
+		assert.Equal(t, idArray[1], CadenceString(ids[2]))
```
jordanschalm, line 2452: This is stored as a map and we are making assumptions about its order. AFAIK Cadence maps are un-ordered by definition. I think they used to happen to iterate in insert-order, but this is no longer true with some recent changes. 
```suggestion
		result := executeScriptAndCheck(t, b, templates.GenerateGetNonOperationalListScript(env), nil)

		idArray := result.(cadence.Array).Values
		assert.Equal(t, len(idArray), 2)
		assert.ElementsMatch(t, idArray, cadence.NewArray([]cadence.Value{CadenceString(ids[0]), CadenceString(ids[2])}))
```

lib/go/test/flow_idtable_staking_test.go
```diff
@@ -2303,3 +2303,246 @@ func TestIDTableStaking(t *testing.T) {
...
+
+		// Calculate Rewards
+		// All the nodes staked 500k FLOW and all the delegators staked 10k FLOW
+		// so the calculations are slightly easier
+		// Node 0 has 5 delegators, node 1 has 5 delegators,
+		// and the other nodes have no delegators
+		// Tested witholding rewards from one of the nodes with delegators
+		// and one of the non-delegated-to nodes
+
+		var delegatorCommitment interpreter.UFix64Value = 1000000000000
```
jordanschalm, line 2498: For this and the following `interpreter.UFix64Value`-typed values: please add a comment noting what the value is or maybe a helper function which converts a more readable numeric representation to the `interpreter.UFix64Value` type

transactions/idTableStaking/admin/set_non_operational.cdc
jordanschalm: I feel like this should be a transaction argument.

lib/go/test/flow_idtable_staking_test.go
```diff
@@ -2303,3 +2303,246 @@ func TestIDTableStaking(t *testing.T) {
...
+				assertEqual(t, CadenceUFix64(totalNodeReward.String()), result)
+			}
+		}
+
+		// Delegators from Node 1
+		for i := 1; i < 6; i++ {
+			result = executeScriptAndCheck(t, b, templates.GenerateGetDelegatorRewardsScript(env), [][]byte{jsoncdc.MustEncode(cadence.String(ids[1])), jsoncdc.MustEncode(cadence.UInt32(i))})
+			assertEqual(t, CadenceUFix64(totalDelegatorReward.String()), result)
+		}
+	})
```
jordanschalm, line 2547: Could we add a test (or assertion to an existing) test which checks that the total amount of rewards is the same regardless of any withholdings? 

I'm not too familiar, but I found it a bit hard to convince myself of this reading through the reward calculation logic.

contracts/FlowIDTableStaking.cdc
```diff
@@ -761,6 +761,22 @@ pub contract FlowIDTableStaking {
             FlowIDTableStaking.account.save<[String]>(nodeIDs, to: /storage/idTableApproveList)
         }
 
+        /// Sets a list of node IDs who will not receive rewards for the current epoch
+        /// This is used during epochs to punish nodes who have poor uptime 
+        /// or who do not update to latest node software quickly enough
+        pub fun setNonOperationalNodesList(_ nodeIDs: {String: UFix64}) {
```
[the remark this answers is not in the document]
joshuahannan: Can you elaborate? I'm not sure what you mean by this.

lib/go/test/flow_idtable_staking_test.go
laynelafrance: > Could we add a test (or assertion to an existing) test which checks that the total amount of rewards is the same regardless of any withholdings?
➕

contracts/FlowIDTableStaking.cdc
joshuahannan: thats a good point that some might interpret it that way. That isn't what I intended, but I'll try to think of better naming

lib/go/test/flow_idtable_staking_test.go
joshuahannan: moved it to near the end of the test and made it a calculation

contracts/FlowIDTableStaking.cdc
joshuahannan: I'm changing it to `rewardDecreaseToPercentage`

laynelafrance:
clarify staking admin and service account relationship , looks great

laynelafrance (approved):

lib/go/test/flow_idtable_staking_test.go
joshuahannan: Good catch. Updated. 👍

lib/go/test/flow_idtable_staking_test.go
joshuahannan: Good catch, I added that and there was a missed calculation in there. The calculation and rewards paid are correct now

joshuahannan:
I addressed all the important PR comments so this should be good to deploy. I'm gonna try to clean up some more things in the tests tomorrow morning before we merge though

lib/go/test/flow_idtable_staking_test.go
```diff
@@ -2495,37 +2499,46 @@ func TestIDTableRewardsWitholding(t *testing.T) {
...
-		nodeRewardPlusDelegators := nodeRewardWithoutWithold.Plus(delegatorRewardNodeCut.Mul(numDelegators)).(interpreter.UFix64Value)
-
-		// For multiplying the reward amounts
-		var numWithheldNodes interpreter.UFix64Value = 200000000
+		nodeRewardPlusDelegators := nodeRewardWithoutWithold.Plus(delegatorRewardNodeCut.Mul(numDelegatorsForNode0)).(interpreter.UFix64Value)
 
 		// Figure out the sum of tokens withheld from all punished nodes
-		amountWithheld := nodeRewardWithoutWithold.Mul(numWithheldNodes).Plus(delegatorReward.Mul(numDelegators)).(interpreter.UFix64Value)
+		amountWithheld := nodeRewardWithoutWithold.Mul(numWithheldNodes).Plus(delegatorReward.Mul(numDelegatorsForNode0)).(interpreter.UFix64Value)
+		fmt.Println(amountWithheld)
```
jordanschalm, line 2530: debug log

jordanschalm (approved):

contracts/FlowIDTableStaking.cdc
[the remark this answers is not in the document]
jordanschalm: I meant pretty much what you added: https://github.com/onflow/flow-core-contracts/pull/249/files#diff-65336be374bb3fc9ad7b822243e065a389d73e758c7c16223e52fc3181cea59bR767-R769 :)

PR #250, opened 2021-10-04
Use borrow instead of load/copy

joshuahannan:
Including two changes in this PR to make the epoch contract perform better:

* Use `borrow` instead of `load` or `copy` for dictionaries in account storage so we don't have to copy the entire dictionary
* In `heartbeat` only get the current block info and epoch metadata in the relevant block of the switch statement.


jordanschalm (approved):

ramtinms (approved):

turbolent (approved):
Nice!

zhangchiqing (approved):

PR #251, opened 2021-10-04
Transaction for registering multiple nodes with the staking collection

joshuahannan:
The parameters to this transaction are expected to match array indexes for the metadata

### DIFF
--- a/transactions/stakingCollection/register_multiple_nodes.cdc
+++ b/transactions/stakingCollection/register_multiple_nodes.cdc
@@ -1,0 +1,41 @@
+import FlowStakingCollection from 0xSTAKINGCOLLECTIONADDRESS
+
+/// Registers a delegator in the staking collection resource
+/// for the specified node information and the amount of tokens to commit
+
+transaction(ids: [String],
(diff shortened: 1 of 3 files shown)


transactions/stakingCollection/register_multiple_nodes.cdc
```diff
@@ -0,0 +1,41 @@
...
+/// Registers a delegator in the staking collection resource
+/// for the specified node information and the amount of tokens to commit
+
+transaction(ids: [String],
+            roles: [UInt8],
+            networkingAddresses: [String],
+            networkingKeys: [String],
+            stakingKeys: [String],
+            amounts: [UFix64],
+            publicKeys: [[String]?]) {
```
Kay-Zee, line 12: why do we need `?` here, if we're always requiring at least 1 pub key?

joshuahannan: We don't require public keys if they are registering nodes that don't require machine accounts, so we allow them to make it `nil` here

Issue #260, opened 2021-11-18
Configurable success threshold in FlowDKG contract


jordanschalm:
## Context

The FlowDKG contract determines whether the DKG process succeeded by observing that the number of agreeing DKG key vector submissions  exceeds some threshold (the safety margin) . Currently that threshold is about 50%, which is the threshold of success for the underlying DKG. 

In the short-term, random beacon keys are required to submit a consensus vote. Therefore the proportion of nodes successfully completing the DKG directly impacts finalization performance. Even <90% completion can significantly degrade finalization. This will be mitigated by signature decoupling in Crypto v2.

To mitigate the problem in the interim we would like to be able to configure the threshold at which FlowDKG contract considers the DKG to have succeeded.

**Target: include this change in Dec 1 Canary spork**

## Definition of Done

* Make the FlowDKG contract support a configurable DKG success safety margin

cc @tarakby

tarakby:
- In order to avoid confusion in the docs and conversations, I suggest to decouple the threshold value `t` (used by the threshold signature and DKG) and the safety rate `r` used by the DKG contract. Currently `r` is set to `t+1` ( note it is not set to `t`). 

- I suggest to keep the definition of `t` in the DKG contract as it is now, and introduce the new value `r`. 
       -  the contract should make the sanity check `n >= r >=  t+1`  ( `r` should not be less than the "native" implementation threshold ). 
       - `r` can be defined as an integer `r = floor (percentage *  n)`  with `n` being the number of nodes and `percentage` a value between 0 and 1 ( I guess fixed for 0.95 for now)

joshuahannan:
@tarakby @jordanschalm Should we allow the admin to change this threshold whenever they want or only when the DKG isn't enabled?

jordanschalm:
I think we should allow changes only when the DKG is not enabled.

PR #261, opened 2021-11-19
add configurable DKG success threshold and tests

joshuahannan:
Closes https://github.com/onflow/flow-core-contracts/issues/260

Adds a configurable success threshold to the DKG contract. This is a percentage that the admin can set.
When checking to see if the DKG has succeeded, the contract will use whatever success threshold is greater, the native threshold (~50%), or the safe threshold that the admin has set. The contract calculates whatever number of nodes is needed from these values.

Added tests


joshuahannan:
@jordanschalm @tarakby Here are the configurable success threshold changes. I added some tests that make sure it works properly and those are passing. I'll probably work on some other tests too, but I'm feeling good about it so far

contracts/epochs/FlowDKG.cdc
```diff
@@ -202,6 +202,19 @@ pub contract FlowDKG {
     /// The Admin resource provides the ability to begin and end voting for an epoch
     pub resource Admin {
 
+        /// Sets the optional safe DKG success threshold
+        pub fun setSafeSuccessThreshold(newThresholdPercentage: UFix64?) {
```
jordanschalm, line 206: why is it an optional?

contracts/epochs/FlowDKG.cdc
```diff
@@ -331,6 +344,34 @@ pub contract FlowDKG {
         return self.uniqueFinalSubmissions
     }
 
+    /// Gets the native threshold that the submission count needs to exceed to be considered complete
+    pub fun getNativeSuccessThreshold(): UInt64 {
+        return UInt64((self.getConsensusNodeIDs().length-1)/2)
+    }
+
+    /// Gets the safe threshold that the submission count needs to exceed to be considered complete
+    /// (always greater than or equal to the native success threshold + 1)
```
jordanschalm, line 353:
```suggestion
    /// (always greater than or equal to the native success threshold)
```

If there is no safetyThreshold set, we just return the native threshold.

contracts/epochs/FlowDKG.cdc
```diff
@@ -202,6 +202,19 @@ pub contract FlowDKG {
     /// The Admin resource provides the ability to begin and end voting for an epoch
     pub resource Admin {
 
+        /// Sets the optional safe DKG success threshold
+        pub fun setSafeSuccessThreshold(newThresholdPercentage: UFix64?) {
+            pre {
+                !FlowDKG.dkgEnabled: "Cannot set the dkg success threshold while the DKG is enabled"
```
jordanschalm, line 208:
```suggestion
            pre {
                !FlowDKG.dkgEnabled: "Cannot set the dkg success threshold while the DKG is enabled"
                newThresholdPercentage < 1.0: "Threshold must be a percentage value (0, 1)"
```

lib/go/test/flow_dkg_test.go
```diff
@@ -566,8 +566,76 @@ func TestDKG(t *testing.T) {
...
+
+		signAndSubmit(
+			t, b, tx,
+			[]flow.Address{b.ServiceKey().Address, DKGAddress},
+			[]crypto.Signer{b.ServiceKey().Signer(), DKGSigner},
+			false,
+		)
+	})
+
+	t.Run("Should be able to set the safe threshold while the DKG is disabled", func(t *testing.T) {
```
jordanschalm, line 597: Could we also have a test for:
* setting `safeThreshold > 1` should revert
* setting `safeThreshold < 0.5` should result in the "native threshold" being used

transactions/dkg/scripts/get_thresholds.cdc
```diff
@@ -0,0 +1,17 @@
...
+
+pub struct Thresholds {
+    pub let native: UInt64
+    pub let safe: UInt64
+    pub let safePercentage: UFix64
+
+    init() {
+        self.native = FlowDKG.getNativeSuccessThreshold()
+        self.safe = FlowDKG.getSafeSuccessThreshold()
+        self.safePercentage = FlowDKG.getSafeThresholdPercentage() ?? 0.0
```
jordanschalm, line 11: Neat idea instantiating these like this

contracts/epochs/FlowDKG.cdc
joshuahannan: It is an optional so that the admin can clear it if it isn't needed any more

lib/go/test/flow_dkg_test.go
joshuahannan: I just included in the pre-condition for the setSafeThreshold function that it has to be > 0.5

contracts/epochs/FlowDKG.cdc
```diff
@@ -202,6 +202,21 @@ pub contract FlowDKG {
     /// The Admin resource provides the ability to begin and end voting for an epoch
     pub resource Admin {
 
+        /// Sets the optional safe DKG success threshold
+        /// Set the threshold to an optional if it isn't needed
+        pub fun setSafeSuccessThreshold(newThresholdPercentage: UFix64?) {
+            pre {
+                !FlowDKG.dkgEnabled: "Cannot set the dkg success threshold while the DKG is enabled"
+                newThresholdPercentage == nil || (newThresholdPercentage! > 0.5 && newThresholdPercentage! < 1.0): "The threshold percentage must be between 0.5 and 1.0"
```
tarakby, line 210: I don't think we should be checking for `> 0.5` here. `>= 0` is enough. The only condition is that the final contract threshold is larger than the native threshold (guaranteed by `getSafeSuccessThreshold`). 

Please note the value `0.5` is only a choice we made for this implementation and should only be reflected inside `getNativeSuccessThreshold`. It can be made larger or smaller if we want to. 
Moreover, I only use `0.5` in discussions for simplicity, but the constant is slightly different. `0.5` should not be used in code or formal documentation.

contracts/epochs/FlowDKG.cdc
```diff
@@ -331,6 +346,35 @@ pub contract FlowDKG {
...
+
+    /// Gets the safe threshold that the submission count needs to exceed to be considered complete
+    /// (always greater than or equal to the native success threshold)
+    pub fun getSafeSuccessThreshold(): UInt64 {
+        var threshold = self.getNativeSuccessThreshold()
+
+        // Get the safety rate percentage
+        if let safetyRate = self.getSafeThresholdPercentage() {
+
+            let safeThreshold = UInt64(safetyRate * UFix64(self.getConsensusNodeIDs().length))
```
tarakby, line 362: This is fine but it's slightly different than what I wrote in the issue (the contract requires the floor of (rate*nodes) + 1  instead of just the floor). 
It's fine, but we should keep this into account when debugging/testing.

lib/go/test/flow_dkg_test.go
```diff
@@ -566,8 +566,103 @@ func TestDKG(t *testing.T) {
...
+
+		signAndSubmit(
+			t, b, tx,
+			[]flow.Address{b.ServiceKey().Address, DKGAddress},
+			[]crypto.Signer{b.ServiceKey().Signer(), DKGSigner},
+			false,
+		)
+	})
+
+	t.Run("Should not be able to set the safe threshold less than 0.5 or greater than 1", func(t *testing.T) {
```
tarakby, line 597: I guess the 0.5 should be adjusted to the native threshold.

transactions/dkg/scripts/get_thresholds.cdc
```diff
@@ -0,0 +1,17 @@
...
+
+pub struct Thresholds {
+    pub let native: UInt64
+    pub let safe: UInt64
+    pub let safePercentage: UFix64
+
+    init() {
+        self.native = FlowDKG.getNativeSuccessThreshold()
+        self.safe = FlowDKG.getSafeSuccessThreshold()
+        self.safePercentage = FlowDKG.getSafeThresholdPercentage() ?? 0.5
```
tarakby, line 11: I suggest to initialize this to `0`, which makes the overall threshold be the native one.
```suggestion
        self.safePercentage = FlowDKG.getSafeThresholdPercentage() ?? 0
```

tarakby (commented):
Thanks for the update 🏄🏼 

I made some comments below.

contracts/epochs/FlowDKG.cdc
jordanschalm: @tarakby I've added some comments to the getter methods that hopefully make this clearer

lib/go/test/flow_dkg_test.go
jordanschalm: Updated

transactions/dkg/scripts/get_thresholds.cdc
jordanschalm: Updated

contracts/epochs/FlowDKG.cdc
```diff
@@ -202,6 +202,21 @@ pub contract FlowDKG {
...
+        /// Set the threshold to an optional if it isn't needed
+        pub fun setSafeSuccessThreshold(newThresholdPercentage: UFix64?) {
+            pre {
+                !FlowDKG.dkgEnabled: "Cannot set the dkg success threshold while the DKG is enabled"
+                newThresholdPercentage == nil || (newThresholdPercentage! > 0.5 && newThresholdPercentage! < 1.0): "The threshold percentage must be between 0.5 and 1.0"
+            }
+
+            FlowDKG.account.load<UFix64>(from: /storage/flowDKGSafeThreshold)
+
+            if let percentage = newThresholdPercentage {
```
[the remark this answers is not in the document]
jordanschalm: Added a test to confirm this

jordanschalm:
@tarakby Made some updates, could you take another look?

contracts/epochs/FlowDKG.cdc
```diff
@@ -346,13 +346,23 @@ pub contract FlowDKG {
...
 
-    /// Gets the native threshold that the submission count needs to exceed to be considered complete
+    /// Gets the native threshold that the submission count needs to exceed to be considered complete [t=floor((n-1)/2)]
+    /// This function returns the NON-INCLUSIVE lower bound of honest participants.
+    /// For the DKG to succeed, the number of honest participants must EXCEED this threshold value.
+    /// 
+    /// Example:
+    /// We have 10 DKG nodes (n=10)
+    /// The threshold value is t=floor(10-1)/2) (t=4)
+    /// There must be AT LEAST 5 honest nodes for the DKG to succeed
```
tarakby, line 358: 👌🏼

contracts/epochs/FlowDKG.cdc
```diff
@@ -369,7 +379,10 @@ pub contract FlowDKG {
         return threshold
     }
 
-    /// Gets the safe threshold percentage 
+    /// Gets the safe threshold percentage. This value must be either nil (semantically: 0) or in [0, 1.0)
+    /// This safe threshold is used to artificially increase the DKG participation requirements to 
+    /// ensure a lower-bound number of Random Beacon Committee members (beyond the bare minimum required
+    /// by the DKG protocol).
```
tarakby, line 387: 👌🏼

tarakby (approved):

jordanschalm:
Hey @joshuahannan we made a few more changes to this last week. Do you mind taking a look before merging?

joshuahannan (commented):
I can't approve it since I am the author, but it looks good to me! I am ok to merge

jordanschalm (approved):

PR #262, opened 2021-11-24
Add new transactions for interacting with staking allow-list

jordanschalm:
Adds two new transactions for interacting with the allow-list, based on experience from sporking operations. 

cc @vishalchangrani


joshuahannan (approved):
Thanks for adding these!

PR #263, opened 2021-12-01
Adds transaction to stake all rewards for all nodes and delegators in a staking collection

### LINKED ISSUE #255: StakingCollection: Add transaction to restake all staking rewards
### Background

A common task for many node operators and delegators is to periodically re-stake their staking rewards. Since the Staking Collection allows for multiple nodes and delegations to be managed by a single account, we should put in place a prescribed transaction that enables a Staking Collection to execute a single transaction which re-stakes all of their staking rewards.

## Further Questions

- Should the "re-stake all" txn re-stake both node operator rewards and delegator rewards, or should there be two re-stake all txns, one for nodes and one for delegations.

### Deliverabless

- [ ] A "re-stake all" transaction(s) exist and associated functionality implemented.

joshuahannan:
Closes #255


JeffreyDoyle (approved):
Thank you!

Issue #266, opened 2021-12-16
Add correct calculation for automated rewards


joshuahannan:
### Issue To Be Solved
We will be enabling automatic reward calculations and payments soon, so we need the correct calculation.

### Suggest A Solution
* Work from this PR: https://github.com/onflow/flow-core-contracts/pull/247
* Calculation for a specific epoch's rewards should be:
```
totalRewards = (total supply of FLOW from prior epoch * 0.0009387127)
```
* This calculation will probably happen before rewards for the current epoch have actually been minted/paid, so it needs to make sure to include what was calculated in the current epoch in the total supply part of the calculation.
* The percentage used to calculate rewards needs to be configurable by the epoch admin without having to upgrade the smart contract

This will also require a transaction signed by the service account to update the `FlowEpoch.configurableMetadata.FLOWsupplyIncreasePercentage` value so it can be used

joshuahannan:
While testing this, I realized that the UFix64 type in cadence isn’t precise enough to hold the value of 0.0009387127
I am currently testing it with 0.00093871 instead, which is hopefully precise enough.

PR #267, opened 2021-12-21
Automated rewards and calculation

joshuahannan:
Closes https://github.com/onflow/flow-core-contracts/issues/266

Updates the epoch contract to calculate and pay rewards automatically every epoch

The epoch admin can enable or disable automatic rewards. 

If automatic rewards are enabled, the contract calculates the rewards for the next epoch based on the supply increase percentage defined in the contract, which the admin can also control. It also pays the rewards for the current epoch.

If automatic rewards are disabled, the contract will only calculate rewards for the current epoch but will not pay them

`resetEpoch` has been updated to not set a new payout amount because it isn't necessary

 The staking contract has a function to get the fee balance because adding a new import address to the epoch contract was breaking the bootstrapping integration in flow-go, which we'll hopefully be able to fix once this is close to merging.


contracts/epochs/FlowEpoch.cdc
```diff
@@ -460,21 +468,31 @@ pub contract FlowEpoch {
...
 
-        // Calculate the new epoch's payout
-        // disabled until we enable automated rewards calculations
-        // let newPayout = FlowToken.totalSupply * (FlowEpoch.configurableMetadata.FLOWsupplyIncreasePercentage / 52.0)
+        if FlowEpoch.automaticRewardsEnabled() {
+            // Calculate the total supply of FLOW after the current epoch's payout
+            // the calculation includes the tokens that haven't been minted for the current epoch yet
+            let currentPayout = FlowIDTableStaking.getEpochTokenPayout()
+            let feeAmount = FlowIDTableStaking.getFeesBalance()
+            let flowTotalSupplyAfterPayout = FlowToken.totalSupply + (currentPayout - feeAmount)
```
jordanschalm, line 487: I don't think this handles the case where fees exceed the payout.

[We burn the excess fees](https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowIDTableStaking.cdc#L910-L911) in that case, so `flowTotalSupplyAfterPayout` should remain the same value as `FlowToken.totalSupply`.

contracts/epochs/FlowEpoch.cdc
```diff
@@ -437,8 +447,6 @@ pub contract FlowEpoch {
                 FlowEpoch.borrowDKGAdmin().forceEndDKG()
             }
```
jordanschalm, line 449: When we enable auto-rewards on Mainnet, how are rewards paid out during weeks where we have a spork?

contracts/epochs/FlowEpoch.cdc
```diff
@@ -460,21 +468,31 @@ pub contract FlowEpoch {
...
+            // Calculate the total supply of FLOW after the current epoch's payout
+            // the calculation includes the tokens that haven't been minted for the current epoch yet
+            let currentPayout = FlowIDTableStaking.getEpochTokenPayout()
+            let feeAmount = FlowIDTableStaking.getFeesBalance()
+            let flowTotalSupplyAfterPayout = FlowToken.totalSupply + (currentPayout - feeAmount)
 
-        if let payout = newPayout {
-            self.borrowStakingAdmin().setEpochTokenPayout(payout)
+            // Calculate the payout for the next epoch
+            let proposedPayout = flowTotalSupplyAfterPayout * FlowEpoch.configurableMetadata.FLOWsupplyIncreasePercentage
```
jordanschalm, line 490: It seems like this PR is changing the meaning of the `FLOWsupplyIncreasePercentage` config value, but the [documentation](https://github.com/onflow/flow-core-contracts/blob/master/contracts/epochs/FlowEpoch.cdc#L215-L217) isn't changed. 

The [documentation](https://github.com/onflow/flow-core-contracts/blob/master/contracts/epochs/FlowEpoch.cdc#L215-L217) says it is an annualized percentage yield value. Here we're using it as a per-payout percentage yield value. We should choose one of these options and make sure the documentation is consistent.

lib/go/test/flow_epoch_test.go
```diff
@@ -847,9 +847,21 @@ func TestEpochQCDKG(t *testing.T) {
...
+		_ = tx.AddArgument(cadence.NewBool(true))
+
+		signAndSubmit(
+			t, b, tx,
+			[]flow.Address{b.ServiceKey().Address, idTableAddress},
+			[]crypto.Signer{b.ServiceKey().Signer(), IDTableSigner},
+			false,
+		)
+
+		tx = createTxWithTemplateAndAuthorizer(b, templates.GenerateEpochCalculateSetRewardsScript(env), idTableAddress)
```
jordanschalm, line 864: Why is calculating the rewards with an admin transaction needed here? Wouldn't the epoch transition do that automatically, given that we enable automatic reward payouts above?

lib/go/test/flow_epoch_test.go
```diff
@@ -882,15 +894,15 @@ func TestEpochQCDKG(t *testing.T) {
 				startView:             startView + numEpochViews,
 				endView:               startView + 2*numEpochViews - 1,
 				stakingEndView:        startView + numEpochViews + numStakingViews - 1,
-				totalRewards:          "1300000.0",
+				totalRewards:          "6572143.3875",
```
jordanschalm, line 897: We should also check here that the rewards are paid out in the `IDTableStaking` contract, for example ensuring the tokens are moved into the rewards bucket.

jordanschalm (commented):
Nice work!

A few thoughts/questions:
* How are rewards paid out during the week of a spork? If `resetEpoch` does not pay the rewards (which makes sense, I don't think these things should be coupled), we will need to do it manually with a multisig on spork weeks. Reward payout timing could be different on spork weeks.
* If an epoch preparation fails, we stay in the current epoch for the duration of the current spork. If this happens when automatic reward payouts are enabled, we would need to revert to manual reward payouts to maintain the schedule. I'm wondering if there's a way to test this here - for example, exceeding the view of the current epoch while it is in the `EpochSetup` phase, then manually paying rewards.


> The staking contract has a function to get the fee balance because adding a new import address to the epoch contract was breaking the bootstrapping integration in flow-go, which we'll hopefully be able to fix once this is close to merging.

What is breaking here exactly? I can help to fix this on the `flow-go` side.

joshuahannan:
Yeah, we'll need to do a reward payout manually and set the new payout manually as well when we do reset epoch, which still fits with this.

If epoch preparation fails, won't we just run reset epoch and pay rewards manually before the next epoch is supposed to begin? why would it have to last until the next spork?

The argument count in `lib/go/contracts/FlowEpoch` doesn't match flow-go bootstrapping any more since I had to add the fees import address so that was causing it to fail

contracts/epochs/FlowEpoch.cdc
joshuahannan: yep, good catch. I'll change that

contracts/epochs/FlowEpoch.cdc
joshuahannan: I assume we'll do it manually

contracts/epochs/FlowEpoch.cdc
joshuahannan: yeah, that is what it originally was, but we decided to just make it constant for weekly increases. I'll update the comments

lib/go/test/flow_epoch_test.go
joshuahannan: As far as I know, getting view counts in the emulator still isn't possible, so we have been able to test automated epochs yet with the emulator. All the tests still just trigger the transitions on their own

jordanschalm:
> If epoch preparation fails, won't we just run reset epoch and pay rewards manually before the next epoch is supposed to begin? why would it have to last until the next spork?

What I mean is, if we do nothing, the epoch counter will stay the same until the next spork. In this case we must ensure rewards are paid out. We could potentially run resetEpoch alongside manual reward payout, though I don't think this is necessary. 

What is the best way to test this before deploying to Mainnet? It might be as simple as a unit test here which calls `payRewards` multiple times without a corresponding epoch transition. Or we could artificially trigger epoch fallback on benchnet and test paying rewards there.

joshuahannan:
We've done this before, so I don't think it would be an issue. The only difference with these additions to the contract is that we'd just disable automatic rewards, then call the pay rewards functions directly in the staking contract instead of through the epoch contract.

lib/go/test/flow_epoch_test.go
jordanschalm: Ah right, good point. Forgot about that.

jordanschalm (approved):

PR #268, opened 2022-01-03
Set staking auction length explicitly in `resetEpoch`

jordanschalm:
`resetEpoch` currently sets the staking auction length based on the config in the contract, rather than inputs to the function, which can be unsafe. See https://github.com/dapperlabs/flow-go/issues/6076 for details. 

This PR adds the `stakingEndView` as an argument to `resetEpoch`. 

⚠️ This will require an upgrade to `FlowEpoch`, and must be rolled out in tandem with the `flow-go` tooling changes described in https://github.com/dapperlabs/flow-go/issues/6076 (in-progress)


joshuahannan (approved):
Looks good to me! Would it make sense to try to include this in the same upgrade that we do for the [automated rewards calculation and payment?](https://github.com/onflow/flow-core-contracts/pull/267)

We can probably just merge this though and I'll integrate the changes into my PR

jordanschalm:
> Would it make sense to try to include this in the same upgrade that we do for the automated rewards calculation and payment?

👍  That makes sense. This should be ready for the next (mid-February) spork cycle.

PR #270, opened 2022-01-18
Update scripts for getting all stakes and all delegations

janezpodhostnik:
This change makes the scripts transactions/lockedTokens/delegator/get_delegator_info.cdc and transactions/lockedTokens/staker/get_staker_info.cdc use FlowStakingCollection.

These scripts are used by the flow-cli to get the stake and delegations of an account and its locked account.


joshuahannan (commented):
The staking collection code [already includes the locked account nodes](https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowStakingCollection.cdc#L907-L914) or delegators when getting node info, so it would probably make more sense to just update the CLI to use the staking collection scripts instead:
https://github.com/onflow/flow-core-contracts/blob/master/transactions/stakingCollection/scripts/get_all_node_info.cdc
https://github.com/onflow/flow-core-contracts/blob/master/transactions/stakingCollection/scripts/get_all_delegator_info.cdc

I'd also prefer that everything in the `lockedTokens` directory only include transactions that use the locked account setup and not the staking collection, which should ideally be all in the `stakingCollection` directory. 

Does that make sense? Let me know if I am missing anything.

janezpodhostnik:
That makes sense, I will just remove these files then and change the CLI to use the other scripts.

janezpodhostnik:
The cli can just use 

https://github.com/onflow/flow-core-contracts/blob/master/transactions/stakingCollection/scripts/get_all_node_info.cdc
https://github.com/onflow/flow-core-contracts/blob/master/transactions/stakingCollection/scripts/get_all_delegator_info.cdc

instead, so this PR is not needed.

PR #271, opened 2022-01-19
Contract Audit Vouchers Merge

tehranifar:
* Added contracts from https://github.com/tehranifar/flow-contract-audit-vouchers and migrated to `overflow` from `gwtf`.
* Added new `overflow` test suite
  * `lib/go/Makefile` was updated to invoke additional tests and ci from `lib/go/test/ContractAuditVouchers`.
  * `lib/go/test/ContractAuditVouchers` contains all the tests and specific config.
  * Contract and transactions follow repo rules.


joshuahannan (approved):
Looks good! I just had a few comments.

It would also be nice to see a few more comments in some of the transaction files, and in the test utilities for the ones that aren't totally obvious. 

Also, can you update the README to list your contract and include a brief description?

contracts/FlowContractAudits.cdc
```diff
@@ -0,0 +1,215 @@
...
+                if v.expiryBlockHeight != nil {
+                    if getCurrentBlock().height > v.expiryBlockHeight! {
+                        FlowContractAudits.vouchers.remove(key: key)
+                        emit VoucherRemoved(key: key, recurrent: v.recurrent, expiryBlockHeight: v.expiryBlockHeight)                
+                    }
+                }
+            }
+        }
+
+        // For testing
```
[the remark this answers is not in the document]
tehranifar: The FVM calls `FlowContractAudits.useVoucherForDeploy` on contract deployment. This call cannot be made in the emulator since the function is scoped as `access(contract)`. As a workaround the `Administrator` resource also provides `useVoucherForDeploy` that calls `FlowContractAudits.useVoucherForDeploy`.

contracts/FlowContractAudits.cdc
```diff
@@ -0,0 +1,215 @@
...
+            return true
+        }
+
+        return false
+    }
+
+    init() {
+        self.vouchers = {}
+
+        self.AdminStoragePath = /storage/contractAuditAdmin
```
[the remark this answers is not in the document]
tehranifar: Thank you for the feedback! I addressed the above (and comments + README) here https://github.com/onflow/flow-core-contracts/pull/271/commits/5a91e6a2c43bc13b22a44c79d4dc374d2aaf90e8
Let me know if you have additional feedback.

contracts/FlowContractAudits.cdc
```diff
@@ -0,0 +1,215 @@
...
+        init(address: Address?, recurrent: Bool, expiryBlockHeight: UInt64?, codeHash: String) {
+            self.address = address
+            self.recurrent = recurrent
+            self.expiryBlockHeight = expiryBlockHeight
+            self.codeHash = codeHash
+        }
+    }
+
+    // Returns all current vouchers
+    pub fun getAllVouchers(): {String: AuditVoucher} {
```
janezpodhostnik, line 52: Not sure what the use case of this method is? Is there a need for anyone to access the list of all vouchers?

It might be better to have a method where you can ask if there is a voucher for specific code and address.

On second thought... everything on chain is public, so I'm not sure if my concerns are valid. The only other concern is that if `self.vouchers` grows large this method will start failing...

contracts/FlowContractAudits.cdc
```diff
@@ -0,0 +1,215 @@
...
+            key = FlowContractAudits.generateVoucherKey(address: nil, codeHash: codeHash)
+            if !FlowContractAudits.vouchers.containsKey(key) {
+                return false
+            }
+        }
+
+        let v = FlowContractAudits.vouchers[key]!
+
+        // ensure contract code matches the voucher
+        if v.codeHash == codeHash  {
```
janezpodhostnik, line 180: It might be better to invert the if statement to reduce nesting.

contracts/FlowContractAudits.cdc
```diff
@@ -0,0 +1,215 @@
...
+                if getCurrentBlock().height > v.expiryBlockHeight! {
+                    FlowContractAudits.vouchers.remove(key: key)
+                    emit VoucherRemoved(key: key, recurrent: v.recurrent, expiryBlockHeight: v.expiryBlockHeight)
+                    return false
+                }
+            }
+
+            // remove the voucher if not recurrent
+            if !v.recurrent {
+                FlowContractAudits.vouchers.remove(key: key)
```
janezpodhostnik, line 194: It might be better to define a `removeVoucher` function that removes the voucher from the dictionary and emits the event. This way it might be harder to forget to emit the event if you are removing a voucher.

This is just a suggestion though. It is also fine as it is.

janezpodhostnik (commented):
Nice. I like the tests!

contracts/FlowContractAudits.cdc
tehranifar: There are multiple mechanisms in place to make sure the dictionary on this contract is not only increasing:
* Non-recurrent vouchers are removed on use
* Vouchers optionally expire based on block height
* Admin can call `cleanupExpiredVouchers` to remove all the vouchers that are currently expired and haven't been used. 
There is no high priority use case with the ability to fetch all vouchers (I have a sample script that does this), but I was thinking there is no way to fetch them all and check/debug for something in the future, hence this function. 
I agree with your feedback that if the "current" voucher list grows (like 10k? maybe we define a lot of recurrent templates) there might be performance issues with fetching them all. This argument holds for any data a contract might return. 
For this case with all the above, I prefer to keep the function but let me know if I'm missing something.

contracts/FlowContractAudits.cdc
tehranifar: Good idea! Addressed here https://github.com/onflow/flow-core-contracts/pull/271/commits/8d7a1e5bda44a3372b4fb2ae1542b2e6a99961ed

contracts/FlowContractAudits.cdc
tehranifar: I agree, this code piece has been repeated multiple times. Here's my take at refactoring: https://github.com/onflow/flow-core-contracts/pull/271/commits/8d7a1e5bda44a3372b4fb2ae1542b2e6a99961ed

janezpodhostnik (approved):

PR #273, opened 2022-01-26
[FlowContractAudits] Adding Hashed Vouchers

tehranifar:
`addVoucher` takes in the full contract code. Having the full contract code on the audit transaction is a feature because the developer can always compare the audited code with the deployment to fix any inconsistencies. This PR makes a second `addVoucherHashed` function that accepts the hashed source code instead. This can be desirable in two ways:
* Since contract audits happen before the actual deployment, information can be leaked about upcoming contracts when the audit voucher is generated. This enables the auditor to optionally make a "private" audit that doesn't disclose the source code before deployment.
* If a contract is too large (larger than 1.5 MB transaction size) it might be deployed to an account with intermediate storage usage. The auditor would have to use the same technique to generate the voucher which can be cumbersome. This PR addresses this case.


joshuahannan (approved):
Looks good!

PR #274, opened 2022-01-27
Fixes DKG Nil Keys Edge case

joshuahannan:
Closes https://github.com/dapperlabs/flow-go/issues/6048

Instead of returning nil and marking the DKG as incomplete if a nil key is found, the `dkgCompleted` method continues to check if there is another submission that has reached the success threshold.


tarakby (approved):
👌🏼

lib/go/test/flow_dkg_test.go
```diff
@@ -830,4 +829,34 @@ func TestDKGNil(t *testing.T) {
 		assert.Equal(t, cadence.NewBool(false), result)
 
 	})
+
+	dkgKey1 := fmt.Sprintf("%0192d", access)
+
+	t.Run("Should count as completed even if 50 of participants sent nil keys", func(t *testing.T) {
```
jordanschalm, line 835:
```suggestion
	// NOTE: one participant submitted a nil key vector in the previous test case
	t.Run("Should count as completed even if 50% of participants sent nil keys", func(t *testing.T) {
```
note about reliance on state from another test case.

jordanschalm (approved):
🎸

PR #279, opened 2022-03-03
Introduce new transaction fee calculation

janezpodhostnik:
Due to upcoming fee calculation changes, the fee deduction calculation needs to change. This relates to [FLIP: Variable transaction fees](https://github.com/onflow/flow/pull/660)

This PR:
- Leaves the old fee calculation in place, but add a depreciation comment to it. It will be removed after the spork that switches to the new fee calculation
- Adds the new calculation that has three parameters: `surgeFactor`, `executionEffortCost`, `inclusionEffortCost` packaged in  `FeeParameters` struct that is stored in storage.

## Breaking Change ⚠️ 

Fee deduction no longer emits a `FlowFees.TokensDeposited` but instead emits a `FlowFees.FeesDeducted`. `FlowFees.FeesDeducted` contains `inclusionEffort` and `executionEffort` info in addition to the final fee value.

This breaking change will come into effect after a spork switches the fee deduction from  `FlowServiceAccount.deductTransactionFee` to `FlowFees.deductTransactionFee`.


contracts/FlowFees.cdc
```diff
@@ -9,6 +9,12 @@ pub contract FlowFees {
     // Event that is emitted when tokens are withdrawn from the fee vault
     pub event TokensWithdrawn(amount: UFix64)
 
+    // Event that is emitted when fees are deducted
+    pub event FeesDeducted(amount: UFix64, inclusionEffort: UFix64, executionEffort: UFix64)
```
joshuahannan, line 13: Would it make sense to just add these parameters to the end of the TokensDeposited Event so there isn't the breaking change?

contracts/FlowFees.cdc
```diff
@@ -33,6 +39,80 @@ pub contract FlowFees {
...
             return <-vault
         }
+
+        // Allows the administrator to change all the fee parameters at once
+        pub fun setFeeParameters(surgeFactor: UFix64, inclusionEffortCost: UFix64, executionEffortCost: UFix64) {
+            let newParameters = FeeParameters(surgeFactor: surgeFactor, inclusionEffortCost: inclusionEffortCost, executionEffortCost: executionEffortCost)
+            FlowFees.setFeeParameters(newParameters)
+        }
+
+        // Allows the administrator to change the fee surge factor
```
joshuahannan, line 49:
```suggestion
        /// Allows the administrator to change the fee surge factor
```

contracts/FlowFees.cdc
```diff
@@ -33,6 +39,80 @@ pub contract FlowFees {
...
+        pub fun setFeeSurgeFactor(surgeFactor: UFix64) {
+            let oldParameters = FlowFees.getFeeParameters()
+            let newParameters = FeeParameters(surgeFactor: surgeFactor, inclusionEffortCost: oldParameters.inclusionEffortCost, executionEffortCost: oldParameters.executionEffortCost)
+            FlowFees.setFeeParameters(newParameters)
+        }
+    }
+
+    // A struct holding the fee parameters needed to calculate the fees
```
joshuahannan, line 57:
```suggestion
    /// A struct holding the fee parameters needed to calculate the fees
```

contracts/FlowFees.cdc
```diff
@@ -33,6 +39,80 @@ pub contract FlowFees {
...
+        }
+
+        let tokenVault = acct.borrow<&FlowToken.Vault>(from: /storage/flowTokenVault)
+            ?? panic("Unable to borrow reference to the default token vault")
+
+        
+        if feeAmount > tokenVault.balance {
+            // In the future this code path will never be reached, 
+            // as payers that are under account minimum balance will not have their transactions included in a collection
+            feeAmount = tokenVault.balance
```
joshuahannan, line 88: Won't this cause the transaction to fail because the vault will be below the minimum balance? Why not just fail with an error message here to be more clear?

contracts/FlowFees.cdc
joshuahannan: I'm also tempted to add an `Address` parameter here

contracts/FlowFees.cdc
```diff
@@ -33,6 +39,80 @@ pub contract FlowFees {
...
+        emit FeesDeducted(amount: feeAmount, inclusionEffort: inclusionEffort, executionEffort: executionEffort)
+    }
+
+    pub fun getFeeParameters(): FeeParameters {
+        return self.account.copy<FeeParameters>(from: /storage/FeeParameters)!
+    }
+
+    access(self) fun setFeeParameters(_ feeParameters: FeeParameters) {
+        // empty storage before writing new FeeParameters to it
+        self.account.load<FeeParameters>(from: /storage/FeeParameters)
```
joshuahannan, line 104: Can we also make the storage path a bit more descriptive? 
`/storage/FlowTxFeeParameters` or something like that?

joshuahannan (commented):
Good start so far!

contracts/FlowFees.cdc
janezpodhostnik: > Would it make sense to just add these parameters to the end of the TokensDeposited Event so there isn't the breaking change?
I was thinking about that too, but that event is also used for account creation, which does not have the extra parameters. We could emit both events, but that increases the event count for all transactions. Another option would perhaps be to create a separate event for account creation instead.

> I'm also tempted to add an Address parameter here
That address will be the address of the payer which you have access to if you have the transaction data, so I'm not sure that adding the address here makes things better for users.

contracts/FlowFees.cdc
janezpodhostnik: If the case is that the payers funds are to low to pay for the transaction, what we would ultimately want is for that transaction to fail much sooner (so it never even gets executed), but that is not in place yet.

The current logic is that if a payer sends a transaction that puts the payer below the minimum account balance, all the payers transactions after that will fail because the account is minimum account balance (unless the payer adds more FLOW to their vault). This means their account will have less and less FLOW until they hit 0.0.

This code is here because of the following case:

Lets say:
- payer P has 99 FLOW
- Minimum account balance is 100 FLOW
- and transaction fees are 5 FLOW per transaction

- after P sends 1 transaction that doesn't increase P's balance, P has 94 FLOW (the transaction failed)
- after P sends 19 transaction that doesn't increase P's balance, P has 4 FLOW (the transactions all failed)
- P sends 1 more transaction that doesn't increase P's balance. The choice is to say P's balance is less than the transaction fees, so don't deduct transaction fees (tx will still fail) **or** deduct the remaining 4 FLOW from P, so P is at 0

This code is the code for the second option.

contracts/FlowFees.cdc
joshuahannan: > That address will be the address of the payer which you have access to if you have the transaction data, so I'm not sure that adding the address here makes things better for users.
Good point. I didn't think about that

> We could emit both events, but that increases the event count for all transactions. Another option would perhaps be to create a separate event for account creation instead.
It really depends on how many people are monitoring the event and if it would affect them. my guess is that it isn't very many, so adding this new one is probably fine

joshuahannan (approved):

contracts/FlowFees.cdc
janezpodhostnik: This is not in scope here, but that is still a problem that needs to be properly solved.

The current idea as I know it is that the payer needs to have enough funds on their account, to cover both inclusion fees (known without running the transaction) and the max execution fees (gas limit) set by the transaction. If there are not enough funds the transaction would never be included in a collection.

contracts/FlowFees.cdc
janezpodhostnik: So you are ok with only triggering `FeesDeducted`?

joshuahannan: yeah, I'm ok with that

contracts/FlowFees.cdc
bluesign: thanks @janezpodhostnik, after reading FVM code, I just understood what you did here. 

I think there is still an edge case here (allowing 0 transaction fees)

- Alice has 100 FLOW on FlowTokenVault 
- Alice links <&FlowToken.Vault{FungibleToken.Balance}>(/public/flowTokenBalance) to this vault 
- Alice links <&FlowToken.Vault>(from: /storage/flowTokenVault) to an empty vault 
- Alice sends a transaction ( passes balance checks probably ) 
- Transaction fee deducted as 0 and succeeds
- Storage checks for updated Accounts succeeds according to : https://github.com/onflow/flow-core-contracts/blob/26afc84a19039ae7e6628b4bc4b4f53083be4abc/contracts/FlowStorageFees.cdc#L65

This is more in scope of FlowStorageFees I guess though.

PR #281, opened 2022-03-04
crypto setup improvement

tarakby:
- bump flow-go/crypto to v0.24.3
- use the flow-go/crypto script to setup the crypto lib. From now on, there is no need to bump the crypto version in different files, `go.mod` only is enough.


janezpodhostnik (approved):
Nice, automation!

PR #287, opened 2022-04-21
Update cadence version to secure-cadence-m4

janezpodhostnik:


joshuahannan (approved):
Looks good to me! Thanks for making these changes.

@JeffreyDoyle @psiemens @Wolog2021 Some of the transactions that were updated here are transactions that are used by Flow Port/ledger:

`transactions/epoch/node/register_node.cdc` is used by node operators to register their nodes
`transactions/lockedTokens/admin/admin_create_shared_accounts.cdc` I believe is used when Flow Port creates a new ledger account for a new user.

So we'll need to make sure that we have these transaction updated for ledger and Flow port when we do the secure Cadence spork.

There are also a few other transactions that blocto and other custody providers are using to create accounts that will need to get updated as well. They all involve locked tokens though, so IMO we should just try to get everyone to phase them out for the regular account creation transactions instead since we don't really need to deliver locked tokens anymore. Thoughts?

PR #290, opened 2022-06-14
[WIP] Update DST for QC signature verification

tarakby:
- update the domain separation tag in QC cluster contract to match the tag used for signature generation in flow-go repo (updated to comply to https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-14).

### DIFF
--- a/contracts/epochs/FlowClusterQC.cdc
+++ b/contracts/epochs/FlowClusterQC.cdc
@@ -299,7 +299,7 @@ pub contract FlowClusterQC {
             let isValid = publicKey.verify(
                 signature: voteSignature.decodeHex(),
                 signedData: voteMessage.decodeHex(),
-                domainSeparationTag: "FLOW-V0.0_Collector-Vote",
+                domainSeparationTag: "FLOW-Collector_Vote-V00-CS00-with-",
                 hashAlgorithm: HashAlgorithm.KMAC128_BLS_BLS12_381
             )
 


lib/go/test/go.mod
```diff
@@ -8,7 +8,6 @@ require (
 	github.com/onflow/flow-core-contracts/lib/go/templates v0.11.2-0.20220422202806-92ad02a996cc
 	github.com/onflow/flow-emulator v0.31.2-0.20220513151845-ef7513cb1cd0
 	github.com/onflow/flow-ft/lib/go/templates v0.2.0
-	github.com/onflow/flow-go v0.25.13-0.20220513151142-7858f76e703b
```
janezpodhostnik, line 11: was this the only reference to flow-go?

janezpodhostnik (approved):
LGTM.

lib/go/test/go.mod
tarakby: - `flow-go` was imported to use domain separation tags from the protocol. Once the protocol is updated, I'm going to import the tags again, though this needs to be done over separate steps.
`github.com/onflow/flow-go/crypto` is also imported below but it that's a separate module.

contracts/epochs/FlowClusterQC.cdc
```diff
@@ -299,7 +299,7 @@ pub contract FlowClusterQC {
             let isValid = publicKey.verify(
                 signature: voteSignature.decodeHex(),
                 signedData: voteMessage.decodeHex(),
-                domainSeparationTag: "FLOW-V0.0_Collector-Vote",
+                domainSeparationTag: "FLOW-Collector_Vote-V00-CS00-with-",
```
satyamakgec, line 302: Is there a way to verify that the provided tag is correct? Just want to double confirm.

satyamakgec (approved):
LGTM - Just a minor comment left

contracts/epochs/FlowClusterQC.cdc
tarakby: Good question. The tests in the repo are supposed to verify the tag is correct: 
 - the test generates signatures using a tag constant imported from `flow-go` repo (it means the tag used by the protocol).
 - the test submits the signatures to get verified by the contract. 
 - if the there is a tag mismatch, the signature verification fails. 
 
 The constant tag is still not read from `flow-go` yet though, it is hardcoded for now in the tests. Once I update `flow-go` with the new constants, I'll update the tests in this repo again (a circular dependency is the reason behind all these steps)

PR #291, opened 2022-06-14
Adding transaction for setting execution memory limit

j1010001:
Use this Transaction for setting/changing execution memory limit.


janezpodhostnik:
I added this to https://github.com/onflow/flow-core-contracts/pull/292 which also includes a test for it, and the script to check it.

PR #292, opened 2022-06-15
Execution Memory Limit

janezpodhostnik:
- New transaction to set the memory limit
- Service Account contract change to add a way to get the memory limit
- Script template to get the memory limit


contracts/FlowServiceAccount.cdc
```diff
@@ -140,6 +140,12 @@ pub contract FlowServiceAccount {
             ?? panic("execution memory weights not set yet")
     }
 
+        // Gets Execution Memory Limit from the service account's storage 
+    pub fun getExecutionMemoryLimit(): UInt64 {
+        return self.account.copy<UInt64>(from: /storage/executionMemoryLimit) 
+            ?? panic("execution memory limit not set yet")
+    }
```
joshuahannan, line 147: Should this return `nil` instead of panicing?

joshuahannan (approved):

contracts/FlowServiceAccount.cdc
janezpodhostnik: It could.

The reason I chose panicking, is that once we set it on the network it is never expected to be "unset" so that it will be nil again.

Do you think it would be better to return nil?

joshuahannan: probably doesn't matter like you said. This is probably fine

PR #295, opened 2022-07-13
Add calculateAccountsCapacity to storage fees

janezpodhostnik:
... for getting the storage capacity of multiple account with one call.

This is mainly intended to be used by the FVM as a performance improvement for the storage capacity check that happens at the end of every transaction.

This is related to https://github.com/dapperlabs/flow-go/issues/6308.


bluesign:
Maybe we can fix calculateAccountCapacity using balance interface at the same time?

joshuahannan (approved):

bluesign:
@janezpodhostnik calculateAccountCapacity using Balance interface ( shared on different path ) , so technically you can have 2 different vaults in Balance and Vault locations. 

This causes issue that transaction fees can be avoided. ( But Vault with balance to balance path, put empty vault on main vault path, storage fees satisfy and doesn't rollback transaction, but transaction fees cannot get the fee ( but accept the 0 value ) ) 

I commented this long ago in detail in another PR.

PR #300, opened 2022-07-21
Add destroy conditional to FlowToken smart contract

joshuahannan:
Adds a conditional in the deposit block of `FlowToken` so that the assignment only happens when it needs to.
This assignment happens every time a FlowToken transfer happens, but the value is basically always zero, so it is a waste of computation.


alilloig (approved):

janezpodhostnik (approved):

janezpodhostnik:
I have been thinking about this, and this should noticeably speed up transaction execution for simple transaction, since FlowToken account will no longer be touched, and thus won't need a storage capacity vs storage used check.

PR #302, opened 2022-07-27
update dependencies and fix breaking changes

joshuahannan:
Updated to:
```
	github.com/onflow/cadence v0.24.2-0.20220627202951-5a06fec82b4a
	github.com/onflow/flow-emulator v0.33.4
	github.com/onflow/flow-go-sdk v0.26.5-0.20220629191626-900f9f91bffc
```
and updated overflow. I commented out some of the event assertions in the audit vouchers tests because there were some issues with overflow. This hopefully shouldn't be an issue because we aren't using that contract any more.

This caused most of the tests to break because arrays in Cadence are typed now, so I had to go through all the tests to fix those

Some of the transactions here will need to be updated for mainnet use


nvdtf (approved):
Thanks for updating the tests! Yes we should deprecate that contract from this repo.

PR #303, opened 2022-07-29
Use correct rewards value when paying rewards and adds new epoch event

joshuahannan:
The `payRewards` transaction was using the new epoch's total rewards value when minting tokens for rewards instead of the previous epoch's amount, which was causing it to mint more tokens than necessary and burn the excess tokens at the end. This makes a change to the function to get a parameter that can indicate the amount of tokens to mint, which is always set to the previous epoch's payout in the `FlowEpoch` contract.

Also adds the `EpochTransition` event to the epoch contract, to show when a new epoch starts. Definitely open to discussion what parameters to include in this event and what it should be named.


contracts/epochs/FlowEpoch.cdc
```diff
@@ -364,11 +367,14 @@ pub contract FlowEpoch {
                     let currentBlock = getCurrentBlock()
                     let currentEpochMetadata = FlowEpoch.getEpochMetadata(FlowEpoch.currentEpochCounter)!
                     if currentBlock.view >= currentEpochMetadata.endView {
+                        let totalRewards = FlowIDTableStaking.epochTokenPayout
```
jordanschalm, line 370: This diff addresses the functional problem, I wonder if we could improve the structure at the same time. To me the underlying problem is that we are relying on a contract-global variable (`epochTokenPayout`) in two different functions. We are also handling `rewardBreakdown` and `totalRewards` separately (one is passed as an argument, one is stored in a struct) when it is critical that they are mutually consistent. 

Could we modify the return value of `calculateRewards` to return a structure that includes the total reward payout? That way we could read the global variable `epochTokenPayout` in only one place in the flow, and set/store the `rewardBreakdown` and `totalRewards` in the same place. To me, this would less prone to the kind of problem this PR addresses. Let me know what you think.

contracts/epochs/FlowEpoch.cdc
```diff
@@ -57,6 +57,9 @@ pub contract FlowEpoch {
         pub case Access
     }
 
+    /// Event that indicates that a new epoch has started
+    pub event EpochTransition(previousEpoch: EpochMetadata, newEpoch: EpochMetadata)
```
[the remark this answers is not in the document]
jordanschalm: Copying [a note from Slack](https://axiomzen.slack.com/archives/CEEGK3HGC/p1659470900579199?thread_ts=1659058387.362459&cid=CEEGK3HGC) regarding why we are adding this new event:

> Is the suggestion to use the NewEpoch event as a signal to the protocol to transition to the new epoch?
If so, we can't safely do that. The reason we use a preset view deadline for the switchover is to satisfy HotStuff safety requirements.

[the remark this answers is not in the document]
joshuahannan: I'll just remove this event then

contracts/epochs/FlowEpoch.cdc
joshuahannan: yep, I'll modify that to use it. Unfortunately, we can't change the epoch metadata struct so we'll still have to use the rewards breakdown array for the event and metadata storage though

transactions/epoch/admin/pay_rewards.cdc
```diff
@@ -6,6 +6,8 @@ transaction {
         let heartbeat = signer.borrow<&FlowEpoch.Heartbeat>(from: FlowEpoch.heartbeatStoragePath)
             ?? panic("Could not borrow heartbeat from storage path")
 
-        heartbeat.payRewards()
+        let previousEpochMetadata = FlowEpoch.getEpochMetadata(FlowEpoch.currentEpochCounter - (1 as UInt64))!
```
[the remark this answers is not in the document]
joshuahannan: Sounds like you understand how it works now, right? We don't actually use this transaction right now, but if we did, we would run it after an epoch ends, not before. This transaction is kind of an in-between step between how we do it now (manually using only the staking contract) and automated rewards (using the epoch contract)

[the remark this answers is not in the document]
joshuahannan: Do you have any other problems with this transaction or can it remain as-is?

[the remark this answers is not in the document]
jordanschalm: > We don't actually use this transaction right now, but if we did, we would run it after an epoch ends, not before

Thanks, yes I understand how we would change the manual reward payment transaction timing now.

From my perspective, the fact that sending the same manual `PayRewards` transaction at slightly different times can result in different outcomes, depending on the current epoch state, is prone to errors. This property is a big part of the recent manual reward payout issue we've been discussing in Slack. 

The proposed `epoch/admin/pay_rewards.cdc` transaction in this PR  maintains the property of timing dependency (just different timing). However:
* if we accidentally send the transaction early (before the epoch switchover), it will fail rather than pay the wrong rewards
* after the epoch switchover, we have 1 week to get the transaction in, rather than only a few hours with the current manual reward payout

So overall I think this in-between manual payout transaction is addressing most of the drawbacks of the one we currently use. Happy to keep it as-is 👍

contracts/epochs/FlowEpoch.cdc
```diff
@@ -509,14 +504,14 @@ pub contract FlowEpoch {
...
     }
 
     /// Pays rewards to the nodes and delegators of the previous epoch
-    access(account) fun payRewards(_ totalRewards: UFix64) {
+    access(account) fun payRewards(_ rewardsSummary: FlowIDTableStaking.EpochRewardsSummary) {
         if let previousEpochMetadata = self.getEpochMetadata(self.currentEpochCounter - (1 as UInt64)) {
             if !previousEpochMetadata.rewardsPaid {
-                let rewardsBreakdownArray = previousEpochMetadata.rewardAmounts
-                self.borrowStakingAdmin().payRewards(rewardsBreakdownArray, totalRewards: totalRewards)
+                self.borrowStakingAdmin().payRewards(rewardsSummary)
```
jordanschalm, line 514: > @jordanschalm I just pushed some new changes, mostly making that new struct that also includes the total rewards amount: [4d78105](https://github.com/onflow/flow-core-contracts/pull/303/commits/4d781056f74d9947d6ad7446c4ca70bff6eedead)
> It is still a little bit awkward, but let me know if you think it is better.

Looks good, for me at least this structure is easier to follow 🙏.

My only other concern is that here, in `payRewards` the `EpochRewardsSummary` is passed in as an argument and this function assumes that it is for the previous epoch. It would be great to avoid the implicit coupling between the reward summary and epoch in the scope of `payRewards`.

Suggestion 1: Is it possible at this point to read both `rewardAmounts` and `totalRewards` from `previousEpochMetadata` (like we do in the [admin transaction](https://github.com/onflow/flow-core-contracts/pull/303/commits/4d781056f74d9947d6ad7446c4ca70bff6eedead#diff-dc2590b4ab366fdfdd96171ab0c059cb6c6e008ed5f9215f7ebe6d46566ec9d5R11))
```
if !previousEpochMetadata.rewardsPaid {
  let rewardsSummary = FlowIDTableStaking.EpochRewardsSummary(totalRewards: previousEpochMetadata.totalRewards, breakdown: previousEpochMetadata.rewardAmounts)
  self.borrowStakingAdmin().payRewards(rewardsSummary)
}
```
* We could fully remove the argument to `payRewards` and rename it `payRewardsForPreviousEpoch`

Suggestion 2: Alternatively, we could include an epoch counter in `EpochRewardsSummary` so that `payRewards` can sanity check that the `previousEpochMetadata` read from storage is consistent with the input argument. I don't think this is simple to do though, since `EpochRewardsSummary` needs to be instantiated in `FlowIDTableStaking` as well, which doesn't know about the epoch counter.

joshuahannan: I think I like what you're saying in Suggestion 1. It seemed a bit awkward to me also. I changed it in my latest commit to do that

contracts/epochs/FlowEpoch.cdc
```diff
@@ -367,8 +367,9 @@ pub contract FlowEpoch {
                         self.calculateAndSetRewards()
                         self.endEpoch()
                         if FlowEpoch.automaticRewardsEnabled() {
-                            self.payRewards()
+                            self.payRewardsForPreviousEpoch()
                         }
+                        let newEpochMetadata = FlowEpoch.getEpochMetadata(FlowEpoch.currentEpochCounter)!
```
jordanschalm, line 372:
```suggestion
```
Seems like an unused variable?

jordanschalm (approved):

contracts/epochs/FlowEpoch.cdc
joshuahannan: yes, good catch. I removed it

satyamakgec (approved):
Syntactically and programmatically, I reviewed these changes and basis on that LGTM. 

Logically I need more knowledge of the core smart contracts. However, changes seem straightforward to me.

PR #305, opened 2022-08-03
update overflow tests

bjartek:
- update to newest overflow
- use the new api


bjartek:
Not really sure what is going on with my commits here, just trying to clean things up.

PR #309, opened 2022-09-01
Slot Selection - WIP

Wolog2021:
Initial work in process commit for review and comment.


joshuahannan:
@Wolog2021 Make sure you pull the latest changes from master. There was an update to the staking, epoch contracts, and tests that you'll need to have

contracts/FlowIDTableStaking.cdc
```diff
@@ -857,6 +857,70 @@ pub contract FlowIDTableStaking {
...
 
+        /// Enforces slot (count) limits for each node type (role). If more nodes
+        /// of a given type are registered than the limit, they are removed
+        /// randomly until the limits are met. When a node is removed, it's
+        /// committed tokens are transfered to it's unstaked bucket.
+        pub fun slotSelect() {
+            let allNodeIDs = FlowIDTableStaking.getNodeIDs()
+
+            let limits = [2, 2, 2, 2, 2]
+            var counts = [0, 0, 0, 0, 0]
```
joshuahannan, line 868: We would prefer the limits field to be stored as fields in the contract instead of hard-coded if possible. See how we added these "fields" in upgrades previous times for `stakingEnabled`.
https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowIDTableStaking.cdc#L1286

https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowIDTableStaking.cdc#L798-L802

joshuahannan: Also, wouldn't it be better to make these as dictionaries indexed by node role so we don't have the awkward off-by-one index?

contracts/FlowIDTableStaking.cdc
```diff
@@ -857,6 +857,70 @@ pub contract FlowIDTableStaking {
...
+            for nodeID in allNodeIDs {
+                let nodeRecord = FlowIDTableStaking.borrowNodeRecord(nodeID)
+                if nodeRecord.initialWeight == 100 {
+                    counts[nodeRecord.role - 1] = counts[nodeRecord.role - 1] + 1
+                }
+            }
+
+            for index, element in counts {
```
joshuahannan, line 876:
```suggestion
            for index, nodeCount in counts {
```

contracts/FlowIDTableStaking.cdc
```diff
@@ -857,6 +857,70 @@ pub contract FlowIDTableStaking {
...
+                while counts[index] > limits[index] {
```
joshuahannan, line 877: You get the element at the index in each iteration of the loop, so there is no need to access the array again. you can just do `while element > limits[index]`

contracts/FlowIDTableStaking.cdc
```diff
@@ -857,6 +857,70 @@ pub contract FlowIDTableStaking {
...
+                    let selection = unsafeRandom() % UInt64(counts[index])
```
joshuahannan, line 878: Same here: `element` instead of `counts[index]`

joshuahannan (commented):
Great start! I just had a few comments. Also, some of the variable names are a bit unclear. Can you make them more descriptive? Thank you!

contracts/FlowIDTableStaking.cdc
Wolog2021: Firstly, thanks very much for the review! I'll change the data structure to a dictionary stored as a contract field. Would you then initialize it in the "init" function with default values as is done with stakingEnabled?

joshuahannan: yes, but you'll also need to make a method in the admin resource that initializes it in the account, because we'll need to run it to set it up after the contract update and if we ever want to change it

Wolog2021: I'm going to push changes shortly that will change the names to "slotLimits" and "nodeCounts" and change the arrays to dictionaries. The slotLimits will now be stored as a contract field as requested. I added a method to set them and I now use it in my test. I also assign default values.

contracts/FlowIDTableStaking.cdc
Wolog2021: As this loop will now be iterating through a dictionary, the "element" variable is no longer present. If it is included, the code still parses but the result is that the keys start at zero regardless of the fact that I coded them to start at one.

contracts/FlowIDTableStaking.cdc
Wolog2021: As mentioned above, the "element" parameter is going away in the change from arrays to dictionaries. But I have a question. As I understand it, "element" would have been a constant within the loop body which means I can't use it to update the values in "counts". Obviously I need to be updating the counts as we're removing nodes. It's not a critical question anymore, just confirming my understanding.

Wolog2021: Oh and as I'm updating counts I also need to be consuming the latest updated values. If "element" is constant, it wouldn't work for that either.

contracts/FlowIDTableStaking.cdc
Wolog2021: This comment references the same line as the prior one. Was it meant to refer to a different line?

contracts/FlowIDTableStaking.cdc
joshuahannan: yeah, the dictionary would change how you do this, so you can ignore this comment

contracts/FlowIDTableStaking.cdc
joshuahannan: I'll take another look

contracts/FlowIDTableStaking.cdc
```diff
@@ -870,6 +870,78 @@ pub contract FlowIDTableStaking {
             }
         }
 
+        /// Set slot (count) limits for each node type (role).
+        pub fun setSlotLimits(slotLimits: {UInt8: UInt8}) {
+            FlowIDTableStaking.account.load<{UInt8: UInt8}>(from: /storage/slotLimits)
+            FlowIDTableStaking.account.save(slotLimits, to: /storage/slotLimits)
```
joshuahannan, line 876:
```suggestion
            FlowIDTableStaking.account.save(slotLimits, to: /storage/flowStakingSlotLimits)
```
Let's try to give this a more descriptive name also

Wolog2021: Done.

contracts/FlowIDTableStaking.cdc
```diff
@@ -870,6 +870,78 @@ pub contract FlowIDTableStaking {
...
+        }
+
+        /// Enforces slot (count) limits for each node type (role). If more nodes
+        /// of a given type are registered than the limit, they are removed
+        /// randomly until the limits are met. When a node is removed, it's
+        /// committed tokens are transfered to it's unstaked bucket.
+        pub fun slotSelect() {
+            let allNodeIDs = FlowIDTableStaking.getNodeIDs()
+
+            let slotLimits = FlowIDTableStaking.account.load<{UInt8: UInt8}>(from: /storage/slotLimits)
```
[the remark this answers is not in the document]
Wolog2021: Done.

contracts/FlowIDTableStaking.cdc
```diff
@@ -870,6 +870,78 @@ pub contract FlowIDTableStaking {
...
+            var nodeCounts: {UInt8: UInt8} = {1: 0, 2: 0, 3: 0, 4: 0, 5: 0}
+            for nodeID in allNodeIDs {
+                let nodeRecord = FlowIDTableStaking.borrowNodeRecord(nodeID)
+                if nodeRecord.initialWeight == 100 {
+                    nodeCounts[nodeRecord.role] = nodeCounts[nodeRecord.role]! + 1
+                }
+            }
+
+            for key in nodeCounts.keys {
+                while nodeCounts[UInt8(key)]! > slotLimits![UInt8(key)]! {
```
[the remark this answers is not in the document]
Wolog2021: Done.

contracts/FlowIDTableStaking.cdc
```diff
@@ -870,6 +870,78 @@ pub contract FlowIDTableStaking {
...
+            // count the nodes of each type
+            var nodeCounts: {UInt8: UInt8} = {1: 0, 2: 0, 3: 0, 4: 0, 5: 0}
+            for nodeID in allNodeIDs {
+                let nodeRecord = FlowIDTableStaking.borrowNodeRecord(nodeID)
+                if nodeRecord.initialWeight == 100 {
+                    nodeCounts[nodeRecord.role] = nodeCounts[nodeRecord.role]! + 1
+                }
+            }
+
+            for key in nodeCounts.keys {
```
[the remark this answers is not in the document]
Wolog2021: It looks like casting isn't necessary anymore. I think it was a holdover from the prior array implementation. I've removed it.

lib/go/test/flow_idtable_staking_test.go
```diff
@@ -2413,6 +2422,119 @@ func TestIDTableStaking(t *testing.T) {
...
+			stakingKeys,
+			networkingKeys,
+			ids)
+
+		//check resulting array length
+		result := executeScriptAndCheck(t, b, templates.GenerateReturnProposedTableScript(env), nil)
+		idArray := result.(cadence.Array).Values
+		assert.Len(t, idArray, nodeCount)
+
+		//remove unapproved nodes to match final state
```
jordanschalm, line 2489: What does "match final state" mean here?

contracts/FlowIDTableStaking.cdc
```diff
@@ -1429,6 +1502,9 @@ pub contract FlowIDTableStaking {
     init(_ epochTokenPayout: UFix64, _ rewardCut: UFix64) {
         self.account.save(true, to: /storage/stakingEnabled)
 
+        let slotLimits: {UInt8: UInt8} = {1: 3, 2: 3, 3: 3, 4: 3, 5: 3}
```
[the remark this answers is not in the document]
jordanschalm: We should also select different defaults before merging this. Assuming we intend to invoke slot selection alongside the auto-recovery process, this will break epoch transitions on many transient networks we use currently (Localnet, Benchnet)

contracts/FlowIDTableStaking.cdc
```diff
@@ -870,6 +870,79 @@ pub contract FlowIDTableStaking {
...
+                                        // move their committed tokens back to their unstaked tokens
+                                        delRecord.tokensUnstaked.deposit(from: <-delRecord.tokensCommitted.withdraw(amount: delRecord.tokensCommitted.balance))
+                                    }
+
+                                    // Request to unstake all tokens
+                                    delRecord.tokensRequestedToUnstake = delRecord.tokensStaked.balance
+                                }
+
+                                // Clear initial weight because the node is not staked any more
+                                nodeRecord.initialWeight = 0
```
jordanschalm, line 932: This (lines 906-932) looks similar to [a part of `removeUnapprovedNodes`](https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowIDTableStaking.cdc#L838-L864). If possible, it would be great to encapsulate the shared logic into a shared helper function.

contracts/FlowIDTableStaking.cdc
```diff
@@ -870,6 +870,79 @@ pub contract FlowIDTableStaking {
...
+            var nodeCounts: {UInt8: UInt8} = {1: 0, 2: 0, 3: 0, 4: 0, 5: 0}
+            for nodeID in allNodeIDs {
+                let nodeRecord = FlowIDTableStaking.borrowNodeRecord(nodeID)
+                if nodeRecord.initialWeight == 100 {
+                    nodeCounts[nodeRecord.role] = nodeCounts[nodeRecord.role]! + 1
+                }
+            }
+
+            for key in nodeCounts.keys {
+                while nodeCounts[key]! > slotLimits[key]! {
```
jordanschalm, line 899: We will need a way to omit some node roles from slot selection entirely, because we cannot immediately allow public access to all node roles.

contracts/FlowIDTableStaking.cdc
```diff
@@ -870,6 +870,79 @@ pub contract FlowIDTableStaking {
...
+        pub fun setSlotLimits(slotLimits: {UInt8: UInt8}) {
+            FlowIDTableStaking.account.load<{UInt8: UInt8}>(from: /storage/flowStakingSlotLimits)
+            FlowIDTableStaking.account.save(slotLimits, to: /storage/flowStakingSlotLimits)
+        }
+
+        /// Enforces slot (count) limits for each node type (role). If more nodes
+        /// of a given type are registered than the limit, they are removed
+        /// randomly until the limits are met. When a node is removed, it's
+        /// committed tokens are transfered to it's unstaked bucket.
+        pub fun slotSelect() {
```
jordanschalm, line 883: When do we expect `slotSelect` to be called in relation to the rest of the staking/epoch flow?

Currently slot selection will remove excess nodes which have sufficient stake and which have `initialWeight == 100`. However, currently all nodes must be a member of the approved list before they are provisionally registered for an epoch (have `initialWeight=100` set). It is unclear to me how we intend for unauthorized Access Nodes to be provisionally registered for an epoch (the step before slot selection).

contracts/FlowIDTableStaking.cdc
```diff
@@ -870,6 +870,79 @@ pub contract FlowIDTableStaking {
...
+                                    delRecord.tokensRequestedToUnstake = delRecord.tokensStaked.balance
+                                }
+
+                                // Clear initial weight because the node is not staked any more
+                                nodeRecord.initialWeight = 0
+
+                                //Decrement the count for this role
+                                nodeCounts[key] = nodeCounts[key]! - 1
+
+                                break
```
jordanschalm, line 937: This is potentially quite expensive - in the worst case we will iterate `nodeIDs` once for each excess provisionally registered node. 
* ⚠️  I believe anyone can register a node, even with 0 stake - these nodes will stay in `nodeIDs` forever ([reference](https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowIDTableStaking.cdc#L1190-L1199))
* Similarly, anyone can provisionally register a node by providing the stake minimum.

I would suggest taking a different approach to this iteration once this moves past the WIP stage.

contracts/FlowIDTableStaking.cdc
```diff
@@ -870,6 +870,79 @@ pub contract FlowIDTableStaking {
...
+            let allNodeIDs = FlowIDTableStaking.getNodeIDs()
+
+            let slotLimits = FlowIDTableStaking.account.load<{UInt8: UInt8}>(from: /storage/flowStakingSlotLimits)
+                ?? panic("Could not load the slot limits object from storage!")
+
+            // count the nodes of each type
+            var nodeCounts: {UInt8: UInt8} = {1: 0, 2: 0, 3: 0, 4: 0, 5: 0}
+            for nodeID in allNodeIDs {
+                let nodeRecord = FlowIDTableStaking.borrowNodeRecord(nodeID)
+                if nodeRecord.initialWeight == 100 {
```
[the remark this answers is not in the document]
joshuahannan: We can't create new contract fields any more, but we could store the constants in storage and retrieve them with a getter function like we do with `getStakingEnabled()` if we want. Do you think we should do that?

contracts/FlowIDTableStaking.cdc
pgebheim: @jordanschalm Yes, you're pointing to an issue.

@Wolog2021 How I expected this to work is as follows:

- During `removedUnapprovedNodes` the contract should track nodes which are transitioning from 0 weight to non-zero weight. This happens on [Line 868](https://github.com/onflow/flow-core-contracts/blob/ea44a5dd9838e44c92119ce470eb724c22772f2e/contracts/FlowIDTableStaking.cdc#L868).

- Then, randomly select from that list of nodes which have transitioned into being available until nodeCounts[type] >= slotLimits[type[.

I believe this whole solution needs to be modified to take this list of possible nodes, and then be called from the end of removeUnapprovedNodes.

contracts/FlowIDTableStaking.cdc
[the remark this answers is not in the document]
pgebheim: Constants for node roles would be nice but I believe that is fine to punt on since a future staking auction system could handle it.

As for defaults, I think they will be set during the initial contract upgrade and that should be fine as it will be documented during the upgrade process.

contracts/FlowIDTableStaking.cdc
pgebheim: I think it may be possible to actually just merge these two functions together so that all of the remove logic happens after slot selection.

pgebheim (changes requested):

contracts/FlowIDTableStaking.cdc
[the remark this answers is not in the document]
jordanschalm: Yeah, that's what I remember - if we can't add "real" constants (`let x = ...`) feel free to disregard this suggestion

contracts/FlowIDTableStaking.cdc
Wolog2021: We can certainly call the function from the end of removeUnapprovedNodes. The reason I didn't do that is that it's doing something fundamentally different i.e. excess otherwise approved nodes are being removed randomly and not because they are unapproved. I didn't want the function name to not be fully descriptive of the operation. Is that a concern here?

contracts/FlowIDTableStaking.cdc
Wolog2021: As noted above if we merge the two functions I think we should consider changing the name. Are you envisioning that we would pre-select the nodes that would be removed by removeUnapprovedNodes and then pre-select those that would be removed by slot selection and then remove them all in one final loop?

contracts/FlowIDTableStaking.cdc
Wolog2021: I'm refactoring this.

contracts/FlowIDTableStaking.cdc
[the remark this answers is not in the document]
Wolog2021: Ok. Changing to UInt16 is easy enough. The defaults were selected to be friendly to my test and they are not rational. What do you think they should be @jordanschalm? I will update them and revise the test accordingly.

lib/go/test/flow_idtable_staking_test.go
Wolog2021: We envisioned that slot selection would run after removeUnapprovedNodes. To prepare for the test I create a series of new nodes. It is necessary to run "removeUnapprovedNodes" to get the system into the final state that is representative of what it would be before slot selection. Under the hood it's setting all of the initialWeight values, but I preferred to do that with the built-in function rather than forcing it.

transactions/idTableStaking/admin/set_slot_limit.cdc
```diff
@@ -0,0 +1,27 @@
...
+    // Local variable for a reference to the ID Table Admin object
+    let adminRef: &FlowIDTableStaking.Admin
+
+    prepare(acct: AuthAccount) {
+        // borrow a reference to the admin object
+        self.adminRef = acct.borrow<&FlowIDTableStaking.Admin>(from: FlowIDTableStaking.StakingAdminStoragePath)
+            ?? panic("Could not borrow reference to staking admin")
+    }
+
+    execute {
```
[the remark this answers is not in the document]
Wolog2021: Done

contracts/FlowIDTableStaking.cdc
```diff
@@ -870,6 +870,79 @@ pub contract FlowIDTableStaking {
...
+        /// Set slot (count) limits for each node type (role).
+        pub fun setSlotLimits(slotLimits: {UInt8: UInt8}) {
+            FlowIDTableStaking.account.load<{UInt8: UInt8}>(from: /storage/flowStakingSlotLimits)
+            FlowIDTableStaking.account.save(slotLimits, to: /storage/flowStakingSlotLimits)
+        }
+
+        /// Enforces slot (count) limits for each node type (role). If more nodes
+        /// of a given type are registered than the limit, they are removed
+        /// randomly until the limits are met. When a node is removed, it's
+        /// committed tokens are transfered to it's unstaked bucket.
```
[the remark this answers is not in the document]
Wolog2021: Done

contracts/FlowIDTableStaking.cdc
```diff
@@ -870,6 +870,79 @@ pub contract FlowIDTableStaking {
...
+            // count the nodes of each type
+            var nodeCounts: {UInt8: UInt8} = {1: 0, 2: 0, 3: 0, 4: 0, 5: 0}
+            for nodeID in allNodeIDs {
+                let nodeRecord = FlowIDTableStaking.borrowNodeRecord(nodeID)
+                if nodeRecord.initialWeight == 100 {
+                    nodeCounts[nodeRecord.role] = nodeCounts[nodeRecord.role]! + 1
+                }
+            }
+
+            for key in nodeCounts.keys {
```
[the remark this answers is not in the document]
Wolog2021: Done

contracts/FlowIDTableStaking.cdc
joshuahannan: We can change the name of the function if we want

contracts/FlowIDTableStaking.cdc
```diff
@@ -833,41 +875,55 @@ pub contract FlowIDTableStaking {
...
+
+            for nodeRole in currentNodeCount.keys {
+                if currentNodeCount[nodeRole]! >= slotLimits[nodeRole]! {
+                    //if all slots are full, remove and refund all pending nodes
+                    for nodeRecord in pendingNodes[nodeRole]! {
+                        self.removeAndRefundNodeRecord(nodeRecord: nodeRecord)
+                    }
+                } else if currentNodeCount[nodeRole]! + UInt16(pendingNodes[nodeRole]!.length) > slotLimits[nodeRole]! {
+                    //randomly remove pending nodes until the limit is met
+                    let deletionCount: UInt16 = currentNodeCount[nodeRole]! + UInt16(pendingNodes[nodeRole]!.length) - slotLimits[nodeRole]!
```
pgebheim, line 905: I think that's an attack here because of this method of deleting nodes given that currently there are likely to be few free slots.


Imagine:
nodeCount[nodeRole] == 5
slotLimits[nodeRole] == 6
pendingNodes[nodeRole] == 10000

This could happen by someone simply generating 10000 keys and doing the node registration transaction for them.

Now, when this function runs we need to reject 9999 nodes in order to fill 1 slot.

Instead, we should be choosing nodes to include. I understand that the above method gets around the duplicate issue. The options I see:

1) Track chosen nodes in a dictionary by id (such as my original example)
2) Scan the selected array for duplicates upon adding each.
3) Use "in place swap" method for choosing a random subset

Given the reality of the number of open slots I think (1) is fine.

pgebheim (changes requested):

contracts/FlowIDTableStaking.cdc
```diff
@@ -765,6 +766,14 @@ pub contract FlowIDTableStaking {
             return <-node
         }
 
+        pub fun setNodeWeight(nodeRecord: &NodeRecord, weight: UInt64) {
```
joshuahannan, line 769: I don't like the idea of the admin having a reference to a node record outside of the contract, so I'd rather the signature of this be `setNodeWeight(nodeID: String, weight: UInt64)`, then borrow the node record in the function and modify it there

contracts/FlowIDTableStaking.cdc
```diff
@@ -815,6 +824,36 @@ pub contract FlowIDTableStaking {
             FlowIDTableStaking.account.save(false, to: /storage/stakingEnabled)
         }
 
+        access(account) fun removeAndRefundNodeRecord(nodeRecord: &NodeRecord) {
```
joshuahannan, line 827: Can you add some comments above this function?

Wolog2021: Will do.

contracts/FlowIDTableStaking.cdc
Wolog2021: Sure. That's an easy enough change and I'll do it. I'm wondering if you could elaborate a little for my benefit on the risk with the reference.

contracts/FlowIDTableStaking.cdc
pgebheim: @Wolog2021 Pointed out that no matter what we need to go through all 10000 nodes in the above example, because the code needs to remove and refund those tokens. So you still always need to touch all nodes.

DUH

contracts/FlowIDTableStaking.cdc
```diff
@@ -833,41 +875,55 @@ pub contract FlowIDTableStaking {
...
+                } else if currentNodeCount[nodeRole]! + UInt16(pendingNodes[nodeRole]!.length) > slotLimits[nodeRole]! {
+                    //randomly remove pending nodes until the limit is met
+                    let deletionCount: UInt16 = currentNodeCount[nodeRole]! + UInt16(pendingNodes[nodeRole]!.length) - slotLimits[nodeRole]!
+                    var deletionArray: [UInt16] = []
+                    while UInt16(deletionArray.length) < deletionCount {
+                        let selection = UInt16(unsafeRandom() % UInt64(pendingNodes[nodeRole]!.length))
+                        if !deletionArray.contains(selection) {
+                            deletionArray.append(selection)
+                        }
+                    }
```
pgebheim, line 915: I'm still moderatley concerned about this. 

@joshuahannan Do you know if using a dictionary with a lookup here will actually be faster than this potential O(n^2) lookup?

Using random_shuffle we could still keep this whole thing to O(n)

contracts/FlowIDTableStaking.cdc
joshuahannan: Having the node reference outside of the contract allows the admin to have access to all of the nodes' tokens, which we don't want. I'm not even sure how you were expecting the reference to the `NodeRecord` to be returned from the contract anyway since the function that borrows it is`access(account)`

contracts/FlowIDTableStaking.cdc
joshuahannan: Can elaborate on the solution you're proposing? I'm not sure what using a dictionary with a lookup means here

contracts/FlowIDTableStaking.cdc
```diff
@@ -833,41 +878,55 @@ pub contract FlowIDTableStaking {
...
-                        // Request to unstake all tokens
-                        delRecord.tokensRequestedToUnstake = delRecord.tokensStaked.balance
+                   (nodeRecord.role != UInt8(5) && approvedNodeIDs[nodeID] == nil) {
+                    self.removeAndRefundNodeRecord(nodeRecord: nodeRecord)
+                } else {
+                    //capture whether this is a current node or a pending node
+                    if nodeRecord.initialWeight > 0 {
+                        currentNodeCount[nodeRecord.role] = currentNodeCount[nodeRecord.role]! + 1
+                    } else if candidateNodes!.contains(nodeID) {
+                        pendingNodes[nodeRecord.role]!.append(nodeRecord)
```
pgebheim, line 888: @joshuahannan Just thinking through this and currently there is an attack surface here because people could add lots of nodes.

What if we have a hard-limit of the number of new pending nodes per epoch right now in order to limit this?

contracts/FlowIDTableStaking.cdc
```diff
@@ -815,6 +825,38 @@ pub contract FlowIDTableStaking {
             FlowIDTableStaking.account.save(false, to: /storage/stakingEnabled)
         }
 
+        /// Removes nodes by setting their weight to zero and refunding
+        /// staked and delegated tokens.
+        access(account) fun removeAndRefundNodeRecord(nodeRecord: &NodeRecord) {
```
pgebheim, line 830: @joshuahannan Should this be `access(self)` since this is just a helper for `removeUnapprovedNodes` and it takes a reference?

Issue #308, opened 2022-08-25
[Rolling upgrade] Create and deploy contract for synchronizing versions


j1010001:
In order to synchronise EN version switchover, as described in a [design document](https://www.notion.so/dapperlabs/Execution-Nodes-rolling-updates-4a3984b868c24e3fb00e1c0d40857d8e?d=02e6681621724dd3a543ad9b1568a680#98da50ac701346e6b105f85ca5f95767) height to version map is to be stored as part of service account, and thus available to every node in an uniform manner.

It should be designed with https://app.zenhub.com/workspace/o/onflow/flow-go/issues/3047

This includes first actually defining and designing the best approach to the problem, and then implementing it.

# Requirements

- We have to release code changes to ENs in between sporks which requires us to have cleanly coordinated code upgrades when a given block height is encountered.
- By having a smart contract maintaining a mapping between height to version we can tightly control when the switchover to a given code version is to happen
- Needs to check input to ensure that a version is not proposed for the immediate next 1000 blocks, eg a cool down period  
- Should be part of the service account so that signers can be gathered for approving the change (will need to confirm the specifics)

sisyphusSmiling:
Posting questions here to close comms loop on a convo I began with @m4ksio 

> A few clarifying questions:
> 1. The contract will be managed by the Flow Service account yes? I'm assuming we'll want administration of this versioning contract to be handled by some unique resource that can later be handed off to some governance contract. Does that sound right?
> 2. Do we want the version to live centrally in the contract or be accessible within a resource owned by the Execution Nodes? My sense is a central mapping would be best, that way ENs could simply check the contract rather than need their resources updated if they go down for some period of time
> 3. Is it safe to assume that all versions will be named in accordance with [semver](https://semver.org/)? (e.g. 2.1.0) I saw [mentioned in a doc](https://www.notion.so/dapperlabs/Execution-Nodes-rolling-updates-4a3984b868c24e3fb00e1c0d40857d8e#8f7aed10031c44cba614653a820075b6) that version names may differ in ways that could make comparisons difficult

and his responses:

> 1. Good question, but I frankly... don't know. Is this how those things are usually done?
> 2. Yes, central place is best, since it will be identical for everyone
> 3. Yes, semver is fine. EN can have extra additions like v0.2.0-whatever but contract just needs a comparison so keeping `something like v0.2.0 is enough

[...]

PR #310, opened 2022-09-15
Sisyphus smiling/execution node versioning contract

sisyphusSmiling:
This PR addresses #308

Since this issue is high priority, I'm opening up this PR while I work on completing the test suite to get eyes on the contract's core functionality as well as supporting transactions & scripts.

On that note, you'll notice that the tests are written in JavaScript. That's because I'm not familiar with Go and wanted to prioritize getting the test done & tested.

## Semantic Version Representation

With regards to representing versions, `Semver` is a struct containing attributes mapping to semantic versioning standards. The struct also contains some custom comparators & a `toString()` method to make things easier at the script layer.

In the design doc referenced in #308, there was the question of signifying to ENs that updates earlier than the defined block height are allowed. To address this, I included `isBackwardsCompatible: Bool`. The idea is that if a `upcomingVersion.isBackwardsCompatible == true`, preemptive updates should be permissible. In a previous iteration, I thought of adding `oldestCompatibleVersion: Semver`, but that lead to recursive construction and it'd be versions all the way down.

### Purpose
The ExecutionNodeVersionBeacon contract is meant to address the need to coordinate software versions across a number of execution nodes (EN). I wanted to make it as easy as possible to expose the current labeled version requirement & the next upcoming version boundary.

### Usage by Execution Nodes

ENs can call `getCurrentExecutionNodeVersion()` to get the version they should be running at the moment in time the method is called.

Similarly, `getNextVersionBoundaryPair()` will return an array containing: `[blockHeightBoundary: UInt64, version: Semver]` if there is an upcoming version boundary defined. This tells them when they need to update and to what version.

Alternatively, callers can simply get the `versionTable` via `getVersionTable()` and do comparison or verification themselves.

Getters for other values including `versionUpdateBuffer` and `versionUpdateBufferVariance`are present as well.

### Contract Administration
Administration of the contract is handled via the `ExecutionNodeVersionKeeper` resource. Key functionality is found in the resource interface `ExecutionNodeVersionAdmin`

```
    pub resource interface ExecutionNodeVersionAdmin {

        /// Update the minimum version to take effect at the given block height
        pub fun addVersionBoundaryToTable(targetBlockHeight: UInt64, newVersion: Semver)

        /// Deletes the last entry in versionTable which defines an upcoming version boundary
        /// Could be helpful in case rollback is needed
        pub fun deleteUpcomingVersionBoundary(blockHeight: UInt64): Semver

        /// Updates the number of blocks that must buffer updates to the versionTable
        /// and the block number the version is targetting
        pub fun changeVersionUpdateBuffer(newUpdateBufferInBlocks: UInt64)

        /// Updates the variance percentage by which the versionUpdateBuffer can be changed
        /// Value must be between 0.0 and 1.0
        pub fun changeVersionUpdateBufferVariance(_ newUpdateBufferVariance: UFix64)
    }
```

Of course, adding to & deleting a version boundary from the table are pretty straight forward. Conditions around both of these actions protect the key expectation ENs have about version updates - that they will have at least `versionUpdateBuffer` number of blocks to react to a version update. Additionally, if this window is to change, it will only do so by a factor defined by `versionUpdateBufferVariance`.

The buffer and buffer variances can also be changed by the Admin, but the buffer must not bring the next block boundary within range the current block height. With this conditional, an edge case arose where the buffer is defined arbitrarily large and could not be reduced if an upcoming block boundary was defined. To prevent this, I added the buffer variance value so that it could at least be reduced by some factor.

## Other Details
Note that `archivedBlockBoundaries` and `upcomingBlockBoundaries` are included to maintain ordered arrays of historical and upcoming block boundaries. This made discovering upcoming version boundaries easier as well as determining compatibility of previous versions in `isCompatibleVersion()`

The method `archiveOldBlockBoundaries()` maintains these arrays based on the block height at the time it's called.


contracts/ExecutionNodeVersionBeacon.cdc
```diff
@@ -0,0 +1,425 @@
...
+                newUpdateBufferVariance >= 0.0 && newUpdateBufferVariance <= 1.0 : "Buffer variance must be value between 0 and 1."
+            }
+            post {
+                ExecutionNodeVersionBeacon.versionUpdateBufferVariance == newUpdateBufferVariance: "Update buffer variance was not properly changed! Reverted."
+            }
+        }
+    }
+
+    /// Admin resource that manages the Execution Node versioning
+    /// maintained in this contract
```
joshuahannan, line 173: This is the same comment that is above the previous resource interface definition. Which one is it supposed to be before?

contracts/ExecutionNodeVersionBeacon.cdc
```diff
@@ -0,0 +1,425 @@
...
+    /// Sorted Array containing future block heights where a version boundary is defined
+    /// We'll maintain this as easy mechanism for determining proximal version block boundaries
+    access(contract) var upcomingBlockBoundaries: [UInt64]
+
+    /// Sorted Array containing historical block heights where version boundaries were defined
+    access(contract) var archivedBlockBoundaries: [UInt64]
+
+    /// Admin interface that manages the Execution Node versioning
+    /// maintained in this contract
+    pub resource interface ExecutionNodeVersionAdmin {
```
joshuahannan, line 132: I'm curious why you made this interface. Do you expect your or other people to define other resources that implement this in the future?

contracts/ExecutionNodeVersionBeacon.cdc
```diff
@@ -0,0 +1,425 @@
...
+    pub fun getVersionTable(): {UInt64: Semver} {
+        return self.versionTable
+    }
+
+    /// Function that returns that version that was defined at the most
+    /// recent block height boundary
+    pub fun getCurrentExecutionNodeVersion(): Semver? {
+        pre {
+            self.versionTable.length > 0: "No current minimum version defined."
+        }
```
joshuahannan, line 274: I'd probably prefer to just return `nil` if there is an error here. no need to panic during a getter function

contracts/ExecutionNodeVersionBeacon.cdc
```diff
@@ -0,0 +1,425 @@
...
+            self.upcomingBlockBoundaries[0],
+            self.versionTable[self.upcomingBlockBoundaries[0]]
+        ]
+    }
+
+    /// Checks whether given version was compatible at the given block height
+    pub fun isCompatibleVersion(blockHeight: UInt64, version: Semver): Bool {
+        pre {
+            blockHeight >= getCurrentBlock().height: "Given block height is in the future."
+        }
```
joshuahannan, line 314: This pre-condition isn't really clear. The conditional is meant to be true here but the error message looks like it is stating the thing that is supposed to be true as an error. So I would either change it to make sure that the correct condition is evaluated and error message given, or just remove this precondition entirely and return `nil` if there is an error

transactions/executionNodeVersionBeacon/scripts/get_current_execution_node_version.cdc
```diff
@@ -0,0 +1,6 @@
+import ExecutionNodeVersionBeacon from 0x02
```
joshuahannan, line 1: I would make sure these have correct placeholder addresses

transactions/executionNodeVersionBeacon/scripts/get_current_execution_node_version_as_string.cdc
```diff
@@ -0,0 +1,11 @@
+import ExecutionNodeVersionBeacon from 0x02
```
joshuahannan, line 1: Need correct address here too

joshuahannan (commented):
Great work! I had no idea this contract was going to be so complex. I only reviewed the cadence code, so I'll trust you and the other reviewers to ensure that there is thorough unit tests.

transactions/executionNodeVersionBeacon/scripts/get_current_execution_node_version.cdc
sisyphusSmiling: Oops, thanks. Need to watch those copy-pastes from playground

contracts/ExecutionNodeVersionBeacon.cdc
sisyphusSmiling: I don't foresee others (outside of those managing EN versions) making implementations. It's primarily an artifact from scoping the contract, but I thought it might be helpful in case of a need for future extensibility. The immediate need doesn't demand an interface, however.

contracts/ExecutionNodeVersionBeacon.cdc
sisyphusSmiling: Thanks for catching that. I got my pre-condition reversed. But you're right, I'll return nil instead of throwing the error in the pre-condition.

contracts/ExecutionNodeVersionBeacon.cdc
```diff
@@ -0,0 +1,408 @@
...
+                // We can now safely change the versionUpdateBuffer
+                ExecutionNodeVersionBeacon.versionUpdateBuffer = newUpdateBufferInBlocks
+            }
+
+            emit ExecutionNodeVersionUpdateBufferChanged(newVersionUpdateBuffer: newUpdateBufferInBlocks)
+        }
+
+        /// Updates the variance percentage by which the versionUpdateBuffer can be changed
+        /// Value must be between 0.0 and 1.0
+        pub fun changeVersionUpdateBufferVariance(_ newUpdateBufferVariance: UFix64) {
```
satyamakgec, line 226: psychologically I don't understand the use of `updateBufferVariance` variable.

contracts/ExecutionNodeVersionBeacon.cdc
satyamakgec: I think it is impossible to test this because js testing doesn't support block height jumping.

contracts/ExecutionNodeVersionBeacon.cdc
```diff
@@ -0,0 +1,399 @@
...
+    /// enforced by insertion/deletion
+    access(contract) let versionTable: {UInt64: Semver}
+
+    /// Number of blocks between block height when version boundary is defined and
+    /// the version boundary being defined. Nodes can expect that changes to the versionTable
+    /// they will have at least this long to respond to a version boundary
+    access(contract) var versionUpdateBuffer: UInt64
+
+    /// Set expectations regarding how much the versionUpdateBuffer can vary between version boundaries
+    access(contract) var versionUpdateBufferVariance: UFix64
```
satyamakgec, line 114: @m4ksio @janezpodhostnik Do you guys think it is needed for the first iteration of the contract?  Highly Opinionated here I think it is a little bit over-engineered because, as a resource owner, I can change the `versionUpdateBufferVariance` freely, and that will allow me to change the `versionUpdateBuffer` freely as well, so in the end, the sole purpose of having the `versionUpdateBufferVariance` got defeated.

If you think it is needed, we can leave this as it is. Otherwise, I am in favour of dropping it; if it is needed later, we can add it again.

contracts/ExecutionNodeVersionBeacon.cdc
```diff
@@ -0,0 +1,408 @@
...
+            self.upcomingBlockBoundaries[0],
+            self.versionTable[self.upcomingBlockBoundaries[0]]
+        ]
+    }
+
+    /// Checks whether given version was compatible at the given historical block height
+    pub fun isCompatibleVersion(blockHeight: UInt64, version: Semver): Bool? {
+        if blockHeight > getCurrentBlock().height {
+            return nil
+        }
```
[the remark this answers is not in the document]
sisyphusSmiling: Agreed, returning false in that case would make sense here.

contracts/ExecutionNodeVersionBeacon.cdc
sisyphusSmiling: You make a good point Satyam. My thought behind introducing `versionUpdateBufferVariance` was to set expectations about how much the buffer would change between updates. 

To your point, it's probably overengineered. Especially because as you mentioned, the resource owner can change them as they wish without calling the designated methods or emitting the events. 

I think the best path forward for this iteration is to remove the variance and any references to it as well as set the `versionUpdateBuffer` to `access(self)`. That way changing the buffer value must go through the designated method and associated checks. The checks being that the change to the buffer doesn't cross a version boundary. 

Does that make sense?

sisyphusSmiling (commented):
Changes by @satyamakgec look good to me based on my understanding of the contract.

satyamakgec:
@m4ksio @j1010001 It is complete on the development side; when you get a chance to look and verify the functionality, please do and merge it as per your convenience.

m4ksio:
I can review that integration point, namely `pub fun getNextVersionBoundaryPair()` looks fine to me, hence approving

m4ksio (approved):

alilloig (approved):

alilloig:
Sorry I messed up with the CI with a commit for a couple of typos 🤦 it seems that it's needed to run some go command to update the go tests like in this commit https://github.com/onflow/flow-core-contracts/pull/310/commits/e7c3e1d1c557125a161c2527f97ba09e01b9d3cc please @satyamakgec do that before merging! 

Anyway due to branch protection rules we need @j1010001 approval for merging

j1010001 (approved):

joshuahannan (approved):

PR #321, opened 2022-11-15
FEATURE: Efficiency Refactoring and Slot Selection for Staking Contracts

joshuahannan:
Feature branch for big changes to the staking contracts to improve efficiency for epoch phase transition transactions and  algorithm for random slot selection for new nodes.

Current Changes:

## Changes approved node ID arguments to dictionaries instead of arrays
Cleans up some unnecessary code and should help with performance of certain transactions

## Stores stakedNodeIDs as a dictionary in account storage and uses it to iterate over instead of all IDs
Should improve the performance of the calculateRewards function because the function will only iterate through staked node IDs instead of all node IDs
Also significantly helps the performance of getter functions like getProposedNodes and getStakedNodes because they just return the list of node IDs instead of having to iterate through the list to construct the nodes that are staked

## Adds Pending Token Movements Dictionary

Adds a stored dictionary that represents nodes and delegators who have submitted a transaction to move tokens between buckets in some way. They are added to this record when they submit their transaction, not when the system chunk checks.

At the end of the epoch, when the moveTokens function is called, the function will only iterate through nodes and delegators who have submitted requests to move tokens, instead of checking all of the nodes and delegators.
This dictionary is cleared when the moveTokens function is called. Nodes who's unstaked tokens are moved to unstaking are automatically added to the dictionary for the next epoch.

This should significantly improve performance for the end epoch transactions.

This did not require any new tests because the existing tests were sufficient.

## Adds a Candidate Node List

A candidate node is defined as a node who is not participating/staked in the current epoch, but has committed enough tokens to participate in the next epoch. 

The candidate node list is a dictionary that maps roles to an array of node IDs:
```
{UInt8: [String]}
```

This is added to make the random slot selection process easier. This process will be much more efficient with the candidate node list because it can select nodes for each role separately instead of having to do them all in one loop.

Some tests were added to test this list as well as some testing utility functions

## Adds Events for Unstaking Requests

Events were added to mark when a node or delegator requests to unstake, indicating how much they requested
Additional `TokensUnstaked` event emissions were added on unstaking requests when tokens in the commited bucket are moved to the tokensUnstaked bucket. it isn't a full unstaking per se, but it matches the rest of the patterns that `TokensUnstaked` is emitted when tokens are moved into the tokens unstaked bucket.

### Role Slot Limits
`{UInt8: UInt16}`
Indicates how many of each node role can participate in each epoch

### Participant Node Role Counts
Indicates how many of each node role are currently staked and participating in the epoch

## Separates `removeAndRefundNodeRecord` into its own function
This process is now done multiple times for checking for approval, staking minimum, and random slot selection, so it is now its own method

## Adds `fillNodeRoleSlots` method to Admin resource
This method randomly select candidate nodes for each slot until there are no candidate nodes left or all the slots for a given role have been filled. Leftover nodes are refunded.
As part of this addition, the need for Access nodes (role=5) to be approved by the admin has been removed. Other node roles will receive this update in the future

## Adds `setNodeWeight` method to Admin resource
Allows the admin to set a node's weight to any number less than 100

NOTES FOR REVIEWERS: Ignore the `FlowIDTableStaking_new.cdc` file. I'm currently using this for testing because the flow-go dependencies don't like it when I make changes to the main contract file and some of the transactions.


contracts/FlowIDTableStaking.cdc
```diff
@@ -1262,10 +1268,21 @@ pub contract FlowIDTableStaking {
...
-    access(contract) fun setCurrentNodeList(_ nodeIDs: [String]) {
-        let list = self.account.load<[String]>(from: /storage/idTableCurrentList)
+    access(contract) fun setCurrentNodeList(_ nodeIDs: {String: Bool}) {
+        let list = self.account.load<{String: Bool}>(from: /storage/idTableCurrentList)
+
+        self.account.save<{String: Bool}>(nodeIDs, to: /storage/idTableCurrentList)
+    }
 
-        self.account.save<[String]>(nodeIDs, to: /storage/idTableCurrentList)
+    /// Gets the node list as a dictionary
```
jordanschalm, line 1277:
```suggestion
    /// Gets the current list of participant nodes as a dictionary.
    /// This represents the present identity table.
```

contracts/FlowIDTableStaking.cdc
```diff
@@ -1262,10 +1268,21 @@ pub contract FlowIDTableStaking {
...
+    pub fun getCurrentNodeList(): {String: Bool}? {
+        return self.account.copy<{String: Bool}>(from: /storage/idTableCurrentList)
+    }
+
+    /// Kept for backwards compatibility
```
jordanschalm, line 1282:
```suggestion
    /// Gets the current list of participant nodes (like getCurrentNodeList) but as a list.
    /// Kept for backwards compatibility
```

contracts/FlowIDTableStaking.cdc
```diff
@@ -1289,56 +1306,25 @@ pub contract FlowIDTableStaking {
...
-            // Access nodes have a minimum of 0, so they need to be strictly greater than zero to be considered proposed
-            if self.isGreaterThanMinimumForRole(numTokens: self.NodeInfo(nodeID: nodeRecord.id).totalCommittedWithoutDelegators(), role: nodeRecord.role)
-               && approved
-            {
-                proposedNodes.append(nodeID)
+        for nodeID in approvedNodeIDs.keys {
+            if FlowIDTableStaking.nodes[nodeID] == nil {
+                approvedNodeIDs[nodeID] = nil
+                continue
             }
```
jordanschalm, line 1316: Would it be possible instead to disallow adding a node ID to the approved list, if that node ID has not already been registered?

The ability to add an un-registered node ID to the approved list is a security issue, if there is an operational oversight, because then anyone can register an approved node ID in between it being approved and the operator registering. 

Currently we mitigate this problem via operational process, where we verify a node ID has been registered before approving it. However, it would remove the latitude for manual error, and potentially simplify the smart contract, if we disallowed this order of operations.

jordanschalm (commented):
Approach is looking good - nice improvements.
* Added a few naming/documentation suggestions
* Added a suggestion to disallow approving an un-registered node ID (fixes an outstanding potential security issue, and would simplify some code added in this PR)

contracts/FlowIDTableStaking.cdc
```diff
@@ -1262,10 +1268,21 @@ pub contract FlowIDTableStaking {
     }
 
     /// Sets a list of approved node IDs for the current epoch
```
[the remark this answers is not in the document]
joshuahannan: I think those make sense. I have been calling "registered" nodes "proposed" nodes, but I can see why registered might be better. I might have also just forgotten to change a few comments, which may have caused some confusion. I'll go through and try to standardize the terminology

contracts/FlowIDTableStaking.cdc
joshuahannan: yeah, I can add that check to the `setApprovedList` function. Do we want that admin check to fail if one of the approved nodes isn't in the ID table, or do we want it to succeed, but just remove that unregistered node from the approved list?

joshuahannan: I'm thinking we should probably just fail the transaction since it might be because of a typo or something

jordanschalm: Agree, I think we should fail the transaction.

contracts/FlowIDTableStaking.cdc
```diff
@@ -768,10 +768,16 @@ pub contract FlowIDTableStaking {
         /// Sets a list of approved node IDs for the next epoch
         /// Nodes not on this list will be unstaked at the end of the staking auction
         /// and not considered to be a proposed/staked node
-        pub fun setApprovedList(_ nodeIDs: [String]) {
-            let list = FlowIDTableStaking.account.load<[String]>(from: /storage/idTableApproveList)
+        pub fun setApprovedList(_ nodeIDs: {String: Bool}) {
+            for id in nodeIDs.keys {
+                if FlowIDTableStaking.nodes[id] == nil {
+                    panic("Approved node ".concat(id).concat(" does not already exist in the identity table"))
```
pgebheim, line 774: This seems like a nice clarification: Someone needs to add their node, then it needs to be approved before the Epoch transition.

contracts/FlowIDTableStaking.cdc
```diff
@@ -1262,10 +1274,22 @@ pub contract FlowIDTableStaking {
     }
 
     /// Sets a list of approved node IDs for the current epoch
-    access(contract) fun setCurrentNodeList(_ nodeIDs: [String]) {
-        let list = self.account.load<[String]>(from: /storage/idTableCurrentList)
+    access(contract) fun setParticipantNodeList(_ nodeIDs: {String: Bool}) {
```
pgebheim, line 1277: Good choice of `participant` here.

contracts/FlowIDTableStaking.cdc
```diff
@@ -1287,58 +1311,24 @@ pub contract FlowIDTableStaking {
...
-            if self.isGreaterThanMinimumForRole(numTokens: nodeRecord.tokensStaked.balance, role: nodeRecord.role)
-               && current
+            if !self.isGreaterThanMinimumForRole(numTokens: self.NodeInfo(nodeID: nodeRecord.id).totalCommittedWithoutDelegators(), role: nodeRecord.role)
             {
-                stakedNodes.append(nodeID)
+                proposedNodeIDs[nodeID] = nil
             }
         }
-        return stakedNodes
+        return proposedNodeIDs.keys
```
pgebheim, line 1331: So, since this returns `keys` here this would still include nodes that don't meet staking requirements. Which kinda makes the above loop not super useful, no?

joshuahannan: if an entry in a dictionary is set to `nil`, then it won't be included in the `keys` list any more afaik. I should test that though

jordanschalm: Seems like Josh is right, I checked this in a [Playground](https://play.flow.com/local-project?type=script&id=local-script-temp-0&storage=none).

I think we could change the code to make the behaviour more obvious -> **Suggestion:** use `proposedNodeIDs.remove(key: nodeID)` instead of setting to nil

jordanschalm (approved):

contracts/FlowIDTableStaking.cdc
joshuahannan: I refactored this method in another PR anyway so we don't use this any more. I changed the other usage of it though to use `remove`

contracts/FlowIDTableStaking.cdc
```diff
@@ -1069,15 +1071,17 @@ pub contract FlowIDTableStaking {
...
                 let nodeRecord = FlowIDTableStaking.borrowNodeRecord(nodeID)
 
-                FlowIDTableStaking.totalTokensStakedByNodeType[nodeRecord.role] = FlowIDTableStaking.totalTokensStakedByNodeType[nodeRecord.role]! + nodeRecord.tokensCommitted.balance
-
                 // mark the committed tokens as staked
                 if nodeRecord.tokensCommitted.balance > 0.0 {
+                    FlowIDTableStaking.totalTokensStakedByNodeType[nodeRecord.role] = FlowIDTableStaking.totalTokensStakedByNodeType[nodeRecord.role]! + nodeRecord.tokensCommitted.balance
                     emit TokensStaked(nodeID: nodeRecord.id, amount: nodeRecord.tokensCommitted.balance)
                     nodeRecord.tokensStaked.deposit(from: <-nodeRecord.tokensCommitted.withdraw(amount: nodeRecord.tokensCommitted.balance))
+                    stakedNodeIDs[nodeRecord.id] = true
```
satyamakgec, line 1084: Shouldn't we check for the minimum stake here?

contracts/FlowIDTableStaking.cdc
```diff
@@ -1069,15 +1071,17 @@ pub contract FlowIDTableStaking {
...
             for nodeID in allNodeIDs {
                 let nodeRecord = FlowIDTableStaking.borrowNodeRecord(nodeID)
 
-                FlowIDTableStaking.totalTokensStakedByNodeType[nodeRecord.role] = FlowIDTableStaking.totalTokensStakedByNodeType[nodeRecord.role]! + nodeRecord.tokensCommitted.balance
-
                 // mark the committed tokens as staked
                 if nodeRecord.tokensCommitted.balance > 0.0 {
+                    FlowIDTableStaking.totalTokensStakedByNodeType[nodeRecord.role] = FlowIDTableStaking.totalTokensStakedByNodeType[nodeRecord.role]! + nodeRecord.tokensCommitted.balance
                     emit TokensStaked(nodeID: nodeRecord.id, amount: nodeRecord.tokensCommitted.balance)
                     nodeRecord.tokensStaked.deposit(from: <-nodeRecord.tokensCommitted.withdraw(amount: nodeRecord.tokensCommitted.balance))
```
satyamakgec, line 1083: Instead of moving the token from one vault to another, why not keep the token in one vault and just update the variable? In that way, it is cheaper and faster.

contracts/FlowIDTableStaking.cdc
```diff
@@ -1191,11 +1199,15 @@ pub contract FlowIDTableStaking {
         pre {
             FlowIDTableStaking.stakingEnabled(): "Cannot register a node operator if the staking auction isn't in progress"
         }
-        let newNode <- create NodeRecord(id: id, role: role, networkingAddress: networkingAddress, networkingKey: networkingKey, stakingKey: stakingKey, tokensCommitted: <-tokensCommitted)
+        let newNode <- create NodeRecord(id: id, role: role, networkingAddress: networkingAddress, networkingKey: networkingKey, stakingKey: stakingKey, tokensCommitted: <-FlowToken.createEmptyVault())
```
satyamakgec, line 1202: Isn't it confusing here that in the `NodeRecord` we are committing an empty vault while at line 1208 we are emitting `TokensCommitted` event with a non-zero value?

contracts/FlowIDTableStaking.cdc
```diff
@@ -768,10 +768,16 @@ pub contract FlowIDTableStaking {
         /// Sets a list of approved node IDs for the next epoch
         /// Nodes not on this list will be unstaked at the end of the staking auction
         /// and not considered to be a proposed/staked node
-        pub fun setApprovedList(_ nodeIDs: [String]) {
-            let list = FlowIDTableStaking.account.load<[String]>(from: /storage/idTableApproveList)
+        pub fun setApprovedList(_ nodeIDs: {String: Bool}) {
```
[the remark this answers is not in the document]
joshuahannan: Contract upgrades don't allow adding fields to composite types, so the lists that we've added and are adding have to be stored in account storage instead of a field. Yes, the account owner can change it, but that is currently gated by the service account multisig, so no transactions would be able to be run without the service account committee approving it.

contracts/FlowIDTableStaking.cdc
```diff
@@ -768,10 +768,16 @@ pub contract FlowIDTableStaking {
...
+            for id in nodeIDs.keys {
+                if FlowIDTableStaking.nodes[id] == nil {
+                    panic("Approved node ".concat(id).concat(" does not already exist in the identity table"))
+                }
+            }
 
-            FlowIDTableStaking.account.save<[String]>(nodeIDs, to: /storage/idTableApproveList)
+            let list = FlowIDTableStaking.account.load<{String: Bool}>(from: /storage/idTableApproveList)
```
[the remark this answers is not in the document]
joshuahannan: We can't add variables in upgrades, so we have to hard code it

contracts/FlowIDTableStaking.cdc
joshuahannan: We originally did it this way because we wanted to have the protections that resources have for managing these tokens. I agree that it is cheaper and faster to do it with your suggestion, but this is what we're stuck with unless we do a big refactor

contracts/FlowIDTableStaking.cdc
joshuahannan: We already check for the minimum stake in the `removeUnapprovedNodes` method on line 837, so there is no need to check again here because any node with tokens in the committed bucket is already committed enough

contracts/FlowIDTableStaking.cdc
joshuahannan: I probably should have included this change in [the next PR](https://github.com/onflow/flow-core-contracts/pull/324) because it relates to the candidate node list. I didn't want to have duplicate code for adding it to the candidate node list, so I just created an empty vault here instead and committed tokens the regular way. Does that make sense?

contracts/FlowIDTableStaking.cdc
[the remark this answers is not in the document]
satyamakgec: That makes sense. Maybe it will be better to have a code comment on top of the contract about the design choice because by seeing this contract, other community members can follow the same trend, and they will miss the key security aspect (Revoked the key of the contract account) to work with this design.

contracts/FlowIDTableStaking.cdc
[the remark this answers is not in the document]
satyamakgec: Yeah, it make sense.. the effects of upgradability.

contracts/FlowIDTableStaking.cdc
satyamakgec: Cool, Make sense

joshuahannan:
I just merged the other two branches into this feature branch, but the CI is failing because the integration with flow-go is still broken. I'll talk with Jordan today about getting that resolved.

I've also written a transaction that we can use for the upgrade. I'll start putting it in a PR in the service account repo and a forum post so we can have a formal plan and announcement for it, but I'm leaving it here so people can start reviewing it.
Basically, there are some storage paths that we introduce as part of the upgrades that need to be initialized, and some existing paths that are having their types and values changed. The transaction performs all of those initializations and then does the upgrade.

```
import FlowIDTableStaking from 0xIDENTITYTABLEADDRESS

/// Transaction to upgrade the FlowIDTableStaking contract
/// to include performance improvements
/// and support permissionless node operation for access nodes
/// This transaction needs to be sent during the epoch setup
/// or epoch commit phase, because it initializes important
/// storage paths that are used to store fields
/// in the contract to track candidate nodes, staking slots,
/// limits, and more

transaction(code: [UInt8]) {

    prepare(stakingAccount: AuthAccount) {

        // Initialize Candidate Node List
        let emptyCandidateNodes: {UInt8: {String: Bool}} = {1: {}, 2: {}, 3: {}, 4: {}, 5: {}}
        stakingAccount.save(emptyCandidateNodes, to: /storage/idTableCandidateNodes)

        // Save Staking Slot Limits to storage
        // These are the current node counts, which access nodes increased by 5
        let slotLimits: {UInt8: UInt16} = {1: 108, 2: 117, 3: 7, 4: 85, 5: 117}
        stakingAccount.save(slotLimits, to: /storage/flowStakingSlotLimits)

        // Initialize the Moves pending list
        let movesPendingList: {String: {UInt32: Bool}} = {}
        stakingAccount.save<{String: {UInt32: Bool}}>(movesPendingList, to: /storage/idTableMovesPendingList)

        // Initialize the candidate Node Limits List
        // Currently, only access nodes are allowed as candidate nodes
        // because the other node roles are approval only
        let candidateNodeLimits: {UInt8: UInt64} = {1: 0, 2: 0, 3: 0, 4: 0, 5: 1000}
        stakingAccount.save<{UInt8: UInt64}>(candidateNodeLimits, to: /storage/idTableCandidateNodeLimits)

        // Borrow the admin to set minimum stake requirement for access nodes
        self.adminRef = acct.borrow<&FlowIDTableStaking.Admin>(from: FlowIDTableStaking.StakingAdminStoragePath)
            ?? panic("Could not borrow reference to staking admin")
        
        // Set Access node minimum stake requirement to 100 FLOW
        let minimums: {UInt8: UFix64} = FlowIDTableStaking.getMinimumStakeRequirements()
        minimums[5] = 100
        self.adminRef.setMinimumStakeRequirements(minimums)

        // Initialize the Staking Slot Counts
        // `getCurrentRoleNodeCounts` calculates the counts based on the current list
        // so we just store the result in storage after calculating
        Let slotCounts = FlowIDTableStaking.getCurrentRoleNodeCounts()
        stakingAccount.save(slotCounts, to: /storage/flowStakingRoleNodeCounts)

        // Approve List needs to be changed from [String] to {String: Bool}
        let approveList: [String] = stakingAccount.load<[String]>(from: /storage/idTableApproveList)
            ?? panic("Could not load the approve list from storage")

        // Create the approve list dictionary with the loaded value, then store the dictionary in the same path
        var approveDict: {String: Bool} = {}
        for nodeID in approveList {
            approveDict[nodeID] = true
        }

        // Participant Node list needs to be changed from [String] to {String: Bool}
        let participantList: [String] = stakingAccount.load<[String]>(from: /storage/idTableCurrentList)
            ?? panic("Could not load the participant list from storage")

        // Create the participant dictionary with the loaded value, then store the dictionary in the same path
        var participantDict: {String: Bool} = {}
        for nodeID in participantList {
            participantDict[nodeID] = true
        }

        acct.contracts.update__experimental(name: "FlowIDTableStaking", code: code)
    }
}
```

lib/go/test/flow_idtable_staking_test.go
```diff
@@ -776,16 +781,65 @@ func TestIDTableApprovals(t *testing.T) {
...
+	signAndSubmit(
+		t, b, tx,
+		[]flow.Address{idTableAddress},
+		[]crypto.Signer{IDTableSigner},
+		false,
+	)
 
-		err := tx.AddArgument(nodeIDDict)
+	// Register an access node
+	amountToCommit = 00000000
```
Kay-Zee, line 803: is it this many zeros on purpose?

contracts/FlowIDTableStaking.cdc
```diff
@@ -1710,16 +1710,35 @@ pub contract FlowIDTableStaking {
...
 
             let nodeRecord = FlowIDTableStaking.borrowNodeRecord(nodeID)
 
-            // To be considered proposed, a node has to have tokens staked + committed equal or above the minimum
-            if self.isGreaterThanMinimumForRole(numTokens: self.NodeInfo(nodeID: nodeRecord.id).totalCommittedWithoutDelegators(), role: nodeRecord.role)
-            {
+            let totalTokensCommitted = nodeRecord.nodeFullCommittedBalance()
+
+            let greaterThanMin = FlowIDTableStaking.isGreaterThanMinimumForRole(numTokens: totalTokensCommitted, role: nodeRecord.role)
+            let nodeIsApproved: Bool =  approvedNodeIDs[nodeID] != nil
```
vishalchangrani, line 1723: can `approvedNodeIDs[nodeID]`  ever be `false` (instead of `nil`)?

lib/go/test/flow_idtable_staking_test.go
joshuahannan: The UFix64 values have 8 decimal places, so that is how I was initializing them to be zero. I guess it could probably just be one zero for this value since it is mean to just be zero

contracts/FlowIDTableStaking.cdc
joshuahannan: not if we use the transactions properly, but I'll change it to `approvedNodeIDs[nodeID] ?? false` so it'll count as false if the value is also false

vishalchangrani (approved):

contracts/FlowIDTableStaking.cdc
```diff
@@ -1710,16 +1710,35 @@ pub contract FlowIDTableStaking {
     pub fun getProposedNodeIDs(): [String] {
```
jordanschalm, line 1710: I am confused about the intended separation of responsibilities between this function and `endStakingAuction`. In particular, with the new changes, this function seems to duplicate some checks from [`removeInvalidNodes`](https://github.com/onflow/flow-core-contracts/pull/321/files#diff-65336be374bb3fc9ad7b822243e065a389d73e758c7c16223e52fc3181cea59bR978). 
- We added the `movesPendingList` in order to limit the number of nodes we would need to check when processing the end of the staking auction, however here we are still iterating through all nodes.
- We are checking the same approval status and stake here as we already did in `removeInvalidNodes`

jordanschalm (commented):
Reviewed new commit https://github.com/onflow/flow-core-contracts/pull/321/commits/81cfbf1e413e803ab9c19d67bf250c10fd12b26d

transactions/idTableStaking/admin/add_approved_and_limits.cdc
```diff
@@ -0,0 +1,54 @@
...
+    prepare(acct: AuthAccount) {
+        // borrow a reference to the admin object
+        self.adminRef = acct.borrow<&FlowIDTableStaking.Admin>(from: FlowIDTableStaking.StakingAdminStoragePath)
+            ?? panic("Could not borrow reference to staking admin")
+    }
+
+    execute {
+		let nodeIDs = FlowIDTableStaking.getApprovedList()
+
+		let candidateNodeLimits = FlowIDTableStaking.getCandidateNodeLimits()
```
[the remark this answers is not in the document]
joshuahannan: We had decided to set the candidate limits to zero for non-access because we didn't want any nodes to register unless we already approved them, but I just realized that that would prevent any new nodes from registering until we increase the limit since we say that a node can't register unless there are open candidate slots.

I agree with your idea for when we have full staking auctions, but for now they've said to just keep the candidate slots closed. I think we still need at least one for each role though, but maybe this is a good enough reason to allow a lot of candidate slots for non-access nodes since spamming is disincentivized anyway by the minimum stake.

I'll also set those restrictions to the set limits methods

lib/go/test/test.go
```diff
@@ -251,6 +251,56 @@ func assertEqual(t *testing.T, expected, actual interface{}) bool {
 	return assert.Fail(t, message)
 }
 
+func assertCadenceArrayEqual(t *testing.T, expected, actual []cadence.Value) bool {
```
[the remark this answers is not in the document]
joshuahannan: We can't check order because we have no guarantee about ordering of elements in the arrays and dictionaries that we're having to verify. I'm only using these to verify arrays from dictionary keys in the staking contract, which never have duplicates, so maybe I should just put these methods in `staking_test_helpers` since they aren't general

lib/go/test/test.go
```diff
@@ -251,6 +251,56 @@ func assertEqual(t *testing.T, expected, actual interface{}) bool {
...
+				actual,
+			)
+
+			return assert.Fail(t, message)
+		}
+	}
+	return true
+}
+
+func assertCadenceNodeDictionaryEqual(t *testing.T, expected, actual []cadence.KeyValuePair) bool {
```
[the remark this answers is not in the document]
joshuahannan: We're only evaluating equalities of keys because we're comparing `{String: Bool}` dictionaries and the values shouldn't matter because they should never be `false`, only `true` or `nil`, but I'll add value equality in as well just to be sure. and I'll move this and the array method into `staking_test_helpers` and add more comments since they are specific to that.

contracts/FlowIDTableStaking.cdc
joshuahannan: You're right about it being more computationally intensive, but I was trying to make it so we don't have to do two upgrades. `FlowEpoch` calls this function to get the list of nodes for the next epoch and I wanted to make the return value of this function accurate to represent all the proposed nodes. If we wanted to upgrade the `FlowEpoch` contract also, we could just make the `endStakingAuction` return the proposed node IDs and use that for `EpochSetup` so we only have to iterate through it once, but we have to get this changes upgraded on testnet today, so I didn't want to make too many changes in case it takes too long to test and upgrade. What do you think?

joshuahannan: yeah, I think we should include this in a future upgrade. We just need to get this method working properly today, and I don't have time tonight to fully figure out the best way to do it and test it. Sorry about that. I'll create an issue.

transactions/idTableStaking/admin/add_approved_and_limits.cdc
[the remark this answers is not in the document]
joshuahannan: actually, I can't add those restrictions yet, because with what leadership has said about having minimal candidate slots, we can't have those restrictions because the staking slot limit would basically always be higher for non-access nodes. I'll create another issue to discuss it though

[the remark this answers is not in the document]
joshuahannan: We have more than a week before the mainnet upgrade, so we'll have time to discuss it

contracts/FlowIDTableStaking.cdc
jordanschalm: > we could just make the endStakingAuction return the proposed node IDs and use that for EpochSetup

👍 Agree with this suggestion, and getting rid of `getProposedList`. The proposed list used to mean, "if the staking auction ends right now, this represents what the identity table would be" -- it was useful as a way to read the tentative identity table between epoch phases, in addition to being used to read the final, real identity table by `FlowEpoch`. But with random slot selection, an accurate tentative identity table doesn't really exist.

transactions/idTableStaking/admin/add_approved_and_limits.cdc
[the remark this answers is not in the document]
joshuahannan: I talked to vishal and he is okay with setting the candidate limits to higher values since spam is disincentivized by the minimum stake requirement. I also just created an issue (#341) for the limits pre-conditions and we'll include that in the next upgrade

contracts/FlowIDTableStaking.cdc
joshuahannan: I created an issue for this too (#342) and we'll include it in the next upgrade

PR #323, opened 2022-11-30
Storage check payer exception

janezpodhostnik:


joshuahannan (commented):
Looks good to me. Do you plan to write any unit tests for it?

janezpodhostnik:
There were tests in the FVM, but just to make it more safe and secure, I added tests here too.

joshuahannan (approved):
Feel free to merge when you're ready

contracts/FlowStorageFees.cdc
```diff
@@ -71,26 +71,46 @@ pub contract FlowStorageFees {
...
+
+            capacities.append(self.accountBalanceToAccountStorageCapacity(balance)) 
         }
         return capacities
     }
 
+    // accountBalanceToAccountStorageCapacity returns the storage capacity of an account
+    // given the flow balance of the account.
+    pub fun accountBalanceToAccountStorageCapacity(_ balance: UFix64): UFix64 {
+            // get address token balance
```
SaveTheRbtz, line 104: nit: whitespace

Do we have something like `cadence fmt` for cadence scripts?

contracts/FlowStorageFees.cdc
```diff
@@ -71,26 +71,46 @@ pub contract FlowStorageFees {
...
+        if balance < self.minimumStorageReservation {
+            // if < then minimum return 0
+            return 0.0
+        } else {
```
SaveTheRbtz, line 108: nit: you can probably remove this `else`

SaveTheRbtz (approved):

contracts/FlowStorageFees.cdc
janezpodhostnik: nice catch.

> Do we have something like cadence fmt for cadence scripts?

not that I know of, but it would be amazing.

>  text seems outdated
not outdated, just badly worded. I changed it.

PR #325, opened 2022-12-02
Version contract

m4ksio:
Renames VersionBeacon contract, adds it to templates. Tests basic functionality.
It's a WiP, the contract itself will be updated later, but so far this enabled working on integration with flow-go


janezpodhostnik:
We probably shouldn't leave the temporary stuff in there, but looks good to me. I had some minor comments about naming, but they are about the test functions which don't really matter

alilloig:
Hi @m4ksio!! 

I made some changes to the contract according to https://github.com/onflow/flow-go/issues/3536

I don't really understand what you mean by service chunks and service events, but I made a separate method to emit the event as you suggested. That method also takes cares of populating the table according to the specs and the sequence number of the event.
I also introduced a transaction that can be used to add one or more versions and / or to delete one or more versions, that finally calls the function to emit the event. I wasn't sure about how you wanted the function for emitting the event to work, but if we want a single method on the Admin resource for doing the updates is just a matter of moving the logic of the new transaction to that method.

I couldn't run the js test, I'll ask @sisyphusSmiling after the holidays for help running them

joshuahannan:
@m4ksio can this be closed since we have #336 instead?

joshuahannan:
@alilloig see above comment

alilloig:
yeah I believe we can

PR #324, opened 2022-11-30
Adds candidate node list and tracking for nodes who have pending token movements

joshuahannan:
Builds upon the refactor in https://github.com/onflow/flow-core-contracts/pull/321 

## Adds Pending Token Movements Dictionary

Adds a stored dictionary that represents nodes and delegators who have submitted a transaction to move tokens between buckets in some way. They are added to this record when they submit their transaction, not when the system chunk checks.

At the end of the epoch, when the moveTokens function is called, the function will only iterate through nodes and delegators who have submitted requests to move tokens, instead of checking all of the nodes and delegators.
This dictionary is cleared when the moveTokens function is called. Nodes who's unstaked tokens are moved to unstaking are automatically added to the dictionary for the next epoch.

This should significantly improve performance for the end epoch transactions.

This did not require any new tests because the existing tests were sufficient.

## Adds a Candidate Node List

A candidate node is defined as a node who is not participating/staked in the current epoch, but has committed enough tokens to participate in the next epoch. 

The candidate node list is a dictionary that maps roles to an array of node IDs:
```
{UInt8: [String]}
```

This is added to make the random slot selection process easier. In the next PR, I'll add a feature similar to what was included in https://github.com/onflow/flow-core-contracts/pull/309 but this process will be much more efficient with the candidate node list because it can select nodes for each role separately instead of having to do them all in one loop.

Some tests were added to test this list as well as some testing utility functions

## Adds Events for Unstaking Requests

Events were added to mark when a node or delegator requests to unstake, indicating how much they requested
Additional `TokensUnstaked` event emissions were added on unstaking requests when tokens in the commited bucket are moved to the tokensUnstaked bucket. it isn't a full unstaking per se, but it matches the rest of the patterns that `TokensUnstaked` is emitted when tokens are moved into the tokens unstaked bucket.

PR #328, opened 2022-12-09
Adds slot limits and random slot selection using candidate node list

joshuahannan:
Builds upon the changes in https://github.com/onflow/flow-core-contracts/pull/321 and https://github.com/onflow/flow-core-contracts/pull/324. Review those in that order before reviewing this one

### Role Slot Limits
`{UInt8: UInt16}`
Indicates how many of each node role can participate in each epoch

### Participant Node Role Counts
Indicates how many of each node role are currently staked and participating in the epoch

## Separates `removeAndRefundNodeRecord` into its own function
This process is now done multiple times for checking for approval, staking minimum, and random slot selection, so it is now its own method

## Adds `fillNodeRoleSlots` method to Admin resource
This method randomly select candidate nodes for each slot until there are no candidate nodes left or all the slots for a given role have been filled. Leftover nodes are refunded.
As part of this addition, the need for Access nodes (role=5) to be approved by the admin has been removed. Other node roles will receive this update in the future

## Adds `setNodeWeight` method to Admin resource
Allows the admin to set a node's weight to any number less than 100


contracts/FlowIDTableStaking_new.cdc
```diff
@@ -1275,6 +1350,21 @@ pub contract FlowIDTableStaking {
             FlowIDTableStaking.account.save<Int>(newLimit, to: /storage/idTableCandidateNodeLimit)
         }
 
+        /// Set slot (count) limits for each node type (role).
```
jordanschalm, line 1353:
```suggestion
        /// Set slot (count) limits for each node type (role).
        /// The slot limit limits the number of participant nodes with the given role which may be added to the network.
        /// It only prevents candidate nodes from joining. It does not cause existing participant nodes to unstake,
        /// even if the number of participant nodes exceeds the slot limit.
```

contracts/FlowIDTableStaking_new.cdc
```diff
@@ -1633,6 +1733,14 @@ pub contract FlowIDTableStaking {
         // Save the candidate nodes limit
         FlowIDTableStaking.account.save<Int>(candidateNodeLimit, to: /storage/idTableCandidateNodeLimit)
 
+        let slotCounts: {UInt8: UInt16} = {1: 0, 2: 0, 3: 0, 4: 0, 5: 0}
+
+        // Save slot limits
+        FlowIDTableStaking.account.save(slotCounts, to: /storage/flowStakingSlotLimits)
```
jordanschalm, line 1739: I think defaulting to 0 for all node roles here would be a breaking change. For example, integration tests and transient networks like Localnet, Benchnet, and Canary would deploy this as soon as they pin this version of the contract. Then, all subsequent staking operations on those networks would fail.

Because of this, I'm inclined to default to unlimited instead of 0. Then staking will be functional with the default configuration.

contracts/FlowIDTableStaking_new.cdc
```diff
@@ -1275,6 +1350,21 @@ pub contract FlowIDTableStaking {
             FlowIDTableStaking.account.save<Int>(newLimit, to: /storage/idTableCandidateNodeLimit)
         }
 
+        /// Set slot (count) limits for each node type (role).
+        pub fun setSlotLimits(slotLimits: {UInt8: UInt16}) {
```
jordanschalm, line 1354: 🏁  There isn't a default value for this in the upgrade path. We will need to manually set this value while upgrading the contract.

jordanschalm (commented):
Generally looks good. Most comments are related to initializing new storage values when upgrading.

contracts/FlowIDTableStaking_new.cdc
joshuahannan: yep, that'll be part of my upgrade plan once this is all finalized

contracts/FlowIDTableStaking_new.cdc
```diff
@@ -1472,6 +1562,16 @@ pub contract FlowIDTableStaking {
         return FlowIDTableStaking.account.copy<{UInt8: [String]}>(from: /storage/idTableCandidateNodes) ?? {}
     }
 
+    /// Get slot (count) limits for each node type (role).
+    pub fun getRoleSlotLimits(): {UInt8: UInt16} {
+        return FlowIDTableStaking.account.copy<{UInt8: UInt16}>(from: /storage/flowStakingSlotLimits) ?? {}
```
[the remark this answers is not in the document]
joshuahannan: yeah, that is a good idea, I can change that

contracts/FlowIDTableStaking_new.cdc
```diff
@@ -1472,6 +1562,16 @@ pub contract FlowIDTableStaking {
...
+    }
+
+    /// Returns a dictionary that indicates how many nodes there are for each role
+    pub fun getCurrentRoleNodeCounts(): {UInt8: UInt16} {
+        return FlowIDTableStaking.account.copy<{UInt8: UInt16}>(from: /storage/flowStakingRoleNodeCounts) ?? {}
```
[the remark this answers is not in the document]
joshuahannan: yeah, that is probably a good idea

contracts/FlowIDTableStaking_new.cdc
joshuahannan: oh yeah, that is a good point. I was just setting it manually after deploying. I'll change that to be unlimited by default

contracts/FlowIDTableStaking_new.cdc
```diff
@@ -1633,6 +1733,14 @@ pub contract FlowIDTableStaking {
         // Save the candidate nodes limit
         FlowIDTableStaking.account.save<Int>(candidateNodeLimit, to: /storage/idTableCandidateNodeLimit)
 
+        let slotCounts: {UInt8: UInt16} = {1: 0, 2: 0, 3: 0, 4: 0, 5: 0}
+
+        // Save slot limits
+        FlowIDTableStaking.account.save(slotCounts, to: /storage/flowStakingSlotLimits)
+
+        // Save slot counts
+        FlowIDTableStaking.account.save(slotCounts, to: /storage/flowStakingRoleNodeCounts)
```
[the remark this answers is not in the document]
joshuahannan: yeah, that is basically what I was going to do in the upgrade transaction, but not directly in the method, but your way is much safer, so I'll include that

lib/go/test/flow_idtable_staking_test.go
```diff
@@ -2485,6 +2469,179 @@ func TestIDTableStaking(t *testing.T) {
...
+		false,
+	)
+
+	// see which node was removed
+	for _, event := range slotResult.Events {
+		if event.Type == "A.179b6b1cb6755e31.FlowIDTableStaking.NodeRemovedAndRefunded" {
+			eventValue := event.Value
+			removedNodeID = eventValue.Fields[0]
+		}
+	}
```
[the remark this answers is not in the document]
joshuahannan: We can't assert that it is a specific ID because the ID that is removed is random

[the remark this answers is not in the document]
joshuahannan: oh, nevermind, this is the test where we only register one node, so it is always bastianID. I've added the check

joshuahannan:
@jordanschalm @satyamakgec I've integrated the changes from the other PR and updated this one, so it is ready for review

contracts/FlowIDTableStaking.cdc
```diff
@@ -903,59 +953,89 @@ pub contract FlowIDTableStaking {
...
+                } else {
+                    nodeRecord.initialWeight = 100
+                }
+            }
+        }
 
-                    emit NodeRemovedAndRefunded(nodeID: nodeRecord.id, amount: nodeRecord.tokensCommitted.balance + nodeRecord.tokensStaked.balance)
+        /// Each node role only has a certain number of slots available per epoch
+        /// so if there are more candidate nodes for that role than there are slots
+        /// nodes are randomly selected from the list to be included
```
jordanschalm, line 976:
```suggestion
        /// nodes are randomly selected from the list to be included.
        /// Nodes which are not selected for inclusion are removed and refunded in this function.
        /// All candidate nodes left staked after this function exits are implicitly selected to fill the 
        /// available slots, and will become participants at the next epoch transition.
```

contracts/FlowIDTableStaking.cdc
```diff
@@ -874,6 +885,43 @@ pub contract FlowIDTableStaking {
             FlowIDTableStaking.account.save<{String: UFix64}>(nodeIDs, to: /storage/idTableNonOperationalNodesList)
         }
 
+        /// Removes nodes by setting their weight to zero and refunding
+        /// staked and delegated tokens.
+        access(self) fun removeAndRefundNodeRecord(nodeRecord: &NodeRecord) {
```
jordanschalm, line 890: Do we need to remove nodes from the candidate list in this code-path? 

For example, in `fillSlots`, we iterate over candidate nodes, and remove some of them using this function. Will those nodes be removed from the candidate list?

contracts/FlowIDTableStaking.cdc
```diff
@@ -903,59 +953,89 @@ pub contract FlowIDTableStaking {
...
 
                 let totalTokensCommitted = nodeRecord.nodeFullCommittedBalance()
 
+                let greaterThanMin = FlowIDTableStaking.isGreaterThanMinimumForRole(numTokens: totalTokensCommitted, role: nodeRecord.role)
+
                 // Remove nodes if they do not meet the minimum staking requirements
-                if !FlowIDTableStaking.isGreaterThanMinimumForRole(numTokens: totalTokensCommitted, role: nodeRecord.role) ||
-                   (approvedNodeIDs[nodeID] == nil) {
+                if (nodeRecord.role != UInt8(5) && (!greaterThanMin || approvedNodeIDs[nodeID] == nil))
+                   || (nodeRecord.role == UInt8(5) && approvedNodeIDs[nodeID] == nil && !greaterThanMin)
```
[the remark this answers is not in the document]
joshuahannan: great suggestion!

contracts/FlowIDTableStaking.cdc
joshuahannan: The candidate node list is fully loaded and then reset anyway during `fillSlots`, so it doesn't matter for the execution of that function.

I think it might matter for when we're removing nodes during `removeUnapprovedNodes`. Nodes are already removed from the candidate node list if they unstake below the minimum, so that part of it isn't a problem, so the only edge-case that would require it would be if a candidate node is removed from the approved list in the same staking auction where they first staked, which is an extremely small edge-case, but still probably worth considering. I'll add a check for that in `removeAndRefundNodeRecord`

jordanschalm (approved):

contracts/FlowIDTableStaking.cdc
```diff
@@ -903,59 +953,105 @@ pub contract FlowIDTableStaking {
...
             for nodeID in allNodeIDs {
                 let nodeRecord = FlowIDTableStaking.borrowNodeRecord(nodeID)
 
                 let totalTokensCommitted = nodeRecord.nodeFullCommittedBalance()
 
-                // Remove nodes if they do not meet the minimum staking requirements
-                if !FlowIDTableStaking.isGreaterThanMinimumForRole(numTokens: totalTokensCommitted, role: nodeRecord.role) ||
-                   (approvedNodeIDs[nodeID] == nil) {
+                let greaterThanMin = FlowIDTableStaking.isGreaterThanMinimumForRole(numTokens: totalTokensCommitted, role: nodeRecord.role)
+                let nodeIsApproved: Bool =  approvedNodeIDs[nodeID] != nil
```
satyamakgec, line 962: I am consfused here, if there is a value in approvedNodeIds for the corresponding `nodeID` then it would be true?

contracts/FlowIDTableStaking.cdc
```diff
@@ -840,6 +840,17 @@ pub contract FlowIDTableStaking {
...
 
+        /// Allows the protocol to set a specific weight for a node
+        /// if their staked amount changes or if they are removed
+        pub fun setNodeWeight(nodeID: String, weight: UInt64) {
+            if weight > 100 {
+                panic("Specified node weight out of range.")
+            }
+
+            let nodeRecord = FlowIDTableStaking.borrowNodeRecord(nodeID)
+            nodeRecord.initialWeight = weight
```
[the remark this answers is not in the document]
joshuahannan: Good question. @peterargue @vishalchangrani Do you think we should emit an event when a node's weight is changed by the protocol/governance?

contracts/FlowIDTableStaking.cdc
```diff
@@ -874,6 +885,43 @@ pub contract FlowIDTableStaking {
...
+            // move their committed tokens back to their unstaked tokens
+            nodeRecord.tokensUnstaked.deposit(from: <-nodeRecord.tokensCommitted.withdraw(amount: nodeRecord.tokensCommitted.balance))
+
+            // Set their request to unstake equal to all their staked tokens
+            // since they are forced to unstake
+            nodeRecord.tokensRequestedToUnstake = nodeRecord.tokensStaked.balance
+
+            // Iterate through all delegators and unstake their tokens
+            // since their node has unstaked
+            for delegator in nodeRecord.delegators.keys {
```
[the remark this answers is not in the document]
joshuahannan: There is no limit, but the service account does not have a gas limit

contracts/FlowIDTableStaking.cdc
joshuahannan: yeah, if the value for the node is non-nil, then that means it is approved

joshuahannan:
@jordanschalm Your reset epoch PR prompted me to consider if we needed to update the reset epoch method in `FlowEpoch` for these changes, considering that some of the "fields" like the candidate node list and staking slots are updated by some of these methods. 

It all seems fine to me, because regardless of whether we run it during the staking auction or during epoch setup/commit, we'll still be removing unapproved nodes and filling staking slots randomly for the next epoch, then moving tokens like usual, so the final state is effectively the same.

Just wanted to run it past you in case you notice something that I might not have noticed, but I don't think any changes are required.

jordanschalm:
> regardless of whether we run it during the staking auction or during epoch setup/commit, we'll still be removing unapproved nodes and filling staking slots randomly for the next epoch, then moving tokens like usual, so the final state is effectively the same.

I agree, the critical piece is in `fillSlots`, and this is tied to `endStakingAuction`, which is already correctly hooked into the epoch state machine. I checked through this codepath again it looked correct to me as well.

contracts/FlowIDTableStaking.cdc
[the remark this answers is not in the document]
vishalchangrani: yes - i think so. It will help us debug when a node is removed via governance and us know when exactly was a node weight modified.

PR #330, opened 2022-12-14
Re-Add calculateAccountsCapacity

janezpodhostnik:
I removed this function during https://github.com/onflow/flow-core-contracts/pull/323.

Which is a horrible Idea, as it will prevent the deployment of the upgraded contract before the spork, and after the spork there were be other problems as the new functions wont exist yet.

So it needs to be re-added.


joshuahannan (approved):
Why would it prevent the deployment of the upgraded contract? Functions can be added and removed

janezpodhostnik:
currently the FVM depends on this method, but after the spork the FVM will depend on `getAccountsCapacityForTransactionStorageCheck`

removing this method before the spork would cause problems. Similarly not having the new method after the spork would also cause problems.

PR #331, opened 2022-12-14
Change verifyPayersBalanceForTransactionExecution

janezpodhostnik:
`minimumRequiredBalance` should actually be equal to `reservedBalance`. 
The `maxTransactionFee` was removed from `minimumRequiredBalance` as it is generally much smaller then the `reservedBalance` and only covers some edge cases, so its simpler to just not include it (as pointed out by @bluesign [here](https://github.com/onflow/flow-go/pull/3473#issuecomment-1295783206)).


contracts/FlowFees.cdc
```diff
@@ -122,9 +121,15 @@ pub contract FlowFees {
         let tokenVault = payerAcct.borrow<&FlowToken.Vault>(from: /storage/flowTokenVault)
             ?? panic("Unable to borrow reference to the default token vault")
```
joshuahannan, line 123: Shouldn't this be deleted since you do it below?

janezpodhostnik: good catch! Forgot to delete it after changing it.

janezpodhostnik:
Any Idea why this would fail tests?

joshuahannan:
@sisyphusSmiling @alilloig Can y'all look at the CI here and try to help figure out why the JS tests are failing?

sisyphusSmiling:
@joshuahannan at a quick glance, I believe it's related to the updates to Flow CLI which make versions beyond 0.42.0 incompatible with flow-js-testing. I was able to fix this yesterday in [this PR to flow-ft](https://github.com/onflow/flow-ft/pull/107#issuecomment-1353643449), and will take a closer look in the morning

PR #336, opened 2023-01-12
Update version contract to emit service events

alilloig:
https://github.com/onflow/flow-go/issues/3536


alilloig:
I'm missing fixing the go tests / ci. The `ExecutionNodeBeacon` contract changed its name to `NodeBeacon` and I'm guessing that's what is making those fail (not sure at all). @joshuahannan or @janezpodhostnik could you help me fixing that? cheers!

joshuahannan:
@alilloig It looks like the CI is passing, or am I missing something?

contracts/NodeVersionBeacon.cdc
```diff
@@ -116,94 +120,133 @@ pub contract ExecutionNodeVersionBeacon {
...
                 key: targetBlockHeight,
                 newVersion
             )
             // Insert the version boundary's target block height into the
             // array maintaining all upcoming block height boundaries
-            ExecutionNodeVersionBeacon.insertUpcomingBlockBoundary(targetBlockHeight)
+            NodeVersionBeacon.insertUpcomingBlockBoundary(targetBlockHeight)
 
-            emit ExecutionNodeVersionTableAddition(height: targetBlockHeight, version: newVersion)
+            //emit NodeVersionTableAddition(height: targetBlockHeight, version: newVersion)
```
joshuahannan, line 146: Why is this commented?

contracts/NodeVersionBeacon.cdc
```diff
@@ -116,94 +120,133 @@ pub contract ExecutionNodeVersionBeacon {
...
             )!
-            ExecutionNodeVersionBeacon.upcomingBlockBoundaries.remove(
-                at: ExecutionNodeVersionBeacon.upcomingBlockBoundaries.firstIndex(of: blockHeight)!
+            NodeVersionBeacon.upcomingBlockBoundaries.remove(
+                at: NodeVersionBeacon.upcomingBlockBoundaries.firstIndex(of: blockHeight)!
             )
 
-            emit ExecutionNodeVersionTableDeletion(height: blockHeight, version: removed)
-
+            //emit NodeVersionTableDeletion(height: blockHeight, version: removed)
```
joshuahannan, line 171: also commented?

contracts/NodeVersionBeacon.cdc
```diff
@@ -116,94 +120,133 @@ pub contract ExecutionNodeVersionBeacon {
...
+        ///
+        /// @param newVersions
+        /// @param deletedVersions
+        ///
+        pub fun emitNodeVersionTableUpdated() {
+            
+            let updatedTable: {UInt64: Semver} = {}
+
+            // Assure that there is at least one working version for the current block
+            let currentVersionBlock = NodeVersionBeacon.getNextVersionBoundaryPair()[0] as! UInt64
```
joshuahannan, line 188: Why index 0? Could this be a constant somewhere?

contracts/NodeVersionBeacon.cdc
```diff
@@ -116,94 +120,133 @@ pub contract ExecutionNodeVersionBeacon {
...
+                    updatedTable[versionBlock] = NodeVersionBeacon.versionTable[versionBlock]
+                }
+            }
+
+            emit NodeVersionTableUpdated(versionTable: updatedTable,
+                sequence: NodeVersionBeacon.nextTableUpdatedSequence)
+
+            NodeVersionBeacon.nextTableUpdatedSequence = NodeVersionBeacon.nextTableUpdatedSequence + 1
+
+        }
```
joshuahannan, line 203: Can you add some more comments here to describe what this is doing?

contracts/NodeVersionBeacon.cdc
```diff
@@ -116,94 +120,133 @@ pub contract ExecutionNodeVersionBeacon {
...
-                ExecutionNodeVersionBeacon.versionUpdateBuffer = newUpdateBufferInBlocks
+                NodeVersionBeacon.versionUpdateBuffer = newUpdateBufferInBlocks
             }
 
-            emit ExecutionNodeVersionUpdateBufferChanged(newVersionUpdateBuffer: newUpdateBufferInBlocks)
+            emit NodeVersionUpdateBufferChanged(newVersionUpdateBuffer: newUpdateBufferInBlocks)
+        }
+
+        init() {
         }
```
joshuahannan, line 240: Empty init method?

contracts/NodeVersionBeacon.cdc
```diff
@@ -354,14 +397,15 @@ pub contract ExecutionNodeVersionBeacon {
 
     init(versionUpdateBuffer: UInt64) {
         /// Initialize contract variables
-        self.ExecutionNodeVersionAdminStoragePath = /storage/ExecutionNodeVersionAdmin
+        self.NodeVersionAdminStoragePath = /storage/NodeVersionAdmin
```
joshuahannan, line 400: Probably should use a more specific path name if possible

joshuahannan (commented):
I thought this contract was already deployed on mainnet? Is the goal to upgrade or deploy a new one?

contracts/NodeVersionBeacon.cdc
alilloig: It should be deleted, the goal of this is to have a function that can be call just to emit the event rather than having an event emitted every time the table is updated.

contracts/NodeVersionBeacon.cdc
alilloig: Absolutely, sorry about that, I didn't considered the code done until Maks have check on it (I actually have to introduce some more changes to the function)

contracts/NodeVersionBeacon.cdc
alilloig: NodeVersionBeaconAdmin perhaps?

contracts/NodeVersionBeacon.cdc
alilloig: I needed to look for the block for the current Version on the table, and Gio already had done that method that returns a 2 position array, where the first position is that block. I was trying to introduce minimal changes in this contract but I can agree that this does not look pretty at all

alilloig:
> @alilloig It looks like the CI is passing, or am I missing something?

@joshuahannan https://github.com/onflow/flow-core-contracts/actions/runs/3950960032

alilloig:
This now should work as @m4ksio was expecting. I have included some comments to make it easier to understand what is doing. I agree with @joshuahannan that we should make a nicer / safer way for checking for the current version block, I'll borrow @sisyphusSmiling during this week for checking how can we make that while not changing unnecessarily  the contract

contracts/NodeVersionBeacon.cdc
joshuahannan: maybe `FlowNodeVersionBeaconAdmin`

contracts/NodeVersionBeacon.cdc
```diff
@@ -90,14 +90,18 @@ pub contract ExecutionNodeVersionBeacon {
     }
 
     /// Event emitted any time a change is made to versionTable
-    pub event ExecutionNodeVersionTableAddition(height: UInt64, version: Semver)
-    pub event ExecutionNodeVersionTableDeletion(height: UInt64, version: Semver)
+    // pub event NodeVersionTableAddition(height: UInt64, version: Semver)
+    // pub event NodeVersionTableDeletion(height: UInt64, version: Semver)
+    pub event NodeVersionTableUpdated(
+        versionTable: {UInt64: Semver},
```
m4ksio, line 96: We need this table as an array of pairs, ordered. Something like
`[ { height, semver}, { height, semver }]`

m4ksio:
We don't need `isBackwardsCompatible` field in semver, it should be removed.

contracts/NodeVersionBeacon.cdc
alilloig: Something like this would make sense?
```
    pub event NodeVersionTableUpdated(
        versionTable: [{UInt64: Semver}],
        sequence: UInt64
    )
```

Since we only have arrays and dictionaries I can't think of other way of storing that info (each array position will be a dictionary with a single entry)

sisyphusSmiling: Some alternatives to an array of dictionaries would be an array of structs defining a version boundary, something like:

```
pub struct VersionBoundary {
    pub let blockHeight: UInt64
    pub let version: Semver
}
```

so the emitted event type would be `[ VersionBoundary, ... ]`

Or, you could emit an array of arrays, each sub-array of length 2 - `[AnyStruct; 2]`. The emitted value would look  like `[ [ height, semver ], ... ]`

alilloig: Thanks Gio! @m4ksio do you have any preference?

m4ksio: I think the version proposed by @sisyphusSmiling would be great - an array of structs is most concrete data type.

alilloig:
@sisyphusSmiling @joshuahannan if you have the bandwidth take a look at it during this week please!

lib/go/test/node_version_beacon_test.go
```diff
@@ -0,0 +1,198 @@
...
+			templates.GenerateEmitVersionTableScript(env),
+			versionBeaconAddress)
+
+		txCheckResults := signAndSubmit(t, b, checkTx,
+			[]flow.Address{versionBeaconAddress},
+			[]crypto.Signer{versionBeaconSigner},
+			false,
+		)
+
+		assert.Len(t, txCheckResults.Events, 1)
```
joshuahannan, line 156: Would be good to add comments here and other similar places when you do an assert describing why the result should be what you're asserting

joshuahannan (approved):
Some of the formatting of the transactions is a little bit weird. Also, for the import placeholders, we'd like to refactor everything to use the flow.json relative imports, so can you change all of those to have those kinds of imports? Thanks!

lib/go/test/node_version_beacon_test.go
alilloig: I believe @m4ksio did those tests, I made a test case for this on the js tests.

joshuahannan:
@alilloig I just merged a few PRs, so you'll need to update your branch and run `make test` again. Is this almost ready to merge? I'm putting a new release together and it would be nice to include this in it

alilloig:
@m4ksio we found that the new import syntax only works while `flow dev` is running, and that is only possible at the moment if all the cadence folders (contracts, transactions, scripts) is under a folder named `cadende`, so we are not changing it for now

PR #350, opened 2023-02-15
add candidate node check for participant nodes

joshuahannan:
A candidate node is defined as a node who is currently not participating, but is eligible for the next epoch. When checking if a node is eligible for candidate node status, the contract was only checking if their staked balance was zero and they were above the minimum. This was a problem after the upgrade because all of the existing approved access nodes had zero stake but were still participant nodes, so if any of them tried to stake their 100 minimum FLOW, they would be considered a participant node and be included in the random slot selection. 

This change adds a condition to `isEligibleForCandidateNodeStatus()` that first checks if they are already a participant node before checking the other conditions and adding them to the candidate node list


contracts/FlowIDTableStaking.cdc
```diff
@@ -422,8 +422,12 @@ pub contract FlowIDTableStaking {
         /// which means that it is a new node who currently is not participating with tokens staked
         /// and has enough committed for the next epoch for its role
         access(self) fun isEligibleForCandidateNodeStatus(_ nodeRecord: &FlowIDTableStaking.NodeRecord): Bool {
+            let participantList = FlowIDTableStaking.getParticipantNodeList()!
```
jordanschalm, line 425: This would account for the case where an operator adds tokens to an existing participant node. 

What if an approved node un-stakes then re-stakes?

joshuahannan: If they un-stake and then re-stake in the same epoch, then they're still in the participant list, so this would still return false because they are still in the participant list. 

If they un-stake below the minimum in one epoch and then re-stake above the minimum in a future epoch, they will have been removed from the participant list, so this will return true because they are no longer a participant and can be considered a candidate node.

Does that answer your question?

jordanschalm: > If they un-stake below the minimum in one epoch and then re-stake above the minimum in a future epoch, they will have been removed from the participant list, so this will return true because they are no longer a participant and can be considered a candidate node.

If an Access Node on the allow-list went through this process (un-stake then re-stake in a future epoch), wouldn't we have the same problem?
- The AN would be considered a candidate node (`isEligibleForCandidateNodeStatus == true`)
- The AN would then be subject to slot selection
- If the AN does not win a slot, they would be removed and refunded but still be included in `getProposedNodeIDs`

jordanschalm: A few days ago on Slack we were discussing having `removeAndRefundNode` also remove a node from the allow-list. I think incorporating that change here would solve this problem.

joshuahannan: you're right. I'll update `removeAndRefundNode` to remove it from the approve list

joshuahannan:
@jordanschalm I updated the PR so it is ready for review again. I had to make `getApproveList` return an optional to avoid an infinite loop situation with `setApproveList` and `removeAndRefundNodeRecord`

contracts/FlowIDTableStaking.cdc
```diff
@@ -894,10 +898,19 @@ pub contract FlowIDTableStaking {
...
         /// staked and delegated tokens.
         pub fun removeAndRefundNodeRecord(_ nodeID: String) {
+
             let nodeRecord = FlowIDTableStaking.borrowNodeRecord(nodeID)
 
             emit NodeRemovedAndRefunded(nodeID: nodeRecord.id, amount: nodeRecord.tokensCommitted.balance + nodeRecord.tokensStaked.balance)
 
+            // remove the refunded node from the approve list
+            // We only do this if the approve list returns a non-nil value
+            // to avoid the infinite loop with `setApprovedList`
```
jordanschalm, line 908: I don't really understand how this prevents the recursive call back to `removeAndRefundNodeRecord` from `setApprovedList`...

In general `/storage/idTableApproveList` is not empty, so `getApprovedList` always returns a non-nil value. I'm not understanding how this relates to the recursive call to `removeAndRefundNodeRecord` 🤔

lib/go/test/flow_idtable_staking_test.go
```diff
@@ -927,6 +927,53 @@ func TestIDTableApprovals(t *testing.T) {
...
+	result = executeScriptAndCheck(t, b, templates.GenerateGetCommittedBalanceScript(env), [][]byte{jsoncdc.MustEncode(cadence.String(accessID))})
+	assertEqual(t, CadenceUFix64("0.0"), result)
+	result = executeScriptAndCheck(t, b, templates.GenerateGetStakedBalanceScript(env), [][]byte{jsoncdc.MustEncode(cadence.String(accessID))})
+	assertEqual(t, CadenceUFix64("0.0"), result)
+	result = executeScriptAndCheck(t, b, templates.GenerateReturnCurrentTableScript(env), nil)
+	idArray = result.(cadence.Array).Values
+	assert.Len(t, idArray, 1)
+	assertEqual(t, CadenceString(accessID), idArray[0])
+
+	t.Run("Adding new stake to a zero-stake access node who is already a participant should not add them to the candidate list", func(t *testing.T) {
```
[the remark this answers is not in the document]
joshuahannan: yeah, I'll add that

contracts/FlowIDTableStaking.cdc
joshuahannan: in `setApproveList()`, we load the approve list from storage, which clears the storage, so `getApproveList` would return `nil` since there is nothing there.

jordanschalm: I see. But we still execute `removeAndRefundNodeRecord` twice - are we sure that is OK?

An example this being problematic is, we emit `NodeRemovedAndRefunded` twice when `removeAndRefundNodeRecord` is the entrypoint:
- `removeAndRefundNodeRecord` is called
- `NodeRemovedAndRefunded` is emitted (1)
- `setApprovedList` is called
- `removeAndRefundNodeRecord` is recursively called from `setApprovedList`
- `NodeRemovedAndRefunded` is emitted (2)
- since `getApprovedList` returns `nil` we avoid calling `setApprovedList` again, but still execute the rest of `removeAndRefundNodeRecord` 

In general, relying on the behaviour and ordering of operations in `setApprovedList` seems brittle to me. 

What about splitting the functions into smaller pieces to avoid the two-way coupling? 
```cadence
pub fun removeAndRefundNodeRecord(...) {
  self.unsafeSetApprovedList(...)
  self.unsafeRemoveAndRefundNodeRecord(nodeRecord)
}

// sets the approved list without validating it (requires caller to validate)
priv fun unsafeSetApprovedList(...) {}

// removes and refunds the node record without also removing them from the approved-list.
priv fun unsafeRemoveAndRefundNodeRecord() {}

// can call unsafeRemoveAndRefundNodeRecord rather than removeAndRefundNodeRecord
// then use unsafeSetApprovedList
pub fun setApprovedList(...) {}
```

joshuahannan: Good suggestion! I did that in my latest commit. I also reorganized the position of some methods in the Admin object

lib/go/test/flow_idtable_staking_test.go
```diff
@@ -927,6 +944,91 @@ func TestIDTableApprovals(t *testing.T) {
...
+	require.NoError(t, err)
+
+	signAndSubmit(
+		t, b, tx,
+		[]flow.Address{idTableAddress},
+		[]crypto.Signer{IDTableSigner},
+		false,
+	)
+
+	t.Run("Adding new stake to a non-participant, approved, zero-stake access node should add them to the candidate list", func(t *testing.T) {
```
jordanschalm, line 1008: Originally I was thinking this case should result in the node _not_ being added to the candidate list. But `removeAndRefundNode` will now remove the node from the approved-list. So we won't encounter the original problem.

The side effect of this is, approved nodes are subject to the random slot selection just like un-approved nodes. I think this is OK though.

jordanschalm (approved):

lib/go/test/flow_idtable_staking_test.go
joshuahannan: yeah, that was intentional. We just had a conversation about this last week with the team and decided that it is ok for now

PR #352, opened 2023-02-21
Update calculation for automated staking rewards

joshuahannan:
In preparation for enabling automated staking rewards on Flow mainnet, this PR adds a few lines to change how rewards are calculated. There was a batch of tokens from Flow Genesis that were minted as a necessary incentive structure for proper decentralization. These tokens are not circulating and were never meant to be included in the true total supply calculation. Because FLOW has hit its decentralization milestones, these tokens are meant to be burned. Most of them have been burnt and will be burnt in the future, but unfortunately, it has been difficult to access all of them at once to burn them, so there are still some remaining.

This change adds a section when calculating the reward amount for the next epoch that subtracts the current amount of bonus tokens remaining from the expected total supply so that the inflation amount is calculated from the expected supply not including bonus tokens instead of the total supply. This will result in less tokens being minted per epoch than there would be if the bonus tokens were included.

The process for updating this value will go through the service account committee. When bonus tokens are burnt, a proposal will be posted as a PR in the service account repo where the committee will review it and schedule a multisign to update the value. These changes will be very infrequent because we would prefer to burn large batches of bonus tokens instead of little amounts every time.

The initial value is expected to be something around 20M


transactions/idTableStaking/admin/add_approved_and_limits.cdc
```diff
@@ -20,40 +20,30 @@ transaction(ids: [String]) {
...
-		for newNodeID in ids {
-			if nodeIDs[newNodeID] != nil {
+		for newNodeID in newApprovedIDs {
+			if existingApprovedIDs[newNodeID] != nil {
     			continue
 			}
 
 			let nodeInfo = FlowIDTableStaking.NodeInfo(newNodeID)
 
-			candidateNodeLimits[nodeInfo.role] = candidateNodeLimits[nodeInfo.role]! + 1
```
jordanschalm, line 41: Why are we not updating candidate limits anymore? Are we going to keep these at high values? (`candidateLimit >> nodeRoleLimit`)

transactions/idTableStaking/admin/add_approved_and_limits.cdc
```diff
@@ -8,7 +8,7 @@ import FlowIDTableStaking from 0xIDENTITYTABLEADDRESS
 /// If any of the provided nodes already exist in the ID table, this
```
jordanschalm, line 8: Could we update the documentation above, lines 5-6:
```
/// It also increases the slot limits by the number of nodes which are added. 
```

transactions/epoch/admin/set_bonus_tokens.cdc
```diff
@@ -0,0 +1,15 @@
...
+/// because they are not meant to be a part of the total supply
+/// 
+/// Eventually, all bonus tokens will be burned and the amount will be zero
+
+transaction(bonusTokenAmount: UFix64) {
+    prepare(signer: AuthAccount) {
+        signer.load<UFix64>(from: /storage/FlowBonusTokenAmount)
+        signer.save(bonusTokenAmount, to: /storage/FlowBonusTokenAmount)
+    }
+}
```
jordanschalm, line 15: If the bonus tokens are stored in a relatively small number of known accounts, could we hard-code those addresses, and read their balance, instead of manually keeping this field in sync? 

I'm not sure how the bonus tokens are stored, so not sure how feasible this is. It would be nice to have the `bonusTokens` value stay in sync automatically, though, if possible.

contracts/FlowIDTableStaking.cdc
```diff
@@ -1212,16 +1212,14 @@ pub contract FlowIDTableStaking {
...
                 for delegator in rewardBreakdown.delegatorRewards.keys {
                     let delRecord = nodeRecord.borrowDelegatorRecord(delegator)
                     let delegatorReward = rewardBreakdown.delegatorRewards[delegator]!
                         
                     delRecord.tokensRewarded.deposit(from: <-rewardsVault.withdraw(amount: delegatorReward))
-
-                    emit DelegatorRewardsPaid(nodeID: rewardBreakdown.nodeID, delegatorID: delegator, amount: delegatorReward)
                 }
+
+                emit RewardsPaid(nodeID: rewardBreakdown.nodeID, amount: nodeReward, delegatorAmounts: rewardBreakdown.delegatorRewards)
```
jordanschalm, line 1222: Have we communicated this proposal to the community at all yet?

jordanschalm (approved):
Depending on how the bonus tokens are structured, I suggested a possible alternative which would avoid needing service account approval for each bonus token burn. Not sure if it is feasible though.

bluesign:
Why don't we 'just' burn them ?

transactions/idTableStaking/admin/add_approved_and_limits.cdc
joshuahannan: Yeah, we already have them at high enough values that we don't need this increase, and when we get to adding in the pre-condition, we'll update them again so that the conditions you recommended are satisfied.

transactions/epoch/admin/set_bonus_tokens.cdc
joshuahannan: I agree, but I think that they are stored in a bunch of accounts. That is a good suggestion though. I'm asking to see if that would be possible

contracts/FlowIDTableStaking.cdc
joshuahannan: Not yet, but I wanted a review first, then we'll start sharing it around for feedback and approvals

bluesign:
@joshuahannan ah makes sense, thanks

satyamakgec (approved):

transactions/epoch/admin/set_bonus_tokens.cdc
joshuahannan: We discussed more and we're going to keep it this way because monitoring the different accounts was going to be more operationally and technically difficult. We think requiring the multisig is good for this because we can just do it at the same time as the burn

joshuahannan:
@jordanschalm We discussed the rewards event emissions and decided that we still wanted it to be in a different transaction than the move tokens transaction, so I moved it to happen during the staking auction phase only when the `currentBlock.view == (currentEpochMetadata.startView + UInt64(1))`. That way, rewards are paid in the block that is immediately after the start block of the epoch. It is in the latest commit.

Is this a safe way to do this? Is it possible for a view to get skipped or for a view to not change at all between blocks? 
There is an additional check in the `payRewardsForPreviousEpoch` method that ensures that rewards can only be paid once even if the method is called multiple times, but it still felt good to be safe. I primarily want to make sure this view won't get skipped because then it would skip paying rewards.

bluesign:
> Is this a safe way to do this? Is it possible for a view to get skipped or for a view to not change at all between blocks? There is an additional check in the `payRewardsForPreviousEpoch` method that ensures that rewards can only be paid once even if the method is called multiple times, but it still felt good to be safe. I primarily want to make sure this view won't get skipped because then it would skip paying rewards.

this would not work, afaik view is increased for each collection.

bluesign:
> Is this a safe way to do this?

I don't think this works:

```cadence 
  pub fun main():UInt64 {
   var b1 =  getCurrentBlock()
   var dif1 = (b1.height - b1.view)

   var b2 = getBlock(at: b1.height-100)!
   var dif2 = (b2.height - b2.view)

   return dif2 - dif1
  }
```

jordanschalm:
> Is this a safe way to do this? Is it possible for a view to get skipped or for a view to not change at all between blocks?

No it is not safe. @bluesign is right, views can be skipped.

joshuahannan:
okay that makes sense. So do we think we're good with just having the staking auction part of `advanceBlock()` be:
```
                case EpochPhase.STAKINGAUCTION:
                    let currentBlock = getCurrentBlock()
                    let currentEpochMetadata = FlowEpoch.getEpochMetadata(FlowEpoch.currentEpochCounter)!
                    // Pay rewards only if automatic rewards are enabled
                    // payRewardsForPreviousEpoch will only pay rewards once because it sets a rewards paid flag
                    // the first time it pays and won't pay rewards any more for that epoch once the flag is set
                    // even though the method is called every time
                    if FlowEpoch.automaticRewardsEnabled() {
                        self.payRewardsForPreviousEpoch()
                    }
```
Added clarification in the comments there too since rewards can only be paid once anyway

contracts/epochs/FlowEpoch.cdc
```diff
@@ -353,6 +353,11 @@ pub contract FlowEpoch {
                 case EpochPhase.STAKINGAUCTION:
                     let currentBlock = getCurrentBlock()
                     let currentEpochMetadata = FlowEpoch.getEpochMetadata(FlowEpoch.currentEpochCounter)!
+                    // Pay rewards only if automatic rewards are enabled and the current view
+                    // is the one immediately after the start view of the currenct epoch
```
janezpodhostnik, line 357: > and the current view is the one immediately after the start view of the currenct epoch

                                                                     ^ typo

where is this condition in the code?

contracts/epochs/FlowEpoch.cdc
```diff
@@ -493,6 +495,14 @@ pub contract FlowEpoch {
                 flowTotalSupplyAfterPayout = FlowToken.totalSupply + (currentPayout - feeAmount)
             }
 
+            // Load the amount of bonus tokens from storage
+            let bonusTokens = FlowEpoch.getBonusTokens()
+
+            // Subtract bonus tokens from the total supply to get the real supply
+            if bonusTokens < flowTotalSupplyAfterPayout {
```
janezpodhostnik, line 502: if we dont think this can be false i think the else branch should panic

contracts/epochs/FlowEpoch.cdc
```diff
@@ -795,6 +805,12 @@ pub contract FlowEpoch {
         return self.account.copy<Bool>(from: /storage/flowAutomaticRewardsEnabled) ?? false
     }
 
+    /// Gets the current amount of bonus tokens left to be destroyed
```
janezpodhostnik, line 808: I think a more detailed description might be nice here

contracts/epochs/FlowEpoch.cdc
joshuahannan: I'm not sure. The only way this could be false would be if we accidentally set it to a number that is too high and I don't know if that is something that warrants going into EECC and requiring a spork

contracts/epochs/FlowEpoch.cdc
joshuahannan: added

janezpodhostnik (approved):

yagop:
> Now, the delegator reward amounts are just emitted in the RewardsPaid event instead, which should significantly decrease the number of events emitted in this transaction.

Even though the number of event's will be much lower, the  `RewardsPaid.delegatorAmounts` will be large too.

> This will be a breaking change if any apps or delegators are relying on the DelegatorRewardsPaid events for their app, so they will need to update to be compatible. We will make sure to broadcast this change very widely to all affected users if it is approved.

This will cause problems as to process rewards events from previous blocks.

Wouldn't be better to address the large number of events by other way? There are many delegators receiving very small amounts like 0.00000006, probably because they can't pay for the transaction fee or something. It makes sense to limit the minimum amount to have staked as to receive rewards

yagop:
Glad to help, many thanks @joshuahannan ! :bow:

PR #359, opened 2023-03-31
Version beacon contract changes

janezpodhostnik:
Made some changes to the version beacon contract.

1. Moved the emit event function to its own resource, because it should never move off the service account, while the admin will probably move to a separate account in the future. Also renamed it to heartbeat to signify that it is called often, and doesn't just emit the event but also keeps track of what the next version boundary is.
2. There is now a zeroth version that is always present at block 0, and simplifies some things as consumers do not need to deal with nulls.
3. The VersionBeacon event is also emitted when passing a boundary (for some redundancy)
4. The VersionBeacon event also contains the current (previous if exactly on a boundary) version the nodes should be on (more redundancy)
5. "buffer" is now called "freeze period" which makes more sense to me
6. There is no function to get the entire list of versions (because that doesn't scale), but there is a paged function to get it by pages.
7. there is a single array of all past and future boundaries and a pointer to the closest future one instead of 2 lists
8. more comments
9. more renaming


contracts/NodeVersionBeacon.cdc
```diff
@@ -85,337 +105,414 @@ pub contract NodeVersionBeacon {
...
+    /// freeze period is measured in blocks (from the current block).
+    pub event NodeVersionBoundaryFreezePeriodChanged(freezePeriod: UInt64)
+
+    /// Canonical storage path for the NodeVersionBeacon.Admin resource.
+    pub let AdminStoragePath: StoragePath
+
+    /// Canonical storage path for the NodeVersionBeacon.Heartbeat resource.
+    pub let HeartbeatStoragePath: StoragePath
+
+    /// Block heigh indexed version boundaries.
```
joshuahannan, line 157:
```suggestion
    /// Block height indexed version boundaries.
```

contracts/NodeVersionBeacon.cdc
```diff
@@ -85,337 +105,414 @@ pub contract NodeVersionBeacon {
...
-        /// Could be helpful in case rollback is needed
-        pub fun deleteUpcomingVersionBoundary(blockHeight: UInt64): Semver {
-            pre {
-                NodeVersionBeacon.versionTable.keys.contains(blockHeight): "No boundary defined at that blockHeight."
-                NodeVersionBeacon.upcomingBlockBoundaries.contains(blockHeight): "Boundary not defined in upcomingBlockBoundaries."
+    /// Heartbeat resource that emits the version beacon event and keeps track of upcoming versions.
+    /// This resource should always be held only by the service account,
+    /// because the service account should be the only one emitting the event, 
+    /// and only during the system transaction
+    pub resource Heartbeat {
```
joshuahannan, line 289: Is it a problem that we also use `Heartbeat` as the name for the epoch resource that advances epoch phases? Might be a bit confusing?

I guess they both effectively are doing some action every block so it might make sense that they have the same names

joshuahannan (approved):
I don't remember, has this contract already been deployed on any networks? These changes would be incompatible if so.

Also, I don't know anything about the business logic for this feature, so I mostly just reviewed the cadence code quality and such. We should find someone else who is more familiar with the requirements to review this and/or potentially get someone else up to speed

contracts/NodeVersionBeacon.cdc
janezpodhostnik: Exactly, both functions have the same use: they need to be called during the system transaction every block. So I thought it might be better that they have the same name.

janezpodhostnik:
I still need to test this in tandem with FVM changes, so I am not ready to merge it yet.

janezpodhostnik:
The contract was not yet deployed anywhere. It will go on the service account.

PR #364, opened 2023-04-18
FEATURES: June Epoch Upgrade PR

### LINKED ISSUE #363: Introduce Delegator Staking Minimums
### Issue To Be Solved
Based on [FLIP 78](https://github.com/onflow/flips/pull/78), we are planning on requiring delegators to have a minimum stake.

### Suggest A Solution

* Delegator staking minimum should be stored at index 0 of the minimum staking requirements dictionary
* In `moveTokens`, if the delegator is below the minimum, move their tokens to unstaked or unstaking, depending on where they already are

joshuahannan:
Includes multiple upgrades that will be going out near the June spork

# Introduce Delegator Staking Minimums

Closes #363 

Based on [FLIP78-Delegator Staking Minimums](https://github.com/onflow/flips/pull/78)

Adds functionality for enforcing that delegators are above a minimum stake to be staked in an epoch.
If a delegator is below the minimum, they will be unstaked during the move tokens operation.
The minimum is stored in key-0 of the existing staking minimum dictionary.
Delegators cannot register unless they provide the minimum.

The current minimum is expected to be 50, but that value can and will probably change in the future.

This change will be breaking for anyone who is using the [basic register delegator transaction](https://github.com/onflow/flow-core-contracts/pull/364/files#diff-a6b98dd066be78bb8617b2b2cf79cc3ca402fb98c3437aa49d3b2997640b6268L3-L8) and the [Locked Tokens register delegator transaction](https://github.com/onflow/flow-core-contracts/pull/364/files#diff-4b12a2f9cda0bf338a12e014869873d01ff1799a7352b367b2004e9c9f5f712dL27-L33) because it adds a new parameter to specify the number of tokens to commit.

# Use capabilities for epoch admins

Adds `EpochOperations` resource interfaces for the `FlowIDTableStaking`, `FlowClusterQC`, and `FlowDKG` `Admin` resources so that these resources can be moved to the service account on the production networks. These interfaces restrict the functionality that is accessible by the epoch smart contracts and epoch account to only that which is required for automated epoch functionality.

New paths are used for the capabilities so that the resources can remain in the same paths for testing purposes

Also moves `resetEpoch()` into the `FlowEpoch.Admin` resource instead of the `FlowEpoch.Heartbeat` resource.


satyamakgec (approved):

contracts/FlowIDTableStaking.cdc
```diff
@@ -77,6 +77,13 @@ pub contract FlowIDTableStaking {
...
     /// The minimum amount of tokens that each node type has to stake
     /// in order to be considered valid
+    /// Keys:
+    /// 0 - Delegators
+    /// 1 - Collector Nodes
+    /// 2 - Consensus Nodes
+    /// 3 - Execution Nodes
+    /// 4 - Verification Nodes
+    /// 5 - Access Nodes
     access(account) var minimumStakeRequired: {UInt8: UFix64}
```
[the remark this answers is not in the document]
joshuahannan: We didn't have enums when we first deployed this contract, and adding one now would be a little awkward because we already have one in the `FlowEpoch` contract for it, so we don't want two different versions of it or a circular dependency.

contracts/FlowIDTableStaking.cdc
```diff
@@ -77,6 +77,13 @@ pub contract FlowIDTableStaking {
 
     /// The minimum amount of tokens that each node type has to stake
     /// in order to be considered valid
+    /// Keys:
+    /// 0 - Delegators
```
jordanschalm, line 81: I am not a big fan of combining the delegator minimum stake with the per-role stake minimums. I think we should prioritize conceptual precision over the convenience of re-using an existing storage slot.

The staking contract has the business logic concepts:
- a node role represents the set of responsibilities and capabilities associated with a node running in the network (collection, consensus, access, verification, execution)
- a staking mechanism is a way to stake tokens and receive rewards (registration, delegation)

I can register a node with a particular role. Then someone can delegate tokens to that node. I cannot register a node with role `delegator`. They are conceptually distinct and should be differentiated in the business logic.

In addition, several comments are now incorrect or imprecise with this change.

> /// Checks if the amount of tokens is greater than the minimum staking requirement for the specified role

> /// The minimum amount of tokens that each node type has to stake in order to be considered valid

Delegators are not a node type or role.

I am being a bit pedantic here, but this is a core system contract. It is important that it is as easy as possible to read and reason about, both for maintainers and community members. Conflating delegation minimums with node role minimums introduces a fairly small additional cognitive overhead to reasoning about the contract ("node roles _but also_ delegator minimums"), but the _but also_ moments add up to making bugs more difficult to spot, and the behaviour less comprehensible to community members. 

Concretely, I would suggest to store the delegator minimum stake as a separate config field with distinct documentation, setters, and getters.

joshuahannan: Thanks for the feedback! That is a great point. I do really like the idea of keeping it a part of the shared field with the node minimums though because it saves some lines of code and streamlines the contract in other small ways, but I understand the potential confusion. The field is just called `minimumStakeRequired`, so it could include delegators if we wanted to expand the definition of the term `role` and fix the documentation. I'm a little torn. I'll see if I can get another opinion.

bluesign: >  The field is just called minimumStakeRequired, so it could include delegators if we wanted to expand the definition of the term role and fix the documentation. I'm a little torn. I'll see if I can get another opinion.

I think this is very good idea.

bjartek: I agree with @jordanschalm here.  Readability is important and some more lines of code here is worth it to be correct.

joshuahannan: Not sure how to resolve this. It looks like there is fairly equal support for both paths and I still am on the side of doing it how it is already done here.

joshuahannan:
I posted in the forum about the plans to upgrade: https://forum.onflow.org/t/breaking-change-coming-to-the-flow-epoch-smart-contracts-to-introduce-delegator-staking-minimums/4928

Expected dates to get the changes in effect is as follows:

CanaryNet: Monday, June 26th, 2023
Testnet: Wednesday, June 28th, 2023

And assuming that the upgrades on the canary and testnet go well,

Mainnet: Wednesday, July 5th, 2023

The upgrade plan with transactions is detailed in this service account PR: https://github.com/onflow/service-account/pull/243

PR #365, opened 2023-04-24
Adds rewards calculation for sporks

joshuahannan:
During a spork, the `FlowEpoch.heartbeat.resetEpoch()` method is called after the spork is complete. 
If automatic rewards are enabled, the reward breakdown for the current epoch needs to be calculated and rewards for the next epoch need to be calculated.
This PR adds `calculateAndSetRewards()` to the `resetEpoch()` method so this is done properly and automatic rewards can be paid after a spork. 

The creation of new epoch metadata has also been moved to before the start of the new epoch in order to have valid epoch metadata for the rewards calculations to modify. This is why it now creates an `EpochMetadata` with  `counter + 1`

This also removes the rewards parameter from `resetEpoch()` because it is not used any more


satyamakgec (approved):
LGTM

contracts/epochs/FlowEpoch.cdc
```diff
@@ -426,7 +426,6 @@ pub contract FlowEpoch {
         pub fun resetEpoch(
             currentEpochCounter: UInt64,
             randomSource: String,
-            newPayout: UFix64?,
```
jordanschalm, line 429: ⚠️  This is a breaking change - we'll need to update the [spork util](https://github.com/onflow/flow-go/blob/c534287f8f66653f4dd582a3aa3187d65e564b61/cmd/util/cmd/epochs/cmd/reset.go) as well. 

When we merge this, let's make sure to communicate the breaking change to spork operators.

lib/go/test/flow_epoch_test.go
```diff
@@ -1278,4 +1274,93 @@ func TestEpochReset(t *testing.T) {
...
 	})
+
+	result := executeScriptAndCheck(t, b, templates.GenerateGetFlowTotalSupplyScript(env), nil)
+	assertEqual(t, CadenceUFix64("7000000000.0"), result)
+
+	t.Run("Can reset the epoch during the staking auction with automatic rewards enabled", func(t *testing.T) {
+
+		tx := createTxWithTemplateAndAuthorizer(b, templates.GenerateEpochSetAutomaticRewardsScript(env), idTableAddress)
+
+		_ = tx.AddArgument(cadence.NewBool(true))
```
jordanschalm, line 1285:
```suggestion
		tx.AddArgument(cadence.NewBool(true))
```

lib/go/test/flow_epoch_test.go
```diff
@@ -1278,4 +1274,93 @@ func TestEpochReset(t *testing.T) {
...
+		tx = createTxWithTemplateAndAuthorizer(b, templates.GenerateEpochPayRewardsScript(env), idTableAddress)
+
+		signAndSubmit(
+			t, b, tx,
+			[]flow.Address{idTableAddress},
+			[]crypto.Signer{IDTableSigner},
+			false,
+		)
+
+		// Verifies that the rewards from the previous epoch does not include the new epoch's amount
```
jordanschalm, line 1341: When we spork, the rewards for the epoch pre-spork will be paid out when the `resetEpoch` transaction is submitted. Currently this is the next day, on Thursday. It would be ideal if this multisig could be moved to Wednesday, so there isn't a 1-day delay in reward payouts, once this is enabled. cc @vishalchangrani

transactions/epoch/admin/reset_epoch.cdc
```diff
@@ -15,7 +15,6 @@ import FlowIDTableStaking from 0xIDENTITYTABLEADDRESS
 
 transaction(currentEpochCounter: UInt64,
```
jordanschalm, line 16: We should update these as well:
-  https://github.com/onflow/service-account/blob/main/templates/reset_epoch.cdc
- https://github.com/onflow/service-account/blob/main/templates/reset_epoch_with_end_staking_auction.cdc

jordanschalm (approved):

lib/go/test/flow_epoch_test.go
joshuahannan: Agreed, it would be nice to reset on the same day as the spork

contracts/epochs/FlowEpoch.cdc
joshuahannan: Yep, thanks for flagging!

contracts/epochs/FlowEpoch.cdc
```diff
@@ -451,23 +450,28 @@ pub contract FlowEpoch {
...
+            // with the new values
             let newEpochMetadata = EpochMetadata(
-                    counter: FlowEpoch.currentEpochCounter,
+                    counter: currentEpochCounter + 1,
                     seed: randomSource,
                     startView: startView,
                     endView: endView,
                     stakingEndView: stakingEndView,
+                    // This will be overwritten in `calculateAndSetRewards` below
                     totalRewards: FlowIDTableStaking.getEpochTokenPayout(),
```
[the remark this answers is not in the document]
joshuahannan: Good comment, I changed it

contracts/epochs/FlowEpoch.cdc
jordanschalm: Here's the PR for the spork util: https://github.com/onflow/flow-go/pull/4286

PR #367, opened 2023-05-02
Restricted capabilities for epoch admins

joshuahannan:
This PR adds `EpochOperations` resource interfaces for the `FlowIDTableStaking`, `FlowClusterQC`, and `FlowDKG` `Admin` resources so that these resources can be moved to the service account. These interfaces restrict the functionality that is accessible by the epoch smart contracts and epoch account to only that which is required for automated epoch functionality.

New paths are used for the capabilities so that the resources can remain in the same paths for testing purposes

Currently, the `FlowClusterQC.Admin` does not need to be restricted at all because all of its methods are epoch operation methods, but the interface is still used in case there are restrictions needed in the future.

Also moves `resetEpoch()` into the `FlowEpoch.Admin` resource instead of the `FlowEpoch.Heartbeat` resource.


joshuahannan:
@jordanschalm The changes to the epoch contract is causing the `flow-go` dependency to fail. When you find the time after you've done an initial pass here, can you help create a new branch in `flow-go` that includes these changes so I can also do a new emulator version with the updated contracts? I'll be able to get the tests running after that.

contracts/epochs/FlowEpoch.cdc
```diff
@@ -338,6 +338,60 @@ pub contract FlowEpoch {
             FlowEpoch.account.load<Bool>(from: /storage/flowAutomaticRewardsEnabled)
             FlowEpoch.account.save(enabled, to: /storage/flowAutomaticRewardsEnabled)
         }
+
+        /// Protocol can use this to reboot the epoch with a new genesis
+        /// in case the epoch setup phase did not complete properly
+        /// before the end of an epoch
```
jordanschalm, line 344:
```suggestion
        /// Ends the currently active epoch and starts a new one with the provided configuration.
        /// The new epoch, after resetEpoch completes, has `counter = currentEpochCounter + 1`.
        /// This function is used during sporks - since the consensus view is reset, and the protocol is
        /// bootstrapped with a new initial state snapshot, this initializes FlowEpoch with the first epoch
        /// of the new spork, as defined in that snapshot.
```

contracts/epochs/FlowEpoch.cdc
```diff
@@ -674,24 +674,36 @@ pub contract FlowEpoch {
...
+            ?? panic("Could not borrow reference to staking admin")
 
         return adminRef
     }
 
     /// Borrow a reference to the ClusterQCs Admin resource
     access(contract) fun borrowClusterQCAdmin(): &FlowClusterQC.Admin {
-        let adminRef = self.account.borrow<&FlowClusterQC.Admin>(from: FlowClusterQC.AdminStoragePath)
-            ?? panic("Could not borrow qc admin")
+        let adminCapability = self.account.copy<Capability>(from: FlowClusterQC.AdminStoragePath)
```
jordanschalm, line 689: Just wondering - why is it using `copy` instead of `getCapability`?

contracts/FlowIDTableStaking.cdc
```diff
@@ -834,10 +834,24 @@ pub contract FlowIDTableStaking {
...
         }
     }
+
+    /// Interface that only contains operations that are part
+    /// of the regular automated functioning of the epoch process
+    /// These are accessed by the `FlowEpoch` contract through a capability
+    pub resource interface EpochOperations {
+        pub fun setEpochTokenPayout(_ newPayout: UFix64)
+        pub fun setSlotLimits(slotLimits: {UInt8: UInt16})
+        pub fun setNodeWeight(nodeID: String, weight: UInt64)
```
jordanschalm, line 844:
```suggestion
```
These two are not used by `FlowEpoch`, from what I can tell. I'm wondering if they should be part of this interface.

contracts/epochs/FlowEpoch.cdc
jordanschalm, line 689: Should we define a separate path for the `EpochOperations` capability? My guess is it would be better to have separate paths for capabilities vs resources, in case they are stored in the same account.

contracts/epochs/FlowEpoch.cdc
```diff
@@ -674,24 +674,36 @@ pub contract FlowEpoch {
 
     /// Borrow a reference to the FlowIDTableStaking Admin resource
     access(contract) fun borrowStakingAdmin(): &FlowIDTableStaking.Admin {
```
jordanschalm, line 676:
```suggestion
    access(contract) fun borrowStakingAdmin(): &FlowIDTableStaking.Admin{FlowIDTableStaking.EpochOperations} {
```

Based on the error in the CI logs, I'm guessing this is causing the `flow-go` failure you mentioned? If so, `borrowClusterQCAdmin` would need a similar update.

> panic: failed to deploy Epoch contract: [Error Code: 1101] error caused by: 1 error occurred:
> 	* transaction execute failed: [Error Code: 1101] cadence runtime error: Execution failed:
> error: cannot deploy invalid contract
>   --> 6f7d6e0ffbb786cbf619254e6b3f9aa1b64d45896a6f8385480ef9fcdf10e62f:23:1
>    |
> 23 | 	serviceAccount.contracts.add(
>    | 	^
> 
> error: mismatched types
>    --> f8d6e0586b0a20c7.FlowEpoch:684:15
>     |
> 684 |         return adminRef
>     |                ^^^^^^^^ expected `&FlowIDTableStaking.Admin`, got `&FlowIDTableStaking.Admin{FlowIDTableStaking.EpochOperations}`

jordanschalm:
@joshuahannan Regarding the `flow-go` failure, from the logs it seems like the contract is invalid independent of `flow-go`? It is reporting a type error when deployment is attempted. See [this comment](https://github.com/onflow/flow-core-contracts/pull/367/files#r1183895325). 

Also, now that `FlowEpoch` requires the admin capability (rather than resource) to be stored in the account, I think deploying these contracts will result in a non-working configuration, as the Heartbeat will fail to load the required capability. **Suggestion:** update the constructors to store both the Admin resource and the capability to the account where the contract is deployed, as the default configuration.

contracts/epochs/FlowEpoch.cdc
joshuahannan: `getCapability` only works for getting a capability directly from a path that has a link, which can only be private or public paths. Since this is a capability from another account, we have to copy it from storage

contracts/FlowIDTableStaking.cdc
joshuahannan: They're not used right now, but I could see them needing to be used in the future when we have permissionless node operation and slashing. Do you think we should just wait until then to include them?

contracts/epochs/FlowEpoch.cdc
joshuahannan: Based on our conversations, there is only ever going to be one of each epoch operations resource and one of each capability and they should never move again after the upcoming change, so I think it makes sense to re-use the path

jordanschalm: I agree that on Mainnet the paths won't conflict. I was also thinking about making sure we have a functional deployment for new instances of the contract (for example for integration tests, or new networks) in https://github.com/onflow/flow-core-contracts/pull/367#issuecomment-1533313574. 

I feel like the simplest way to accomplish this is to store both the admin resources and capabilities to the same account, by default, which would require specifying distinct path names for the new capabilities.

contracts/FlowIDTableStaking.cdc
jordanschalm: I'm actually happy to include them, you're right that they'll likely be used eventually. 👍

joshuahannan:
@jordanschalm I updated the contract to use different paths for the capabilities so it is ready for another review now.

@janezpodhostnik Do you think you could do a second review on this?

jordanschalm (approved):

janezpodhostnik (commented):
Sorry for the late review. This look ok to me.

PR #370, opened 2023-06-27
Adds metadata views to `FlowToken` and removes init parameter

joshuahannan:
Adds standard fungible token metadata views to the `FlowToken` smart contract
Also removes the init parameter from the contract so that it is easier for users to deploy it to the playground


contracts/FlowToken.cdc
```diff
@@ -1,4 +1,5 @@
 import FungibleToken from 0xFUNGIBLETOKENADDRESS
+import FungibleTokenMetadataViews from "./FungibleTokenMetadataViews.cdc"
 
 pub contract FlowToken: FungibleToken {
```
sisyphusSmiling, line 4: I think the contract should implement `ViewResolver` if possible. Implies the need to add the contract interface methods to this contract.

contracts/FlowToken.cdc
```diff
@@ -83,6 +84,64 @@ pub contract FlowToken: FungibleToken {
                 FlowToken.totalSupply = FlowToken.totalSupply - self.balance
             }
         }
+
+        /// Get all the Metadata Views implemented by FlowToken
+        ///
+        /// @return An array of Types defining the implemented views. This value will be used by
+        ///         developers to know which parameter to pass to the resolveView() method.
+        ///
+        pub fun getViews(): [Type]{
```
sisyphusSmiling, line 93: Between `getViews()` and `resolveView()`, don't we want `Vault` to implement `MetadataViews.Resolver`?

contracts/FlowToken.cdc
```diff
@@ -167,32 +226,32 @@ pub contract FlowToken: FungibleToken {
         }
     }
 
-    init(adminAccount: AuthAccount) {
+    init() {
```
sisyphusSmiling, line 229: I'm thinking we'll want to coordinate this change with DevEx. I'm concerned there could be a large blast radius if we make this change to the contract interface without plans for a fast follow release on dev tooling. It'd be helpful to get the scope of impact and timeline so we can communicate this planned update to the community.

contracts/FlowToken.cdc
joshuahannan: great point! I'll add that

joshuahannan: ope, nevemind, I just remembered that we can't add new interface conformances in an upgrade, so that won't be possible. I can still add the methods to the contract though

contracts/FlowToken.cdc
joshuahannan: We can't add interface conformances in an upgrade, so we can't do that here unfortunately

sisyphusSmiling: [Docs](https://developers.flow.com/cadence/language/contract-updatability#structs-resources-and-interfaces) mention adding an interface conformance is a valid upgrade. Is the issue here adding a new conformance?

contracts/FlowToken.cdc
```diff
@@ -83,6 +84,64 @@ pub contract FlowToken: FungibleToken {
...
+            switch view {
+                case Type<FungibleTokenMetadataViews.FTView>():
+                    return FungibleTokenMetadataViews.FTView(
+                        ftDisplay: self.resolveView(Type<FungibleTokenMetadataViews.FTDisplay>()) as! FungibleTokenMetadataViews.FTDisplay?,
+                        ftVaultData: self.resolveView(Type<FungibleTokenMetadataViews.FTVaultData>()) as! FungibleTokenMetadataViews.FTVaultData?
+                    )
+                case Type<FungibleTokenMetadataViews.FTDisplay>():
+                    let media = MetadataViews.Media(
+                            file: MetadataViews.HTTPFile(
+                            url: "https://assets.website-files.com/5f6294c0c7a8cdd643b1c820/5f6294c0c7a8cda55cb1c936_Flow_Wordmark.svg"
```
sisyphusSmiling, line 114: Do we know where this is hosted? It'd be great to get this pinned in IPFS or some other Web3 native solution.

joshuahannan:
Tests are passing, but I'm having a hard time getting CI to pass because there are diffs in `lib/go/test/go.sum` that I don't know how to resolve properly. I have to get to some other things, so I will come back to this in a couple days

satyamakgec (approved):

contracts/FlowToken.cdc
```diff
@@ -167,32 +241,32 @@ pub contract FlowToken: FungibleToken {
...
         let vault <- create Vault(balance: self.totalSupply)
-        adminAccount.save(<-vault, to: /storage/flowTokenVault)
+        self.account.save(<-vault, to: /storage/flowTokenVault)
 
         // Create a public capability to the stored Vault that only exposes
         // the `deposit` method through the `Receiver` interface
         //
-        adminAccount.link<&FlowToken.Vault{FungibleToken.Receiver}>(
+        self.account.link<&FlowToken.Vault{FungibleToken.Receiver, FungibleToken.Balance, MetadataViews.Resolver}>(
             /public/flowTokenReceiver,
```
austinkline, line 256: Are you able to resolve the FTVaultData metadata view and use its path values here and in other places. That way you can guarantee they're the same, and anyone who pokes around on the flow token contract to find paths will see an example of how to resolve that metadata

contracts/FlowToken.cdc
```diff
@@ -1,6 +1,9 @@
-import FungibleToken from 0xFUNGIBLETOKENADDRESS
+import FungibleToken from "FungibleToken"
+import MetadataViews from "MetadataViews"
+import FungibleTokenMetadataViews from "FungibleTokenMetadataViews"
+import ViewResolver from "ViewResolver"
```
austinkline, line 4: This applies elsewhere as well, but is there an issue with using quotes-only imports? I'm guessing it comes down to sdk support?

contracts/FlowToken.cdc
```diff
@@ -83,6 +86,24 @@ pub contract FlowToken: FungibleToken {
...
+        }
+
+        /// Get a Metadata View from FlowToken
+        ///
+        /// @param view: The Type of the desired view.
+        /// @return A structure representing the requested view.
+        ///
+        pub fun resolveView(_ view: Type): AnyStruct? {
+            return FlowToken.resolveView(view)
+        }
```
bluesign, line 106: I feel like `balance` should be somewhere in some view, and we should resolve that view here.

contracts/FlowToken.cdc
```diff
@@ -96,6 +117,59 @@ pub contract FlowToken: FungibleToken {
...
+
+    /// Get a Metadata View from FlowToken
+    ///
+    /// @param view: The Type of the desired view.
+    /// @return A structure representing the requested view.
+    ///
+    pub fun resolveView(_ view: Type): AnyStruct? {
+        switch view {
+            case Type<FungibleTokenMetadataViews.FTView>():
+                return FungibleTokenMetadataViews.FTView(
```
bluesign, line 134: Maybe some view to `totalSupply ` can also be added

contracts/FlowToken.cdc
joshuahannan: Sorry, I'm a little confused about what you're suggesting. You're just saying that we should resolve the view in the `init` method just to show an example of how it is done on a common contract? Not anything practical?

I think that is probably a good idea, so I can include that

contracts/FlowToken.cdc
joshuahannan: We might just store it on chain in xml format since it is small

contracts/FlowToken.cdc
joshuahannan: I want to use quotes only imports, but I didn't want to do all of it in this PR, so I'll be doing the rest of it in a future PR

contracts/FlowToken.cdc
joshuahannan: `balance` is already accessible directly on the `Vault`, so there is no need to put it into a view, right?

contracts/FlowToken.cdc
joshuahannan: I'll make a PR to add that to `FungibleTokenMetadataViews` then add that here

joshuahannan: https://github.com/onflow/flow-ft/pull/139

contracts/FlowToken.cdc
bluesign: yeah you are right

contracts/FlowToken.cdc
joshuahannan: Anybody have any ideas about which metadata view would be the best to use for storing an entire XML file? This one would only be about 2KB so we can probably just store it all on chain, but I don't know which one to use or if we have to create a new view for it

bluesign: @joshuahannan `HTTPFile` with `data` URI ?

joshuahannan: So I would just save the entire XML file as a `String` in the URI field?

austinkline: So long as you encode the start with the right encoding that should work, I think. Only returning the string won't tell platforms how to interpret the data being returned

bluesign: it will be something like: 

```data:text/xml;base64,iVBORw0KGgoAAA
ANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4
//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU
5ErkJggg==```

PR #379, opened 2023-08-09
Epoch phase performance improvements, access node open slot limits, and `EpochStart` event

joshuahannan:
Makes three changes to improve the performance of the epoch transition:

1. Moves `FlowEpoch.calculateAndSetRewards` to the end of the epoch setup phase/beginning of epoch commit, which currently does not have much complexity. This removes calculating rewards from the end of the epoch, which when combined with `moveTokens`, was causing the execution to take a long time.

2. Changes `FlowIDTableStaking.setNewMovesPending()` to be able to add a new node or delegator to the moves pending list without loading and storing to and from storage. If many nodes and delegators were unstaking tokens, this was very costly at the end of each epoch. The change uses a reference to allow the moves pending list to be modified in memory and not loaded or stored during node registration transactions.

3. Updates the methods that are run at the end of the service auction to construct the proposed Node ID list as they go instead of the epoch contract having to query it fully at the end, which requires iterating through all nodes in the identity table to construct it.

Adds some new features to the epoch and staking contracts

5. Adds a `FlowEpoch.EpochStart` event that is emitted at the start of every epoch and includes important information related to `FlowToken` and epoch information.

6. Adds a feature to open up new access node slots each epoch instead of the service account committee having to do it manually

Also updates some comments and error messages for some transactions

No tests are needed for the first two improvements because they don't affect the behavior of the contracts, just the performance


contracts/FlowIDTableStaking.cdc
```diff
@@ -1035,11 +1038,13 @@ pub contract FlowIDTableStaking {
...
                     delRecord.tokensRequestedToUnstake = delRecord.tokensStaked.balance
-                    FlowIDTableStaking.setNewMovesPending(nodeID: nodeRecord.id, delegatorID: delegator)
+                    movesPendingList = FlowIDTableStaking.setNewMovesPending(nodeID: nodeRecord.id, delegatorID: delegator, existingList: movesPendingList)
                 }
             }
 
-            FlowIDTableStaking.setNewMovesPending(nodeID: nodeRecord.id, delegatorID: nil)
+            movesPendingList = FlowIDTableStaking.setNewMovesPending(nodeID: nodeRecord.id, delegatorID: nil, existingList: movesPendingList)
+            // Save the updated moves pending list
+            FlowIDTableStaking.account.save<{String: {UInt32: Bool}}>(movesPendingList, to: /storage/idTableMovesPendingList)
```
janezpodhostnik, line 1047: It might be more readable if load and save of idTableMovesPendingList are packaged into functions

contracts/FlowIDTableStaking.cdc
```diff
@@ -1636,12 +1667,14 @@ pub contract FlowIDTableStaking {
...
+    access(contract) fun setNewMovesPending(nodeID: String, delegatorID: UInt32?, existingList: {String: {UInt32: Bool}}?): {String: {UInt32: Bool}} {
         if self.nodes[nodeID] == nil {
-            return
+            return existingList ?? {}
         }
-        let movesPendingList = self.account.load<{String: {UInt32: Bool}}>(from: /storage/idTableMovesPendingList)
-            ?? panic("No moves pending list in account storage")
+
+        let movesPendingList = existingList ?? (self.account.load<{String: {UInt32: Bool}}>(from: /storage/idTableMovesPendingList)
+            ?? panic("No moves pending list in account storage"))
```
janezpodhostnik, line 1677: are the outer parenthesis needed?

contracts/FlowIDTableStaking.cdc
```diff
@@ -1125,11 +1142,16 @@ pub contract FlowIDTableStaking {
                 //  - New ANs may be unapproved, but must have submitted sufficient stake
                 if nodeRecord.role == UInt8(5) && !greaterThanMin && !nodeIsApproved {
                     self.removeAndRefundNodeRecord(nodeID)
+                    if participantList[nodeRecord.id] != nil {
```
janezpodhostnik, line 1145: is this check needed?
why not use remove instead of setting to nil?

janezpodhostnik (commented):
Can you explain in more detail why we can do this? I'm not sure I understand.

If the rewards are calculated earlier is there not a chance they would change?

contracts/FlowIDTableStaking.cdc
joshuahannan: I don't know if it is really worth adding utility functions that are only a single line. The contract is large enough as it is. And we already do that most of the time we call it, like in the node and delegator objects. Here it is separate to reduce the number of storage operations that we have to do

contracts/FlowIDTableStaking.cdc
joshuahannan: Probably not, but it felt better to put them there just to be clear

contracts/FlowIDTableStaking.cdc
joshuahannan: Good idea. I'll make the change

joshuahannan:
I'm moving the rewards calculation to earlier so that it doesn't have to happen in the same transaction where the epoch transitions and tokens are moved. We currently don't have any logic to change rewards once they have been calculated anyway, so it doesn't really matter if they are calculated earlier.

joshuahannan:
@jordanschalm No rush, but I could use another review on this when you get the chance. We could also schedule a call so I can walk you through the changes if you need help

contracts/FlowIDTableStaking.cdc
```diff
@@ -1196,10 +1206,22 @@ pub contract FlowIDTableStaking {
                     // No action is needed to mark the nodes as added because they are already included
                     currentNodeCount[role] = currentNodeCount[role]! + UInt16(candidateNodesForRole.keys.length)
                 }
+
+                nodesToAdd.appendAll(candidateNodesForRole.keys)
+
+                // Add the desired open slots for each role to the current node count
+                // to make the slot limits
+                if let openSlotsForRole = openSlots[role] {
+                    slotLimits[role] = currentNodeCount[role]! + openSlotsForRole
```
jordanschalm, line 1215: We could _lower_ the slot limit for a role here, which I don't think is the desired behaviour. 

For example, suppose `slotLimits[r]=100`, `currentNodeCount[r]=50`, and `openSlots[role]=10`. Then we would end up setting `slotLimits[r]=60`.

contracts/epochs/FlowEpoch.cdc
```diff
@@ -420,13 +448,13 @@ pub contract FlowEpoch {
                     }
                 case EpochPhase.EPOCHSETUP:
                     if FlowClusterQC.votingCompleted() && (FlowDKG.dkgCompleted() != nil) {
+                        self.calculateAndSetRewards()
```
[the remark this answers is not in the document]
joshuahannan: `calculateAndSetRewards` takes up the majority of the execution time because it currently has to iterate through every delegator for all the staked node operators, even if they aren't staked, because it still has to check if their stake is non-zero. So it will move it to a different place, but it at least separates it into a different transaction from the other end epoch operations to make it not quite as large since it used to be a single transaction. And then we can ideally work on a more fundamental solution in the meantime.

I'm not sure if there is a clean way to improve it fundamentally though because currently, the delegator information is stored in the `NodeRecord` resource and we can add new fields to it to track only active delegators for example. So we'd have to either do some sort of migration to delete any delegator IDs that currently don't have any stake or we'd have to make another stored dictionary separate from `NodeRecord` that tracks all delegators,  which feels kind of dirty. That is probably out of this scope of this upgrade though.

lib/go/test/epoch_test_helpers.go
```diff
@@ -528,6 +570,47 @@ func verifyConfigMetadata(
...
+
+	var i uint64
+	i = 0
+	for i < 1000 {
+		results, _ := adapter.GetEventsForHeightRange(context.Background(), "A."+epochAddress.String()+".FlowEpoch.EpochStart", i, i)
+
+		for _, result := range results {
+			for _, event := range result.Events {
+				if event.Type == "A."+epochAddress.String()+".FlowEpoch.EpochStart" {
+					emittedEvent = EpochStartEvent(event)
```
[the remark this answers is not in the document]
joshuahannan: good suggestion!

lib/go/test/epoch_test_helpers.go
```diff
@@ -528,6 +570,47 @@ func verifyConfigMetadata(
...
+				}
+			}
+		}
+
+		i = i + 1
+	}
```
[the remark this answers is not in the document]
joshuahannan: yeah, that should probably work. I'll test it out

contracts/FlowIDTableStaking.cdc
```diff
@@ -1636,12 +1660,14 @@ pub contract FlowIDTableStaking {
 
     /// Adds a node and/or a delegator to the list of node IDs who have pending token movements
     /// or whose delegators have pending movements
-    access(contract) fun setNewMovesPending(nodeID: String, delegatorID: UInt32?) {
+    /// Can pass an existing list to add it to that instead of the saved one, like during `moveTokens`
+    access(contract) fun setNewMovesPending(nodeID: String, delegatorID: UInt32?, existingList: {String: {UInt32: Bool}}?): {String: {UInt32: Bool}} {
```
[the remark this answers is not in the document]
joshuahannan: yeah, I think that is a great suggestion. I'll work on that change

lib/go/test/flow_idtable_staking_test.go
```diff
@@ -109,20 +109,20 @@ func TestIDTableDeployment(t *testing.T) {
 
 		// Check that the reward ratios were initialized correctly for each node role
 
-		result = executeScriptAndCheck(t, b, templates.GenerateGetRewardRatioScript(env), [][]byte{jsoncdc.MustEncode(cadence.UInt8(1))})
-		assertEqual(t, CadenceUFix64("0.168"), result)
+		// result = executeScriptAndCheck(t, b, templates.GenerateGetRewardRatioScript(env), [][]byte{jsoncdc.MustEncode(cadence.UInt8(1))})
```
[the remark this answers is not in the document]
joshuahannan: not sure why I commented those out. I'll make sure they work properly

contracts/FlowIDTableStaking.cdc
joshuahannan: good catch. So we could just check to see if the calculated slot slot limits is less than the current slot limits and only set it if it is greater, but I could potentially see this behavior being desirable if we decide we want less slot limits than what we currently have

contracts/epochs/FlowEpoch.cdc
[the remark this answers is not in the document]
jordanschalm: Agreed it's out of scope for this PR. I was just curious.

contracts/FlowIDTableStaking.cdc
jordanschalm: > I could potentially see this behavior being desirable if we decide we want less slot limits than what we currently have

Yeah, I'm not completely sure which behaviour we want. If we allow the `openSlots` process to _remove_ slots though, I feel we should document it in the parts of code dealing with `slotLimits` (eg. [here](https://github.com/onflow/flow-core-contracts/blob/access-slots/transactions/idTableStaking/admin/set_slot_limits.cdc#L3-L8))

joshuahannan: documented in the latest commit

contracts/FlowIDTableStaking.cdc
```diff
@@ -1634,14 +1657,21 @@ pub contract FlowIDTableStaking {
...
-    /// or whose delegators have pending movements
-    access(contract) fun setNewMovesPending(nodeID: String, delegatorID: UInt32?) {
+    /// Adds a node and/or delegator to the moves pending list
+    /// and returns the list
+    /// If `existingList` is `nil`, this loads the current list from storage and modifies it
+    /// If `existingList` is non-`nil`, this modifies the given list instead of loading from storage
+    access(contract) fun modifyNewMovesPending(nodeID: String,
+                                               delegatorID: UInt32?,
+                                               existingList: {String: {UInt32: Bool}}?)
+                                               : {String: {UInt32: Bool}}
```
turbolent, line 1667: Passing a dictionary and and returning a dictionary is very expensive: For each call of the function, the argument for the parameter is copied, modified in the function, and then copied once more.

If the dictionary should be modified, make the parameter a reference. Further, don't load (copy and remove) the input from storage (line below), but simply borrow from storage.

joshuahannan: yep, I've already made this change locally. Just haven't pushed it yet

joshuahannan:
@jordanschalm Can you take a look at my last two commits? I've added in all the performance improvements we discussed this week

contracts/FlowIDTableStaking.cdc
```diff
@@ -921,23 +913,31 @@ pub contract FlowIDTableStaking {
             FlowIDTableStaking.account.save(slotLimits, to: /storage/flowStakingSlotLimits)
         }
 
+        /// Sets the number of open node slots to allow per epoch
+        /// Only access nodes are used for this currently,
+        /// but other node types will be added in the future
+        pub fun setOpenNodeSlots(openSlots: {UInt8: UInt16}) {
+            pre {
+                openSlots[UInt8(5)] != nil: "Need to have a value set for access nodes"
```
turbolent, line 921: No need to convert `Int` to `UInt8`, just use a `UInt8` value directly:
```suggestion
                openSlots[5] != nil: "Need to have a value set for access nodes"
```

turbolent, line 921: What is the magic number 5 here?

contracts/FlowIDTableStaking.cdc
```diff
@@ -1405,8 +1424,10 @@ pub contract FlowIDTableStaking {
                 ?? panic("No moves pending list in account storage")
 
             // Reset the movesPendingList
-            let movesPendingList: {String: {UInt32: Bool}} = {}
-            FlowIDTableStaking.account.save<{String: {UInt32: Bool}}>(movesPendingList, to: /storage/idTableMovesPendingList)
+            var emptyMovesPendingList: {String: {UInt32: Bool}} = {}
+            FlowIDTableStaking.account.save<{String: {UInt32: Bool}}>(emptyMovesPendingList, to: /storage/idTableMovesPendingList)
```
turbolent, line 1428: No need for explicit type argument, can use type inference. Maybe also remove in the rest of the contract while at it.

```suggestion
            FlowIDTableStaking.account.save(emptyMovesPendingList, to: /storage/idTableMovesPendingList)
```

contracts/FlowIDTableStaking.cdc
joshuahannan: It is the access node role number. We can't add fields in the contract, so we just use the magic number

coverage.json
```diff
@@ -0,0 +1,72 @@
+{
```
jordanschalm, line 1: Did you intend to commit this?

contracts/FlowIDTableStaking.cdc
```diff
@@ -1512,9 +1513,6 @@ pub contract FlowIDTableStaking {
                 nodeRecord.tokensRequestedToUnstake = 0.0
             }
 
-            // Save the updated moves pending list
-            FlowIDTableStaking.storeNewMovesPending(movesPendingList)
```
jordanschalm: We don't need this anymore, because modifications to a borrowed object are always persisted back to storage? Just checking :)

jordanschalm (approved):

coverage.json
joshuahannan: no, I'll remove it and add it to gitignore

contracts/FlowIDTableStaking.cdc
joshuahannan: yes, that is correct

contracts/FlowIDTableStaking.cdc
```diff
@@ -1020,8 +1016,10 @@ pub contract FlowIDTableStaking {
                 FlowIDTableStaking.account.save(currentRoleNodeCounts, to: /storage/flowStakingRoleNodeCounts)
             }
 
+            var movesPendingList = FlowIDTableStaking.account.borrow<&{String: {UInt32: Bool}}>(from: /storage/idTableMovesPendingList)
+                ?? panic("No moves pending list in account storage")
```
turbolent, line 1020: Maybe make this a function and call it in the other places where this borrow occurs. This way the type/path is in one place, the code becomes less brittle

joshuahannan: I only do it twice, so creating a whole other function seems kind of unnecessary

contracts/epochs/FlowEpoch.cdc
```diff
@@ -57,9 +57,36 @@ pub contract FlowEpoch {
         pub case Access
     }
 
+    /// The Epoch Start service event is emitted when the contract transitions
+    /// to a new epoch in the staking auction phase.
+    pub event EpochStart (
```
bjartek, line 62: Will this be emitted along with NewEpoch from FlowIDTableStaking?

bjartek:
I think it is very important that every set of service chunk events that transitions something contains the counter field. Atm NewEpoch and Payout event from FlowIDTableStaking does not contain this. This makes it very hard to identify what epoch this payout was for. In some cases these events happen before NewEpoch and in others it hapends after. If we have to "fix stuff"

contracts/epochs/FlowEpoch.cdc
joshuahannan: yes, it will be emitted in the same transaction

joshuahannan:
@bjartek That would be nice to have, but that might require the staking contract to import the `FlowEpoch` contract, which would be a circular dependency, and we want to avoid that. I guess we could pass the epoch counter as an argument to the `payRewards()` and `moveTokens()` functions so we can use it in the events, but that still feels a little weird. It isn't enough that you can infer the epoch from when the event was emitted? We could potentially include it if it is that important to you

bjartek:
Is it possible to have another event emitted when rewards are paid from FlowEpoch that has the counter?

contracts/FlowIDTableStaking.cdc
turbolent: Saw three occurrences, but also works as-is (there a few other cases for other paths, too). Keeping it DRY is in general a good idea to avoid future problems

jordanschalm:
> I guess we could pass the epoch counter as an argument to the payRewards() and moveTokens() functions so we can use it in the events

Assuming it's relatively straightforward, I think we should just do this and include the epoch counter in all events where it could be useful. Inferring auxiliary data about events based on the current state when they are emitted, or based on other events, is error-prone and more complex.

joshuahannan:
Okay. I'll add those to the events in the staking contract then

contracts/epochs/FlowEpoch.cdc
```diff
@@ -857,7 +944,14 @@ pub contract FlowEpoch {
                                            numViewsInDKGPhase: numViewsInDKGPhase,
                                            numCollectorClusters: numCollectorClusters,
                                            FLOWsupplyIncreasePercentage: FLOWsupplyIncreasePercentage)
-        
+
+        // Set a reasonable default for the epoch timing config targeting 1 block per second
+        let defaultEpochTimingConfig = EpochTimingConfig(
+            duration: numViewsInEpoch,
+            refCounter: currentEpochCounter,
+            refTimestamp: UInt64(getCurrentBlock().timestamp))
```
jordanschalm, line 952:
```suggestion
            refTimestamp: UInt64(getCurrentBlock().timestamp)+numViewsInEpoch)
```
Default reference timestamp should be at the _end_ of the first epoch. Currently it is at the beginning. This is my mistake from #395

joshuahannan:
I did the upgrade on canary and everything seems to be working properly. We'll upgrade testnet on monday the 27th and stick with the dec 6th mainnet upgrade. I'm going to go ahead and merge this though unless anyone has any objections

PR #380, opened 2023-08-14
Add epoch field to new epoch

bjartek:
currently the NewEpoch event does not have the epoch/counter as an event field. This PR proposes to fix this. Note that I have called the field `epoch` here but we can always change it `counter` to be consistent with EpochSetup/EpochCommit. 

I tried to add to RewardsPaid aswell but i decided to start with this one.


joshuahannan (commented):
I'd almost rather just create a new event in the `FlowEpoch` contract that contains this and other identifying information and metadata about the new epoch instead of having to pass the epoch counter into the staking contract. What do you think? I could make the change

bjartek:
That works. Should we do the same for Rewards? Only having to care about the FlowEpoch contracts to get all relevant events to follow the flow makes sense to me.

joshuahannan:
I think since the staking contract is actually the one paying the rewards, has all the specific data that is required for it, and is also emitted the individual node and delegator rewards events, we can keep the event emission for rewards in there. Is there a specific reason why you would want it in the epoch contract instead?

bjartek:
No readon, keep it where it is best.

bjartek:
I think that works really well.

bjartek:
Looks like was closed since i deleted the fork i had.

joshuahannan:
no worries. I can add it to my PR.

AlexHentschel:
In general, I think the more lifecycle events the Flow platform exposes the better. 

Regarding:
```
/// The total rewards that will be paid out at the end of the current epoch.
totalRewards: UFix64,
```
I would suggest to refine the documentation a bit. Context:
* At the moment, Flow nodes largely receive a fixed reward. For network safety and economical sustainability, the problem of freeloaders needs to be addressed (nodes that stake for a roll in the protocol but don't do the work, e.g. are offline for prolonged period of time, way too slow etc):
   * At the moment, there is a partially-manual governance process in place to potentially reduce the reward for offline nodes. Because of the required manual effort, this process only scales to handling the few really bad nodes.
   * At some point somewhat soon, we will replace this human-based fallback by an algorithmic solution. Essentially, the Flow protocol itself already tracks the contributions of each node (e.g. consensus signatures on a block, verification node's approvals of execution results, etc). Based on this data that the chain already records, the network (specifically system smart contracts) should track the performance of each node. 
     
       This will produce an objective metric for each node. E.g. something like: out of 1432 finalized blocks that Alice should have voted on, she managed to contribute her vote to 1174 blocks' Quorum Certificates, i.e. Alice successfully contributed to network progress in 82% of the cases. Lets call that the 'Contribution Success' metric. 

My intuition is that when we have performance-dependent rewards in the future, the `totalRewards` might not be known at the beginning of the Epoch. I think a definition of `totalRewards` as "The total rewards that will be paid out at the end of the current epoch for the idealized scenario, where all nodes perform perfectly." The reward payed out in reality will likely be a bit smaller, because we statistically expect some nodes to have problems or be slow. 

I want to emphasize that this ☝️ is still in an early conceptual stage. Up to this point, I would say that we have an idea how to solve the problem, but a precise proposal still needs to be worked out. For example, we don't have a precise formula for rewarding nodes based on their Contribution Success metric. In other words, there is little ambiguity how to measure a node's performance from a technical perspective, but we still need to decide how to translate the performance number (e.g. 82%) to reward (number of tokens).

joshuahannan:
Those are all great points, Alex, but didn't we decide that if a node's rewards get withheld, then the amount that is withheld is split up among the other node operators? That is how it works in the smart contract right now. That way, the total reward amount remains the same

jordanschalm:
This looks good to me. Some thoughts:
- I think it is fine to include `totalRewards`. We could adjust the documentation as Alex mentioned or perhaps set the name to be eg. `totalAvailableRewards`
- If there is some user need for more identity table data (sounds like there isn't so far) we could consider a compressed version, like only specifying mode IDs in the event

PR #383, opened 2023-09-14
RandomBeaconHistory impl + transactions & scripts

sisyphusSmiling:
Adding initial historical source of randomness contract implementation.

Submitting PR as-is for impressions & initial feedback before proceeding with more in depth testing on base branch.


transactions/randomBeaconHistory/record_source_of_randomness.cdc
```diff
@@ -0,0 +1,19 @@
+import "RandomBeaconHistory"
+
+/// Commits the source of randomness for the requested block height from the RandomBeaconHistory Heartbeat
+///
+// TODO: Replace transaction parameter with randomSourceHistory() call
+transaction(sor: [UInt8]) {
```
janezpodhostnik, line 6: This transaction is not really needed, only the system transaction will be doing this.

transactions/randomBeaconHistory/scripts/get_source_of_randomness.cdc
```diff
@@ -0,0 +1,15 @@
...
+pub fun main(atBlockHeight: UInt64): [UInt8] {
+    // Get the account storing the heartbeat
+    let heartbeatAccount = getAccount(RandomBeaconHistory.getRandomBeaconHistoryAddress())
+    // Borrow the HeartbeatPublic Capability
+    let randomBeaconHistoryHeartbeat = heartbeatAccount.getCapability<&{RandomBeaconHistory.HeartbeatPublic}>(
+            RandomBeaconHistory.HeartbeatPublicPath
+        ).borrow()
+        ?? panic("Couldn't borrow RandomBeaconHistory.Heartbeat Resource")
+    // Return the source of randomness for the requested block height
+    return randomBeaconHistoryHeartbeat.sourceOfRandomness(atBlockHeight: atBlockHeight)
```
janezpodhostnik, line 14: This looks way to verbose! Can this be packaged into `RandomBeaconHistory.sourceOfRandomness(atBlockHeight: atBlockHeight)`.

contracts/RandomBeaconHistory.cdc
```diff
@@ -0,0 +1,119 @@
...
+            )
+
+            return self.randomSourceHistory[index]
+        }
+
+        /// Callable by owner of the Heartbeat resource, Flow Service Account, records the provided random source
+        ///
+        /// @param randomSourceHistory The random source to record
+        ///
+        pub fun heartbeat(randomSourceHistory: [UInt8]) {
```
janezpodhostnik, line 94: IMO the heartbeat resource should have only one method `pub fun heartbeat(randomSourceHistory: [UInt8])` the rest should be on the contract itself.

contracts/RandomBeaconHistory.cdc
```diff
@@ -0,0 +1,119 @@
...
+            let now = getCurrentBlock().height
+            if self.initHeight == nil {
+                self.initHeight = now
+                emit HearbeatInitialized(atBlockHeight: now)
+            }
+            self.randomSourceHistory.append(randomSourceHistory)
+            emit NewSourceOfRandomness(atBlockHeight: now, randomSource: randomSourceHistory)
```
janezpodhostnik, line 101: I would avoid emitting an event here. It adds no value as this will happen predictably every block.

contracts/RandomBeaconHistory.cdc
```diff
@@ -0,0 +1,119 @@
...
+        }
+    }
+
+    /// Returns the address of this contract
+    ///
+    pub fun getRandomBeaconHistoryAddress(): Address {
+        return self.account.address
+    }
```
janezpodhostnik, line 109: I'm a bit confused what purpose this serves...

contracts/RandomBeaconHistory.cdc
```diff
@@ -0,0 +1,119 @@
...
+
+    /// The Heartbeat resource containing each block's source of randomness in sequence
+    ///
+    pub resource Heartbeat : HeartbeatPublic {
+        /// The height at which the first source of randomness was recorded
+        access(self) var initHeight: UInt64?
+        /// Sequence of random sources recorded by the Heartbeat, stored as an array over a mapping to reduce storage
+        // TODO: Test at scale
+        access(self) let randomSourceHistory: [[UInt8]]
```
janezpodhostnik, line 42: this should be implemented as a limited length cyclic array.

- There should be a max length for it
- When writing a new value you overwrite the oldest value, and you keep the index of the latest block stored somewhere
- We might also need admin functions (on a admin resource) to increase and decrease the max length of the array.

janezpodhostnik (commented):
Left some comments

contracts/RandomBeaconHistory.cdc
sisyphusSmiling: My understanding from @dete's comments [here](https://github.com/onflow/flips/pull/123#issuecomment-1693863438) was that we would store `randomSourceHistory` indefinitely. If that's not the path forward, I can proceed with initial plan to implement a cyclic array.

contracts/RandomBeaconHistory.cdc
sisyphusSmiling: I thought it might be helpful for offchain accounting and in the event someone wanted to gether sor from an indexed sources. @bluesign do you have thoughts on including events in this contract?

transactions/randomBeaconHistory/scripts/get_source_of_randomness.cdc
sisyphusSmiling: Sure, can do. I'll wire up the methods against the contract.

transactions/randomBeaconHistory/record_source_of_randomness.cdc
sisyphusSmiling: Got it, it's included here only for the purpose of testing until I can get tests running using the [system chunk transaction](https://github.com/onflow/flow-go/blob/1231763cb306d6d81cdf7c922c5761097d1dcfb9/fvm/blueprints/scripts/systemChunkTransactionTemplate.cdc#L19).

contracts/RandomBeaconHistory.cdc
bluesign: it can be good to have, I would even think adding SoRA and SoRB together, so we don't need to add it to transaction result etc. @tarakby knows much better, so he can comment here better I guess. 

Also event we can listen with streaming API from execution state, no need to query chain ( constant poll )

contracts/RandomBeaconHistory.cdc
janezpodhostnik: Ah, of course! forgot about that.

bluesign: I think reading single value from this will cost like ~100 computation usage after a year.

tarakby: Yes the FLIP got updated to propose a non-limited history.

contracts/RandomBeaconHistory.cdc
tarakby: As an inner protocol developer, I don't use the network much "from outside" to be able to assess the advantage of listening to events Vs querying the chain. But here are some comments, probably all in favor of not emitting events 😅 

- `SoR-A` and `SoR_B` events would be emitted for every block, and in that sense they are expected side effects of executing every block, and not an event users want to be notified about (joining Janez's idea of predictable events). 
- the overhead of two event by block may be not negligible, If a user doesn't want to query for each block, they can  batch-query for every N blocks instead of every block. 
- If we emit events and ask users to track the values, we may not need to have a history array of sources (SoR_B) in the first place. I think the history stored on-chain allows users to query any past value, without having to constantly follow the blocks and listen to all events. If the chain is storing data for you, why would you want to store the same data off-chain (if you are able to listen to events, then you are able to query the chain too).
- I see the advantage of exposing SoR_A, which we discussed in another PR. Still not sure if the best approach is to include it in transaction results (only those calling `revertibleRandom`) or to emit it as an event. But anyway `SoR_A` isn't currently visible to this contract, so it can't be done in this PR.

sisyphusSmiling: cc @bjartek curious about your thoughts on event emission when source of randomness is committed.

bjartek: I think the events here are ok, there is no more information to add in them really, if we want to have richer event we have to send in metadata that is then emitted, and not sure that is worth it.

sisyphusSmiling: In your opinion, is there a benefit to including the events if the history is easily queryable? @janezpodhostnik correct me if I'm off, but I think the concern is an outsized performance impact (given the execution every block) by their inclusion for little to no benefit. Is there meaningful reason/use case we should expose the data to offchain systems via events?

janezpodhostnik: The performance/size impact overall would be very minimal, but in my opinion so would the information that developers get from it, because not everyone will even be interested in this. 

I also think we can add the events, if there are use cases where the events would be more useful compared to just querying the data.

sisyphusSmiling: Makes sense, @janezpodhostnik. I've updated the contract to remove the events.

contracts/RandomBeaconHistory.cdc
```diff
@@ -0,0 +1,90 @@
+/// RandomBeaconHistory (FLIP 123)
+///
+/// This contract stores the history of random sources generated by the Flow network. The defined Heartbeat resource is
+/// updated by the Flow Service Account at the end of every block with that block's source of randomness.
+///
+/// While the source values are random, generated by the Random Beacon and transmitted into the execution 
+/// environment via the committing transaction, using the raw values from this contract does not guarantee secure
+/// randomness. The Hearbeat is intended to be used in conjunction with a pseudo-random number generator (PRNG)
+/// construct and commit-reveal mechanism to provide an onchain source of secure, non-revertible randomness.
```
tarakby, line 9: IMO the current formulation suggests that using a PRG makes the random source more secure. The PRG is only useful to extend the small random source into a longer random sequence.
I also added minor suggestions, feel free to update the wording.

```suggestion
/// While the source values are safely generated by the Random Beacon (non-predictable, unbiasable, verifiable) and transmitted into the execution 
/// environment via the committing transaction, using the raw values from this contract does not guarantee non-revertible
/// randomness. The Hearbeat is intended to be used in conjunction with a
/// commit-reveal mechanism to provide an onchain source of non-revertible randomness.
// It is also recommended to use the source values with a pseudo-random number 
// generator (PRNG) to generate an arbitrary-long sequence of random values. 
// 
// For usage of randomness where result abortion is not an issue, it is recommended
// to use the Cadence built-in function `revertibleRandom`, which is also based on 
// the safe Random Beacon. 
```

contracts/RandomBeaconHistory.cdc
```diff
@@ -0,0 +1,90 @@
...
+                RandomBeaconHistory.initHeight = currentBlockHeight
+            }
+
+            RandomBeaconHistory.randomSourceHistory.append(randomSourceHistory)
+        }
+    }
+
+    /// Getter for the source of randomness at a given block height. Panics if the requested block height either
+    /// precedes or exceeds the recorded history. Note that a source of randomness for block n will not be accessible
+    /// until block n+1.
```
tarakby, line 46: 👌🏼

tarakby (commented):
Thanks for the clear docs 👏🏼 
It looks good to me, but I let people who are more familiar with Cadence approve.

contracts/RandomBeaconHistory.cdc
sisyphusSmiling: Ah I see, thanks for the clarification!

contracts/RandomBeaconHistory.cdc
```diff
@@ -0,0 +1,96 @@
...
+// For usage of randomness where result abortion is not an issue, it is recommended
+// to use the Cadence built-in function `revertibleRandom`, which is also based on 
+// the safe Random Beacon. 
+///
+/// Read the full FLIP here: https://github.com/onflow/flips/pull/123
+///
+access(all) contract RandomBeaconHistory {
+
+    /// The height at which the first source of randomness was recorded
+    access(contract) var initHeight: UInt64?
```
janezpodhostnik, line 22: In case we decide to change this in the future (to clear some history or something), I feel like a more appropriate name would be `lowestHeight`

contracts/RandomBeaconHistory.cdc
```diff
@@ -0,0 +1,96 @@
...
+        return self.randomSourceHistory[index]
+    }
+
+    /// Getter for the totality of recorded randomness source history
+    ///
+    /// @return An array of random sources, each source an array of UInt8
+    ///
+    access(all) view fun getRandomSourceHistory(): [[UInt8]] {
+        return self.initHeight != nil ? self.randomSourceHistory : panic("History has not yet been initialized")
+    }
```
janezpodhostnik, line 79: If I'm not mistaken returning `self.randomSourceHistory` will copy the underlying array. This will quickly render this function un-callable. 

If we really want something like this we can add bounds to the call:
```cadence
access(all) view fun getRandomSourceHistory(startHeight: Uint64, endHeight: Uint64): [[UInt8]] 
```

or model it as a paged result.
```cadence
getRandomSourceHistoryPage(page: UInt64, perPage: UInt): RandomSourceHistoryPage
//...
struct RandomSourceHistoryPage{
    perPage: UInt
    totalRecords: UInt
    startBlockHeight: UInt
   values: [[UInt8]] 
}
```

bluesign: > If I'm not mistaken returning self.randomSourceHistory will copy the underlying array. This will quickly render this function un-callable.

yeah this will grow really fast. 

btw after stable cadence we can return a reference probably. iirc modifications require entitlements after stable cadence.

sisyphusSmiling: Very true. I think returning non-entitled reference will be most efficient come stable cadence. For now though, I think modeling a paged result would be consistent with other core contracts. Thanks for the callout, will update shortly

janezpodhostnik (approved):

PR #390, opened 2023-10-19
New public function to DKG contract

tarakby:
Add new function to DKG contract to improve debugging and investigating DKG state.


jordanschalm (approved):
Looks good. You'll need to update the static files (make build I believe) to pass CI. 

For Josh, is it possible to read a private field from a contract in a script? I feel like it is, or at least should be, in which case we wouldn't need this.

tarakby:
> For Josh, is it possible to read a private field from a contract in a script? I feel like it is, or at least should be, in which case we wouldn't need this. 

I did try to read the private field but it didn't work (unless I didn't know how to). 
I had to write a script with a similar logic to the contract using public functions only (querying all submissions per node and counting, which is more complicated than exporting the private field).

joshuahannan:
yeah, it still isn't possible to read private fields in scripts

joshuahannan (approved):
I ran the tests and pushed the updates

joshuahannan:
I'm gonna hold off on merging this until we have a plan for the upgrade along with https://github.com/onflow/flow-core-contracts/pull/379. makes sense to do them all at the same time

Issue #394, opened 2023-11-07
[FLIP 204] Update `FlowEpoch` smart contract to specify target end time for epoch


jordanschalm:
## Context

See [FLIP 204](https://github.com/onflow/flips/blob/main/protocol/20231003-service-account-specified-epoch-switchover.md.md) for details.

## Definition of Done

- Add `targetEndTime` field to `EpochSetup` event, `EpochMetadata`
- Add config for determining `targetEndTime` to smart contract `ConfigMetadata`
- Add logic to compute `targetEndTime` to `startEpochSetup`
- Add function for service account to adjust new config
- Testing
    - Validate field is set as expected
    - Validate field is computed correctly
    - Validate setter/getter for new config values

PR #395, opened 2023-11-10
[FLIP 204] Add `TargetEndTime` to `FlowEpoch` smart contract

jordanschalm:
Addresses https://github.com/onflow/flow-core-contracts/issues/394.

This PR adds the new field `TargetEndTime` to the `EpochSetup` event, which will be used by the Cruise Control system to moderate block rate so that epochs end at the desired time. See details in [FLIP 204](https://github.com/onflow/flips/blob/main/protocol/20231003-service-account-specified-epoch-switchover.md.md)

I have not added `targetEndTime` to the `EpochStart` event. Here is why:
- We can't easily persist the target end time, without creating a new ad-hoc storage path. I don't think this is worth the extra complexity. 
- The timing config can change between the `EpochSetup` and `EpochStart` events. If we _don't_ store the value, we may communicate different values in different events for the same epoch.


contracts/epochs/FlowEpoch.cdc
```diff
@@ -253,6 +256,34 @@ pub contract FlowEpoch {
...
+    /// Each epoch is assigned a target end time when it is setup (within the EpochSetup event). 
+    /// The configuration defines a reference epoch counter and timestamp, which defines 
+    /// all future target end times. If `targetEpochCounter` is an upcoming epoch, then
+    /// its target end time is given by:
+    ///
+    ///   targetEndTime = refTimestamp + duration * (targetEpochCounter-refCounter)
+    ///
+    pub struct EpochTimingConfig {
+        /// The duration of each epoch, in seconds
+        pub let duration: UInt64
```
joshuahannan, line 269: Will every epoch always have the same duration? I understand the config can change, but what if the duration changes, but we are using a reference epoch that is before the change? Would we just never do that?

joshuahannan (approved):
Looks good! I'll push back the dates on the upgrades so we can get this in

contracts/epochs/FlowEpoch.cdc
jordanschalm: Well right now, all epochs have the same duration (1 week on Mainnet). If/when we want to change the duration, then we would change the timing config.

The reference epoch needs to contain a real reference counter that has been used in the past. But the associated duration can be anything we want (it does not need to match the duration that the reference epoch actually took, when it occurred). So there isn't anything stopping us from using a reference epoch that is before the duration change, if we want to.

jordanschalm:
> I'll push back the dates on the upgrades so we can get this in

👍 Initially I thought we would need to push a `flow-go` patch before including this change in an upgrade. However, it seems that the relevant service event parsing is already backward-compatible (tested in https://github.com/onflow/flow-go/pull/5013), so we won't need to do this.

We should still make sure to go through an epoch transition on Canary before upgrading Testnet/Mainnet to be safe, but otherwise there's no blockers for including this in the next upgrade.

PR #407, opened 2024-01-23
Add Burner contract to allow resource callbacks before destruction

austinkline:
This contract is meant to help replace the removal of custom destructors from the language. Many use cases exist currently which use custom destructors including:

- Updating the total supply of a Fungible Token or Non-Fungible Token
- Preventing the destruction of a resource which should not be burned yet

While this contract is not something which users or platforms can be forced to be used, it is a safeguard for any developers who want some confidence that they will not accidentally destroy something they didn't intend to.

Two examples of this include:
1. Flowty Loans - Loan resources cannot be destroyed if the NFT they take into escrow is still present. This prevents the loss of collateral for borrowers
2. Lost and Found - Prevent the destruction of a ticket that has resources for a user to redeem

With custom destructors gone, these two contracts can instead migrate to the proposed Burner contract since they are in control of the destruction of their own resources.

Separately, in Fungible Token V2 standards discussions, we do not want to emit a destruction every time a FungibleToken vault is destroyed because every time deposit is called, the depositing vault is drained and _then_ destroyed. If we were to emit events on destruction using the new Resource Destroyed hook, every deposit event would now emit a new useless event. With the Burner contract, FungibleToken can implement `Burnable` and anyone with FungibleToken vaults they wish to destroy can use the new `burn` method to emit destruction events.

**NOTE: It is fully acknowledged that no one can be forced to use this contract. This is not a complete stand-in for custom destructors nor should it be considered as one**

Original PR for this contract can be found here:
https://github.com/green-goo-dao/flow-utils/pull/31


joshuahannan:
Thanks for opening this! Honestly, I would prefer that this goes in the flow-ft repo though since that is what this will be primarily used for. What do you think?

contracts/Burner.cdc
```diff
@@ -0,0 +1,79 @@
...
+    // When Crescendo (Cadence 1.0) is released, custom destructors will be removed from cadece.
+    // Burnable is an interface meant to replace this lost feature, allowing anyone to add a callback
+    // method to ensure they do not destroy something which is not meant to be, or to add logic based on destruction
+    // such as tracking the supply of an NFT Collection
+    //
+    // NOTE: The only way to see benefit from this interface is to call the burnCallback method yourself,
+    // or to always use the burn method in this contract. Anyone who owns a resource can always elect **not**
+    // to destroy a resource this way
+    pub resource interface Burnable {
+        pub fun burnCallback()
```
joshuahannan, line 17:
```suggestion
        access(contract) fun burnCallback()
```

If this method is public and we use it to emit the `FungibleToken.Burned` event, wouldn't that be able to be spoofed or spammed since anyone can call it from their resource? I propose making this `access(contract)` so that only the `burn` function in this contract can call it.

joshuahannan: wait, is that even possible? If I make an `access(contract)` method in an interface, does that mean that only the contract the defines the resource interface implementation can access it, or does it mean the `Burner` contract can access it?

PR #411, opened 2024-02-20
[Cadence 0.x] Add `ProtocolStateVersionUpgrade` service event

jordanschalm:
This PR adds a minimal version of a service event to upgrade the protocol state version (see also https://github.com/onflow/flow-go/pull/5428). 

---

For now, I have added this service event to the `NodeVersionBeacon` contract, which has some benefits and drawbacks:
- Conceptually `ProtocolStateVersionUpgrade` does something similar to the existing Version Beacon, so it makes sense for their logic to be together
- The Version Beacon contract already has a Heartbeat resource hooked up to the system chunk, which is necessary to emit service events
- The existing version beacon code is written in a way that suggests the version it is operating on is global, rather than specifically scoped to the Execution State Machine
  - This, combined with the limitations of upgrading contracts, makes adding `ProtocolStateVersionUpgrade` cleanly difficult


contracts/NodeVersionBeacon.cdc
```diff
@@ -280,6 +285,13 @@ pub contract NodeVersionBeacon {
             
             emit NodeVersionBoundaryFreezePeriodChanged(freezePeriod: newFreezePeriod)
         }
+
+        /// Adds a pending protocol state version upgrade, which will be emitted
+        /// as a service event during the next heartbeat.
+        pub fun setPendingProtocolStateVersionUpgrade(newProtocolVersion: UInt64, activeView: UInt64) {
+            let pendingUpgrade = NodeVersionBeacon.ViewActivatedVersion(version: newProtocolVersion, activeView: activeView)
+            NodeVersionBeacon.storePendingProtocolStateVersionUpgrade(upgrade: pendingUpgrade)
```
janezpodhostnik, line 293: Just to confirm my understanding:

The reason we don't just emit the event here is that the service events are only observed from the service transaction.

jordanschalm: Exactly. The idea was to be as restrictive as possible for what qualifies as a service event, because they are so security-sensitive, at the cost of convenience in this kind of case.

contracts/NodeVersionBeacon.cdc
```diff
@@ -336,6 +350,17 @@ pub contract NodeVersionBeacon {
...
         }
+
+        /// Emit a ProtocolStateVersionUpgrade event, if a pending upgrade is in storage.
+        access(self) fun emitProtocolStateVersionUpgradeEventIfNeeded() {
+            let pendingUpgrade = NodeVersionBeacon.popPendingProtocolStateVersionUpgrade()
+            if pendingUpgrade == nil {
+                return
+            }
+            emit NodeVersionBeacon.ProtocolStateVersionUpgrade(
+                newProtocolVersion: pendingUpgrade!.version,
```
joshuahannan, line 361: It doesn't look like it from looking at the other code, but I just want to confirm that there is no way that this optional can be `nil` here

transactions/nodeVersionBeacon/admin/set_protocol_state_version.cdc
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  execute {
+    self.adminRef.setPendingProtocolStateVersionUpgrade(newProtocolVersion: newProtocolVersion, activeView: activeView)
+  }
+
+//   post {
+//     NodeVersionBeacon.peekPendingProtocolStateVersionUpgrade().version! == newProtocolVersion : "version not updated"
+//     NodeVersionBeacon.peekPendingProtocolStateVersionUpgrade().activeView! == activeView : "active view not updated"
+//   }
```
joshuahannan, line 24: Should these be commented out or removed?

joshuahannan (approved):

contracts/NodeVersionBeacon.cdc
jordanschalm: I'm pretty sure it cannot be nil, because of the `if` statement above:
```cadence
if pendingUpgrade == nil {
    return
}
```

transactions/nodeVersionBeacon/admin/set_protocol_state_version.cdc
jordanschalm, line 24:
```suggestion
```

transactions/nodeVersionBeacon/admin/set_protocol_state_version.cdc
```diff
@@ -0,0 +1,25 @@
+import NodeVersionBeacon from 0xNODEVERSIONBEACONADDRESS
+
+/// Transaction that allows NodeVersionAdmin to specify a new protocol state version.
+/// The new version will become active at view `activeView`.
+/// If `activeView` is in the past, or 0, it will become active immediately.
```
jordanschalm, line 5:
```suggestion
/// The new version will become active at view `activeView` if the service event
/// is processed and applied to the protocol state within a block `B` such that
/// `B.view + ∆ < activeView`, for a protocol-defined safety threshold ∆.
/// Service events not meeting this threshold are discarded.
```

joshuahannan (approved):

joshuahannan:
Are you planning on doing this upgrade before or after the Cadence 1.0 upgrade? If the answer is after, we should make these changes based on the `stable-cadence` branch instead of `master`

jordanschalm:
> Are you planning on doing this upgrade before or after the Cadence 1.0 upgrade?

After Cadence 1.0 -> 👍 I'll do that.

jordanschalm:
Closing in favour of https://github.com/onflow/flow-core-contracts/pull/419. (Keeping the branch open for now as it's pinned by a feature branch.)

PR #413, opened 2024-03-22
Don't copy dictionary in getEpochMetadata

janezpodhostnik:
The `self.account.storage.copy<{UInt64: EpochMetadata}>(` made the computation used by the system transaction do from triple digit to **534k**. and the time taken was also unbelievably high (on a testnet clone).

I changed the copy to borrow, which had the problem of getting a reference (this was also described in the comments of the method) so I added a `copyRef` method to `EpochMetadata` to get a non-ref copy of the referenced struct.

To be precise: there might be other locations that contributed to such high computation, but this is the one I was led to first. After we fix this one we will see if there are any others (hopefully not).


janezpodhostnik:
I am yet to test this. I will report back when I do.

turbolent (approved):

janezpodhostnik:
before: 

`"computation_used":534841,"memory_used":4389129140,"time_spent_in_ms":6564`

after:

`"computation_used":517,"memory_used":17165741,"time_spent_in_ms":40`

joshuahannan (approved):

PR #415, opened 2024-03-25
Simplify getting metadata

turbolent:
The `copy` function is unnecessary, if all it does is return the value as-is (`self`).

Simply return the value (epoch metadata struct) from the function.


turbolent:
Nevermind. The function is needed, because the indexing into the dictionary returns a reference.

Issue #414, opened 2024-03-25
SoR history gaps when heartbeat call fails


tarakby:
### Problem

The current SoR history contract assumes a heartbeat function is called at every block to record the SoR corresponding to that block. This is currently done by a system transaction included in every block. Recently (Feb 2024), an update to core-contracts has called the heartbeat function to fail (this is because of a circular dependency in core-contracts that is outside the scope of this issue). This resulted in the SoR history not being recorded for a range of blocks (till the core-contracts got back to a working state).
The SoR corresponding to a block of height `h` is now only available after the gap length (about 200 blocks), instead of right away in height `h+1`. While there is no security impact of this failure as the SoR committed to in a height `h` is still unpredictable and is only revealed in a future block, the delay in revealing the SoR isn't deal for applications. Moreover, accessing the SoR history on a block index `h`, returns the the SoR generated by the protocol in block `h-200`. 

### Fix

- first option: fix the root cause which is the circular dependency of core-contracts that can break a contract during an update. This is the topic of another issue.
- second option regardless of whether the first option is implemented: update the SoR history contract to be more robust against occasional non calls of the heartbeat function in some blocks. This is the topic of this issue. Updating the contract should also fix the current on-chain state of the SoR array

tarakby:
This is currently being discussed in https://discord.com/channels/613813861610684416/1210247669683986442
cc @bluesign

PR #416, opened 2024-03-27
SoR History update

tarakby:
closes https://github.com/onflow/flow-core-contracts/issues/414

The implementation:
- adds a new `backfiller` resource to track and backfill past empty entries in the array
- adds events when a past gap in detected and backfillled
- adds a minimum SoR length check
- updates and expands tests to cover the gap cases


contracts/RandomBeaconHistory.cdc
```diff
@@ -22,10 +22,58 @@ access(all) contract RandomBeaconHistory {
     access(contract) var lowestHeight: UInt64?
     /// Sequence of random sources recorded by the Heartbeat, stored as an array over a mapping to reduce storage
     access(contract) let randomSourceHistory: [[UInt8]]
+    /// Start index of the first gap in the `randomSourceHistory` array where random sources were not recorded, because
+    /// of a heartbeat failure.
+    /// There may be non contiguous gaps in the history, `gapStartIndex` is the start index of the lowest-height
+    /// gap.
+    /// If no gap exists, `gapStartIndex` is equal to the `randomSourceHistory` array length.
+    access(contract) var gapStartIndex: UInt64
```
tarakby, line 30: As mentioned in the Discord [discussion](https://discord.com/channels/613813861610684416/1210247669683986442/1222263812359716944), the contract can't be updated by adding a new field, so an alternative solution needs to be figured out for this field. 
Note that this field is added because we limit backfilling to a maximum number of entries per transaction (to limit the total computation in the transaction/block). If we allow backfilling to go through all missing entries without a limit, the logic won't need to store this field. 

This means we have 2 options:
 - update the logic to allow unlimited backfilling and remove this field. Note that backfilling is only needed in the very rare case where the heartbeat transaction fails, and this won't always result in a large computation block.
 - keep the current logic and decide where to save the new field. The opinion of someone more familiar with Cadence is appreciated in this case. Should another contract be added to store this field?

sisyphusSmiling: WRT dealing with the limitation on adding fields, I think we could deal with this a couple of ways.

1. Add a contract `ServiceAccountReservedFields` which contains mappings of `{String: AnyStruct}` and `@{String: AnyResource}` and add values to those mappings across any service account contracts that may need it.
2. Define a new struct `GapStartIndex` with a single optional variable `value: UInt64` and save that in the service account at some storage path (`/storage/randomBeaconHistoryGapStartIndex`) from which the contract could access and update where needed.
3. Define a new resource `Backfiller` and offload the entirety of the gap state tracking and filling task to that resource. We'd move the `gapStartIndex` field to that resource along with the backfilling logic, all of which are [legal updates](https://cadence-lang.org/docs/1.0/language/contract-updatability#structs-resources-and-interfaces). During every SoR commitment, we would then borrow a reference to the resource and call `backfill()` which would execute the backfill logic if its required at that moment.

IMO 3. is cleaner and is easier to reason about looking at the contract

sisyphusSmiling: Opened #417 against this branch with the changes suggested in 3 for feedback

contracts/RandomBeaconHistory.cdc
```diff
@@ -26,6 +26,19 @@ access(all) contract RandomBeaconHistory {
...
+    // Event emitted when missing SoRs from past heartbeats are detected and will be backfilled:
+    //  - `blockHeight` is the height where the gap is detected
+    //  - `gapStartHeight` is the height of the first missing entry detected
+    access(all) event RandomHistoryMissing(blockHeight: UInt64, gapStartHeight: UInt64)
+
+    // Event emitted when missing SoRs are backfilled on the current heartbeat:
+    //  - `blockHeight` is the height where the backfill happened, it also defines the SoR used to backfill
+    //  - `gapStartHeight` is the height of the first backfilled entry
+    //  - `count` is the number of backfilled entries
+    // Note that in very rare cases, the backfilled gap may not be contiguous. This event does not
```
tarakby, line 38: The reason for adding 2 events and not only `RandomHistoryBackfilled` is the edge case specifically mentioned in the comment. 
If that edge case happens while only  emitting `RandomHistoryBackfilled`, there is no clear way to know what exact indices got backfilled.

tarakby: I copied the suggested event definition from https://github.com/bluesign/flow-core-contracts/blob/randomBackfill/contracts/RandomBeaconHistory.cdc

joshuahannan:
Are you planning on doing this upgrade before or after Cadence 1.0? If the answer is after, I would prefer that you make these changes based on the `stable-cadence` branch instead of `master`

tarakby:
> Are you planning on doing this upgrade before or after Cadence 1.0?

I don't know exactly when Cadence 1.0 will be on MN. This PR change should go as soon as the the PR is merged to fix a current on-chain issue, so I guess it's gonna be before Cadence 1.0

sisyphusSmiling (commented):
Left a couple of minor comments and one question about the backfilled value guarantees

contracts/RandomBeaconHistory.cdc
```diff
@@ -36,14 +49,127 @@ access(all) contract RandomBeaconHistory {
...
+            // optional optimization for the happy common path: if no gaps are detected, 
+            // backfilling isn't needed, return early
+            if self.gapStartIndex + 1 == arrayLength {
+                self.gapStartIndex = arrayLength
+                return
+            }
+
+            // backFilling is limited to a max entries only to limit the computation cost.
+            // This means a large gap may need a few transactions to get fully backfilled.
+            let maxEntries = 100
```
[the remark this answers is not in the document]
tarakby: https://github.com/onflow/flow-core-contracts/pull/416/commits/c5b9d92a2fc2e3a4a1bb2de8b33cd3347b3fd990

Please have a careful look at this one since I'm very new to Cadence. In particular, I made `borrowBackfiller` public (not sure if that's fine). I also remember the heartbeat function is only callable from the service account, the setter of the backfiller limit should also be callable only by the service account. Is this still the case?

contracts/RandomBeaconHistory.cdc
```diff
@@ -36,14 +49,127 @@ access(all) contract RandomBeaconHistory {
...
+                while index < arrayLength && RandomBeaconHistory.randomSourceHistory[index] != [] {
+                    index = index + 1
+                }
+                // if we reach the end of the array then all existing gaps got filled
+                if index == arrayLength {
+                    break
+                }
+                // back fill the empty entry
+                newEntry = HashAlgorithm.SHA3_256.hash(newEntry)
+                RandomBeaconHistory.randomSourceHistory[index] = newEntry
```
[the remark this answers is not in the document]
tarakby: good point. It is guaranteed in this case, I added a comment https://github.com/onflow/flow-core-contracts/pull/416/commits/a9c658ea2d38811def7d65464b9efba690805990

contracts/RandomBeaconHistory.cdc
```diff
@@ -36,14 +49,127 @@ access(all) contract RandomBeaconHistory {
...
+                RandomBeaconHistory.account.save(<-create Backfiller(), to: /storage/randomBeaconHistoryBackfiller)
+            }
+            let backfiller = RandomBeaconHistory.borrowBackfiller() ?? panic("Problem borrowing backfiller")
+
+            // next index to fill with the new random source
+            // so that evetually randomSourceHistory[nextIndex] = inputRandom
+            let nextIndex = currentBlockHeight - RandomBeaconHistory.lowestHeight!
+
+            // if a gap is detected, emit an event
+            if nextIndex > UInt64(RandomBeaconHistory.randomSourceHistory.length) {
```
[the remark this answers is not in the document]
tarakby: good point, I updated the code here: https://github.com/onflow/flow-core-contracts/pull/416/commits/8b78cad193d719375f0efa590392ddcbc1e2aeb3

contracts/RandomBeaconHistory.cdc
```diff
@@ -36,14 +49,127 @@ access(all) contract RandomBeaconHistory {
...
+                let gapStartHeight = RandomBeaconHistory.lowestHeight! + UInt64(RandomBeaconHistory.randomSourceHistory.length)
+                emit RandomHistoryMissing(blockHeight: currentBlockHeight, gapStartHeight: gapStartHeight)
+            }
 
+            // If a new gap is detected in the current transaction, fill the gap with empty entries.
+            // This happens in the rare case where a new gap occurs because of a system transaction failure.
+            while nextIndex > UInt64(RandomBeaconHistory.randomSourceHistory.length) {
+                RandomBeaconHistory.randomSourceHistory.append([])
+            }
```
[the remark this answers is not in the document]
tarakby: You are right, there used to be code updating `gapStartIndex` outside of the `backfiller`, and we had to use a setter and getter of `gapStartIndex`. I already refactored that logic in https://github.com/onflow/flow-core-contracts/pull/416/commits/e2094e790cf5a0cb4e89275aa54d83e8f80ca958, so all touches to `gapStartIndex` now happen within the `backfiller` only.

I'll see if it makes sense to encapsulate the line you commented inside the `backfiller` too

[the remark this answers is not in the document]
tarakby: Ok I ended up moving that line (append the empty entries) within the `backfiller` too : [bf7d1da](https://github.com/onflow/flow-core-contracts/pull/416/commits/bf7d1da01aa639589873e6d6e9f259dd73d3f36a)

contracts/RandomBeaconHistory.cdc
```diff
@@ -158,6 +325,11 @@ access(all) contract RandomBeaconHistory {
         return self.lowestHeight ?? panic("History has not yet been initialized")
     }
 
+    /// Getter for the contract's Backfiller resource
+    access(all) fun borrowBackfiller(): &Backfiller? {
```
sisyphusSmiling, line 329: We need to update access control here since we added an `access(all)` setter for `maxEntriesPerCall`. I'll push changes here in a minute

tarakby: As mentioned in my previous [comment](https://github.com/onflow/flow-core-contracts/pull/416#discussion_r1552561873), I wanted to confirm that `setMaxEntriesPerCall` can only be called by the service account ?

sisyphusSmiling: Sorry to have missed that comment. As far as I know, access to `heartbeat()` is mediated by the contract and storage pattern of the `Heartbeat` resource, unless there's some special casing in FVM I'm not aware of. In this case, we'll also want to restrict public access on `Backfiller` due to the `access(all) fun setMaxEntriesPerCall` declaration. The `access(contract)` restriction I made in [a11baea](https://github.com/onflow/flow-core-contracts/pull/416/commits/a11baea87e7fc56232783d5a00100e96a19397e0) should fit the need here.

tests/test_random_beacon_history.cdc
```diff
@@ -69,6 +70,15 @@ fun testRecordRandomSource() {
         admin
     )
     Test.expect(txResult, Test.beSucceeded())
+    // get backfiller max entries
```
tarakby, line 73: The backfiller max entries value has a dedicated test `testBackfillerMaxEntryPerCall`. Does this case check a specific behaviour non tested in `testBackfillerMaxEntryPerCall` ?

contracts/RandomBeaconHistory.cdc
```diff
@@ -36,25 +49,167 @@ access(all) contract RandomBeaconHistory {
...
+                RandomBeaconHistory.randomSourceHistory.append([])
+                arrayLength = arrayLength + 1
+            }
+
+            var newEntry = randomSource
+            var index = self.gapStartIndex
+            var count = 0 as UInt64
+            while count < self.maxEntriesPerCall {
+                // move to the next empty entry
+                while index < arrayLength && RandomBeaconHistory.randomSourceHistory[index] != [] {
```
turbolent, line 169: Here and elsewhere: Use a cheap `.length > 0` to check for non-emptiness, instead of creating a whole new empty array, each time:

```suggestion
                while index < arrayLength && RandomBeaconHistory.randomSourceHistory[index].length > 0 {
```

contracts/RandomBeaconHistory.cdc
```diff
@@ -36,25 +49,167 @@ access(all) contract RandomBeaconHistory {
...
+
+        access(all) view fun getMaxEntriesPerCall() : UInt64 {
+            return self.maxEntriesPerCall
+        }
+
+        access(all) fun setMaxEntriesPerCall(max: UInt64) {
+            assert (
+                max > 0,
+                message : "the maximum entry per call must be strictly positive"
+            )
```
turbolent, line 113:
```suggestion
            assert(
                max > 0,
                message: "the maximum entry per call must be strictly positive"
            )
```

contracts/RandomBeaconHistory.cdc
```diff
@@ -36,25 +49,167 @@ access(all) contract RandomBeaconHistory {
...
+
+            let currentBlockHeight = getCurrentBlock().height
+            // correct index to fill with the new random source
+            // so that eventually randomSourceHistory[correctIndex] = inputRandom
+            let correctIndex = currentBlockHeight - RandomBeaconHistory.lowestHeight!
+
+            // if a new gap is detected, emit an event
+            var arrayLength = UInt64(RandomBeaconHistory.randomSourceHistory.length)
+            if correctIndex > arrayLength {
+                let gapStartHeight = RandomBeaconHistory.lowestHeight! + arrayLength
```
turbolent, line 130: Maybe extract the repeated `RandomBeaconHistory.lowestHeight!` into a variable

contracts/RandomBeaconHistory.cdc
```diff
@@ -36,25 +49,167 @@ access(all) contract RandomBeaconHistory {
...
+                newEntry = HashAlgorithm.SHA3_256.hash(newEntry)
+                RandomBeaconHistory.randomSourceHistory[index] = newEntry
+                index = index + 1
+                count = count + 1
+            }
+
+            // emit an event about backfilled entries
+            if count > 0 {
+                let gapStartHeight = RandomBeaconHistory.lowestHeight! + self.gapStartIndex
+                emit RandomHistoryBackfilled(blockHeight: getCurrentBlock().height, gapStartHeight: gapStartHeight, count: count)
```
turbolent, line 188: Some line breaks make the expression more readable:

```suggestion
                emit RandomHistoryBackfilled(
                    blockHeight: getCurrentBlock().height, 
                    gapStartHeight: gapStartHeight, 
                    count: count
                )
```

turbolent (commented):
Nice!

contracts/RandomBeaconHistory.cdc
```diff
@@ -143,30 +127,28 @@ access(all) contract RandomBeaconHistory {
         //
         /// gaps only occur in the rare event of a system transaction failure. In this case, entries are still
         /// filled using a source not known at the time of block execution, which guarantees unpredicatability.
-        access(contract) fun backfill(randomSource: [UInt8]) {
-
-            let correctIndex = self.findGapAndReturnCorrectIndex()
-            var arrayLength = UInt64(RandomBeaconHistory.randomSourceHistory.length)
-            // optional optimization for the happy common path: if no gaps are detected,
+        access(contract) fun backfill(randomSource: [UInt8], currentArrayLength: UInt64, correctIndex: UInt64) {
```
sisyphusSmiling, line 130: My preference for this interface is to feed the random source and gather the provided values directly from contract fields, especially since this method lacks conditional checks. The perceived risk here is that the method can be called with arbitrary values which are not guaranteed to be limited to the current block height.

IMO the use of `findGapAndReturnCorrectIndex()` in conjunction with `backfill(randomSource: [UInt8])` corrects gaps limited by the current block height with as far as I can tell a single duplicate call to `RandomBeaconHistory.randomSourceHistory.length`, which is cheap constant time.

sisyphusSmiling (commented):
Looks like changes in [9e5d257](https://github.com/onflow/flow-core-contracts/pull/416/commits/9e5d257092612bdc431372fec17903e067a2a1ae) overwrote interface changes in #418 - is there a reason we want to specify the values when calling `backfill` instead of deriving them in scope?

contracts/RandomBeaconHistory.cdc
tarakby: oh my bad I didn't mean to overwrite anything, the commit https://github.com/onflow/flow-core-contracts/pull/416/commits/9e5d257092612bdc431372fec17903e067a2a1ae purpose was only to address Bastian's comments, but I think there is a mess up in my local branches and I force-pushed, let me clean that up.

tarakby: It should be fixed now, commit https://github.com/onflow/flow-core-contracts/pull/416/commits/75d3500157010ad01bca07d1ad0bde0ef50d6bfc addresses Bastian's comments and shouldn't overwrite other things.

contracts/RandomBeaconHistory.cdc
```diff
@@ -36,25 +49,172 @@ access(all) contract RandomBeaconHistory {
...
             let currentBlockHeight = getCurrentBlock().height
+            // init lowestBlockHeight if it is not set yet
             if RandomBeaconHistory.lowestHeight == nil {
                 RandomBeaconHistory.lowestHeight = currentBlockHeight
             }
+            // Create & save Backfiller if it is not saved yet
+            if RandomBeaconHistory.borrowBackfiller() == nil {
+                RandomBeaconHistory.account.save(<-create Backfiller(), to: /storage/randomBeaconHistoryBackfiller)
+            }
+            let backfiller = RandomBeaconHistory.borrowBackfiller() ?? panic("Problem borrowing backfiller")
```
janezpodhostnik, line 73:
```suggestion
            var maybeBackfiller = RandomBeaconHistory.borrowBackfiller()
            if maybeBackfiller == nil {
                RandomBeaconHistory.account.save(<-create Backfiller(), to: /storage/randomBeaconHistoryBackfiller)
                maybeBackfiller = RandomBeaconHistory.borrowBackfiller()
            }
            let backfiller = maybeBackfiller ?? panic("Problem borrowing backfiller")
```

This should be les operations which is really important for the system transaction.
(disclaimer: untested code)

contracts/RandomBeaconHistory.cdc
```diff
@@ -97,13 +257,17 @@ access(all) contract RandomBeaconHistory {
         }
         let index = blockHeight - self.lowestHeight!
         assert(
-            index >= 0 && index < UInt64(self.randomSourceHistory.length),
+            index >= 0,
             message: "Problem finding random source history index"
         )
+        assert(
+            index < UInt64(self.randomSourceHistory.length) && self.randomSourceHistory[index].length > 0,
+            message: "Source of randomness is currently not available but will be available soon"
```
janezpodhostnik, line 265: nice message!

contracts/RandomBeaconHistory.cdc
```diff
@@ -36,25 +49,172 @@ access(all) contract RandomBeaconHistory {
...
+            }
+
+            var newEntry = randomSource
+            var index = self.gapStartIndex
+            var count = 0 as UInt64
+            while count < self.maxEntriesPerCall {
+                // move to the next empty entry
+                while index < arrayLength && RandomBeaconHistory.randomSourceHistory[index].length > 0 {
+                    index = index + 1
+                }
```
janezpodhostnik, line 172: repeated code, maybe package into a function.

janezpodhostnik (approved):

contracts/RandomBeaconHistory.cdc
tarakby: good point: https://github.com/onflow/flow-core-contracts/pull/416/commits/a14b5e23f1e1902e5a14a88fce8f5cf322d4a290

contracts/RandomBeaconHistory.cdc
tarakby: I wrote a function to look for the next index but I think this ends up having more instructions/cost than the inlined version (new variables, function call)
The inlined version is used twice and it's too simple IMO to be packaged into a function, unless there is another point I missed

contracts/RandomBeaconHistory.cdc
```diff
@@ -36,25 +49,170 @@ access(all) contract RandomBeaconHistory {
...
+            // correct index to fill with the new random source
+            // so that eventually randomSourceHistory[correctIndex] = inputRandom
+            let correctIndex = currentBlockHeight - RandomBeaconHistory.lowestHeight!
+
+            // cache the array length in a local variable to avoid multiple calls to `length`
+            // when the array is too large
+            var arrayLength = UInt64(RandomBeaconHistory.randomSourceHistory.length)
+
+            // if a new gap is detected, emit an event
+            if correctIndex > arrayLength {
```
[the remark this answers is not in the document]
tarakby: Since accessing the length is constant time and cheap, I added this minor optimization FYI https://github.com/onflow/flow-core-contracts/pull/416/commits/e913f3aa42484ff822b844027b4a531f35fba4e7

tarakby:
@joshuahannan can you please have a look? I think the PR can only be merged after your approval.

joshuahannan (approved):
After we merge this, you'll need to make an upgrade plan and post it in a forum post like we've done for [previous upgrades](https://forum.flow.com/t/upgrade-coming-to-flow-epoch-smart-contracts-for-security-improvement/5443).

Then you'll also need to make a branch based off of the `stable-cadence` branch that includes these changes so that we can make sure the upgrades are working properly in the Cadence 1.0 migrations.

tarakby:
- https://forum.flow.com/t/upgrade-coming-to-the-randomness-history-smart-contract/5903
- https://github.com/onflow/flow-core-contracts/issues/421

PR #417, opened 2024-03-28
Move backfill updates to new Backfiller resource

sisyphusSmiling:
Related: #414 
Proposing changes to #416 due to contract updatability restrictions


tarakby (approved):
Awesome, thank you!

I have minor comments that I can address in the main PR, mainly to consolidate all the backfilling logic under the new resource, which is a minor refactor.

I assume the purpose of this PR is to solve the updatability issue. For this I let another reviewer confirm and we can merge.

contracts/RandomBeaconHistory.cdc
```diff
@@ -130,15 +88,84 @@ access(all) contract RandomBeaconHistory {
...
+    access(all) resource Backfiller {
+        /// Start index of the first gap in the `randomSourceHistory` array where random sources were not recorded, because
+        /// of a heartbeat failure.
+        /// There may be non contiguous gaps in the history, `gapStartIndex` is the start index of the lowest-height
+        /// gap.
+        /// If no gap exists, `gapStartIndex` is equal to the `randomSourceHistory` array length.
+        access(contract) var gapStartIndex: UInt64
+
+        init() {
+            self.gapStartIndex = 0
```
bluesign, line 114: here maybe we some value other than 0, maybe init parameter can be good option

bluesign:
I think we can move gap logic into backfiller, It will be cheaper and cleaner.

sisyphusSmiling:
Agreed @bluesign, makes sense to move that logic to the backfiller resource. Think it's alright to do that in #416?

contracts/RandomBeaconHistory.cdc
tarakby: Yes, I have added an item about this "init" to the todo list of #416. The value can be the SoR array length, it should work for both a fresh init and the with the current state init.

tarakby: If this PR #417 solves the adaptability issue, then we can merge it and address the logic points back in #416

tarakby:
> I think we can move gap logic into backfiller, It will be cleaner.

Totally agree. This is the refactor I mentioned in my review to keep all the backfilling logic in one place :)

tarakby:
Merging this for now, and addressing the mentioned ideas in #416

PR #419, opened 2024-04-08
[Cadence 1.0] Add `ProtocolStateVersionUpgrade` service event

jordanschalm:
This PR adds a minimal version of a service event to upgrade the protocol state version (see also https://github.com/onflow/flow-go/pull/5428). This is a Cadence 1.0 compatible port of https://github.com/onflow/flow-core-contracts/pull/411.

---

For now, I have added this service event to the `NodeVersionBeacon` contract, which has some benefits and drawbacks:
- Conceptually `ProtocolStateVersionUpgrade` does something similar to the existing Version Beacon, so it makes sense for their logic to be together
- The Version Beacon contract already has a Heartbeat resource hooked up to the system chunk, which is necessary to emit service events
- The existing version beacon code is written in a way that suggests the version it is operating on is global, rather than specifically scoped to the Execution State Machine
  - This, combined with the limitations of upgrading contracts, makes adding `ProtocolStateVersionUpgrade` cleanly difficult


contracts/NodeVersionBeacon.cdc
```diff
@@ -336,6 +350,17 @@ access(all) contract NodeVersionBeacon {
...
+
+        /// Emit a ProtocolStateVersionUpgrade event, if a pending upgrade is in storage.
+        access(self) fun emitProtocolStateVersionUpgradeEventIfNeeded() {
+            let pendingUpgrade = NodeVersionBeacon.popPendingProtocolStateVersionUpgrade()
+            if pendingUpgrade == nil {
+                return
+            }
+            emit NodeVersionBeacon.ProtocolStateVersionUpgrade(
+                newProtocolVersion: pendingUpgrade!.version,
+                activeView: pendingUpgrade!.activeView)
```
turbolent, line 362: Can use an if-let here instead of force-unwrapping:

```suggestion
            if let pendingUpgrade = NodeVersionBeacon.popPendingProtocolStateVersionUpgrade() {
                emit NodeVersionBeacon.ProtocolStateVersionUpgrade(
                    newProtocolVersion: pendingUpgrade.version,
                    activeView: pendingUpgrade.activeView
                )
            }
```

contracts/NodeVersionBeacon.cdc
```diff
@@ -495,6 +520,47 @@ access(all) contract NodeVersionBeacon {
...
+    /// Reads the pending ProtocolStateVersionUpgrade from storage, without removing it, and returns it.
+    /// If no ProtocolStateVersionUpgrade was in storage, returns nil.
+    access(all) fun peekPendingProtocolStateVersionUpgrade(): ViewActivatedVersion? {
+        return self.account.storage.copy<ViewActivatedVersion>(from: /storage/PendingProtocolStateVersionUpgrade)
+    }
+
+    /// Stores the pending ProtocolStateVersionUpgrade to storage.
+    /// No ProtocolStateVersionUpgrade may already be stored, otherwise the transaction will revert.
+    access(contract) fun storePendingProtocolStateVersionUpgrade(upgrade: ViewActivatedVersion) {
+        self.account.storage.save<ViewActivatedVersion>(upgrade, to: /storage/PendingProtocolStateVersionUpgrade)
```
turbolent, line 561: Type argument for type parameter can get inferred here

```suggestion
        self.account.storage.save(upgrade, to: /storage/PendingProtocolStateVersionUpgrade)
```

PR #420, opened 2024-04-12
EpochRecover service event and transaction

kc1116:
This PR updates the FlowEpoch smart contract to support recovering the network while in Epoch Fallback Mode. It adds a new service event `EpochRecover` which contains the metadata for the recovery epoch. This metadata is generated out of band using the bootstrap utility `util epoch efm-recover-tx-args` [PR](https://github.com/onflow/flow-go/pull/5576) and submitted to the contract with the [recovery_epoch.cdc transaction](https://github.com/onflow/flow-core-contracts/blob/f7a9be1fb91c966409257b1bf6b54e94c4dcd386/transactions/epoch/admin/recover_epoch.cdc). The FlowEpoch contract will end the current epoch, start the recovery epoch and store the metadata for the recovery epoch in storage. This metadata will then be emitted to the network during the next heartbeat interval.


joshuahannan:
@kc1116 Can you base this off the `stable-cadence` branch instead of `master` since we wouldn't be doing the upgrade until after Cadence 1.0?

contracts/epochs/FlowEpoch.cdc
```diff
@@ -406,6 +541,92 @@ pub contract FlowEpoch {
...
+            for nodeID in nodeIDs {
+                /// @TODO: if node is SN ensure weight > 0
+                /// throw exception, or add with 0 weight
+                nodes.append(FlowIDTableStaking.NodeInfo(nodeID))
+            }
+
+            // Compute the target end time for the next epoch
+            let timingConfig = FlowEpoch.getEpochTimingConfig()
+            let proposedTargetDuration = timingConfig.duration
+            let proposedTargetEndTime = timingConfig.getTargetEndTimeForEpoch(FlowEpoch.proposedEpochCounter())
```
AlexHentschel, line 586: I don't have enough context, but I am worried that we might not be able to use the happy path timing config, depending on what we want the recovery epoch to do. 
Even if it is ok to use the happy path `EpochTimingConfig`, I would like a comment that explains why this is safe and which assumptions we are making (i.e. under which conditions this code would need to change).

contracts/epochs/FlowEpoch.cdc
```diff
@@ -406,6 +541,92 @@ pub contract FlowEpoch {
...
+
+            // Create new Epoch metadata for the next epoch
+            // with the new values
+            let newEpochMetadata = EpochMetadata(
+                counter: FlowEpoch.currentEpochCounter + 1,
```
AlexHentschel, line 591: I am not sure this is correct. During EFM, the Protocol State formally remains in the same epoch, which is regularly being extended until we reach an EpochRecovery event. However, the Epoch Smart contract does not know this and behaves as if the epochs where just progressing on the happy path. So very likely, the Epoch smart contract and the Protocol State have a different notion of what the "next" epoch counter should be. 

Same suggestion as before:
Even if it is ok to compute the epoch counter like this, I would like a comment that explains why this is safe and which assumptions we are making (i.e. under which conditions this code would need to change).

contracts/epochs/FlowEpoch.cdc
```diff
@@ -406,6 +541,92 @@ pub contract FlowEpoch {
...
+                seed: randomSource,
+                startView: startView,
+                endView: endView,
+                stakingEndView: stakingEndView,
+                // This will be overwritten in `calculateAndSetRewards` below
```
AlexHentschel, line 596: When you write: "This will be overwritten" I thought it only applies to the immediately next field. Though, I think you are changing multiple of the following field later, right? I that is the case, I would suggest:
```suggestion
                // The following fields will be overwritten in `calculateAndSetRewards` below
```

contracts/epochs/FlowEpoch.cdc
```diff
@@ -406,6 +541,92 @@ pub contract FlowEpoch {
...
+                totalRewards: UFix64(0.0),
+                collectorClusters: [],
+                clusterQCs: [],
+                dkgKeys: dkgPubKeys)
+
+            FlowEpoch.saveEpochMetadata(newEpochMetadata)
```
AlexHentschel, line 602: so we save the `EpochMetadata` and _then_ change its fields? Is that safe? Can we avoid such confusion by maybe moving the `saveEpochMetadata` call further below, or add a comment (?)

contracts/epochs/FlowEpoch.cdc
```diff
@@ -406,6 +541,92 @@ pub contract FlowEpoch {
...
+            // Calculate rewards for the current epoch
+            // and set the payout for the next epoch
+            FlowEpoch.calculateAndSetRewards()
+            // Start a new Epoch, which increments the current epoch counter
+            FlowEpoch.startNewEpoch()
```
AlexHentschel, line 607: The Protocol State is the ultimate source of truth, because it controls what the network is _actually_ doing.  

Generally, when we process the Epoch Recovery transaction (governance transaction), I am not sure it is correct to start the new epoch right away. We should only start it once we reach the specified first view

contracts/epochs/FlowEpoch.cdc
AlexHentschel: furthermore, aren't we changing parameters for _all_ upcoming epochs (assuming no further human intervention via governance transactions) ?

AlexHentschel (commented):
Overall, I have limited background about the epoch smart contracts. Though, I feel there are a bunch of important details in this PR that still need to be worked out. Overall, we are filling various fields with values according to the happy path. But I miss explanations why those values we use (e.g. epoch counter, number of views for various phase, etc) are also correct for the unhappy path.

The general bar for us as BFT-engineers is that we can explain for every line of code in our PR, why it needs to be precisely like we wrote it. If that reasoning isn't absolutely trivial, a comment in the code is highly desired. There were multiple places, where I got stuck just not understanding why it is correct to chose a value according to your code. 

I would suggest to go over the code again with the mindset:
* during EFM, the Protocol State controlling the nodes has diverged from the Epoch Smart contract. Algorithmically, they are irreconcilable. The Epoch Recovery is a human intervention to reconcile both states:
   * Due to the prior diversion, the happy path algorithm for moving to the next epoch doesn't apply to either states anymore. That is fine, because we are having humans force-override values via a governance transaction. 
   * So here, we are at the point where the Epoch Smart contract received the human override (governance transaction). The human provides some values, others are need to be computed. But how precisely? Keep in mind that the Epoch Smart contract potentially went though multiple Epochs that do not exist in reality. 
   * When you write this code for processing the human override, for every value relevant to epochs (modified or not) you need to have a argument essentially proving that (i) when the value is processed by the Protocol State, it brings the Protocol State back to the happy path _and_ the resulting state it is compatible with _all_ assumptions and consistency requirements of the Epoch Protocol State and (ii) when the value is  processed by the Epoch Smart contract,  the new value brings the Epoch Smart contract back to the happy path _and_ the resulting state it is compatible with _all_ assumptions and consistency requirements of the Epoch Smart contract.

   Literally, I would suggest to go over every value you are writing in this PR. Try to answer questions (i) and (ii) and modify your algorithm until you feel confident that you have a bullet-proof argument. If in doubt, talk to Jordan. 

This will take time and we expect it to take time. Don't feel pressured to crank PRs out. Correctness and good documentation is by far a higher priority in our field of work than development speed.

contracts/epochs/FlowEpoch.cdc
```diff
@@ -147,6 +147,141 @@ pub contract FlowEpoch {
...
+        pub let nodeInfo: [FlowIDTableStaking.NodeInfo]
+
+        /// The first view (inclusive) of the RecoveryEpoch.
+        pub let firstView: UInt64
+
+        /// The last view (inclusive) of the RecoveryEpoch.
+        pub let finalView: UInt64
+
+        /// The last view (inclusive) of the staking phase.
+        pub let stakingEndView: UInt64
```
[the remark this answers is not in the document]
kc1116: Overlooked unused var removed https://github.com/onflow/flow-core-contracts/pull/420/commits/381ca6f1120bf8f599c87320c3c32ee18373672a

contracts/epochs/FlowEpoch.cdc
kc1116: It is safe to use the epoch counter in the smart contract because the counter is not incremented when we enter EFM state. Som, despite adding EpochExtensions the counter will not be incremented until the manual intervention via epoch recover transaction. Added a comment https://github.com/onflow/flow-core-contracts/pull/420/commits/4c0dcf7bbfd903fb6a08e8bd61520542e8d70135

contracts/epochs/FlowEpoch.cdc
```diff
@@ -432,6 +559,97 @@ access(all) contract FlowEpoch {
...
+                collectorClusters: collectorClusters,
+                randomSource: randomSource,
+                dkgPhase1FinalView: startView + FlowEpoch.configurableMetadata.numViewsInStakingAuction + FlowEpoch.configurableMetadata.numViewsInDKGPhase - 1 as UInt64,
+                dkgPhase2FinalView: startView + FlowEpoch.configurableMetadata.numViewsInStakingAuction + (2 as UInt64 * FlowEpoch.configurableMetadata.numViewsInDKGPhase) - 1 as UInt64,
+                dkgPhase3FinalView: startView + FlowEpoch.configurableMetadata.numViewsInStakingAuction + (3 as UInt64 * FlowEpoch.configurableMetadata.numViewsInDKGPhase) - 1 as UInt64,
+                targetDuration: targetDuration,
+                targetEndTime: targetEndTime,
+                clusterQCVoteData: clusterQCVoteData,
+                dkgPubKeys: dkgPubKeys,
+            )
```
turbolent, line 622: Remove the casts, they should not be necessary.

Maybe also extract the repeated long expressions into variables, so the code is more readable:


```suggestion
            /// Create the recovery epoch metadata, this struct is used to hold 
            /// EpochRecover service event data.
            
            let numViewsInStakingAuction = FlowEpoch.configurableMetadata.numViewsInStakingAuction
            let numViewsInDKGPhase = FlowEpoch.configurableMetadata.numViewsInDKGPhase
            
            let recoverEpochMetadata = FlowEpoch.RecoverEpochMetadata(
                counter: FlowEpoch.proposedEpochCounter(),
                nodeIDs: nodeIDs,
                firstView: startView,
                finalView: endView,
                collectorClusters: collectorClusters,
                randomSource: randomSource,
                dkgPhase1FinalView: startView + numViewsInStakingAuction + numViewsInDKGPhase - 1,
                dkgPhase2FinalView: startView + numViewsInStakingAuction + (2 * numViewsInDKGPhase) - 1,
                dkgPhase3FinalView: startView + numViewsInStakingAuction + (3 * numViewsInDKGPhase) - 1,
                targetDuration: targetDuration,
                targetEndTime: targetEndTime,
                clusterQCVoteData: clusterQCVoteData,
                dkgPubKeys: dkgPubKeys,
            )
```

contracts/epochs/FlowEpoch.cdc
```diff
@@ -745,9 +985,9 @@ access(all) contract FlowEpoch {
...
-                        DKGPhase2FinalView: proposedEpochMetadata.startView + self.configurableMetadata.numViewsInStakingAuction + (2 as UInt64 * self.configurableMetadata.numViewsInDKGPhase) - 1 as UInt64,
-                        DKGPhase3FinalView: proposedEpochMetadata.startView + self.configurableMetadata.numViewsInStakingAuction + (3 as UInt64 * self.configurableMetadata.numViewsInDKGPhase) - 1 as UInt64,
+                        dkgPhase1FinalView: proposedEpochMetadata.startView + self.configurableMetadata.numViewsInStakingAuction + self.configurableMetadata.numViewsInDKGPhase - 1 as UInt64,
+                        dkgPhase2FinalView: proposedEpochMetadata.startView + self.configurableMetadata.numViewsInStakingAuction + (2 as UInt64 * self.configurableMetadata.numViewsInDKGPhase) - 1 as UInt64,
+                        dkgPhase3FinalView: proposedEpochMetadata.startView + self.configurableMetadata.numViewsInStakingAuction + (3 as UInt64 * self.configurableMetadata.numViewsInDKGPhase) - 1 as UInt64,
```
turbolent, line 990: Same as above: Remove unnecessary casts and introduce variables for the repeated long expressions

contracts/epochs/FlowEpoch.cdc
```diff
@@ -234,7 +362,7 @@ access(all) contract FlowEpoch {
             self.dkgKeys = keys
         }
     }
-
+
```
turbolent, line 365:
```suggestion

```

contracts/epochs/FlowEpoch.cdc
```diff
@@ -147,6 +147,134 @@ access(all) contract FlowEpoch {
...
+            finalView: UInt64,
+            collectorClusters: [[String]],
+            randomSource: String,
+            dkgPhase1FinalView: UInt64,
+            dkgPhase2FinalView: UInt64,
+            dkgPhase3FinalView: UInt64,
+            targetDuration: UInt64,
+            targetEndTime: UInt64,
+            clusterQCVoteData: [FlowClusterQC.ClusterQCVoteData],
+            dkgPubKeys: [String]) {
```
turbolent, line 260:
```suggestion
            dkgPubKeys: [String]
        ) {
```

contracts/epochs/FlowEpoch.cdc
```diff
@@ -147,6 +147,134 @@ access(all) contract FlowEpoch {
         dkgPubKeys: [String]
     )
 
+    /// The Epoch Recover service event is emitted when a recoverEpoch transaction is submitted to the network. When the network 
+    /// gets into an EFM state the recoverEpoch transaction is used to override the current epoch and recover the network without 
+    /// sporking. This service event is processed by protocol participants to update their local protocol state with the new recover
+    /// epoch data.
+    access(all) event EpochRecover (
```
turbolent, line 154: That's a massive event. Are you sure it's OK to include so much data right into the event?

It also seems like this event has (almost?) the same contents as `RecoverEpochMetadata`?

contracts/epochs/FlowEpoch.cdc
```diff
@@ -432,6 +559,97 @@ access(all) contract FlowEpoch {
...
+            endView: UInt64,
+            targetDuration: UInt64,
+            targetEndTime: UInt64,
+            collectorClusters: [[String]],
+            clusterQCVoteData: [FlowClusterQC.ClusterQCVoteData],
+            dkgPubKeys: [String],
+            nodeIDs: [String])
+        {
+            pre {
+                FlowEpoch.isValidPhaseConfiguration(stakingEndView-startView+1, FlowEpoch.configurableMetadata.numViewsInDKGPhase, endView-startView+1):
```
turbolent, line 580: Even though it is only used here and not added, but maybe improve `isValidPhaseConfiguration` and add argument labels, it's very hard to see what the arguments passed to the function are, and if they are correct. They all seem to have the same type too, which makes it very easy to mix them up.

contracts/epochs/FlowEpoch.cdc
```diff
@@ -432,6 +559,97 @@ access(all) contract FlowEpoch {
...
+            let newEpochMetadata = EpochMetadata(
+                /// When the network enters EFM the epoch counter will not be incremented 
+                /// so it's safe to use current_epoch_counter + 1 
+                counter: FlowEpoch.proposedEpochCounter(),
+                seed: randomSource,
+                startView: startView,
+                endView: endView,
+                stakingEndView: stakingEndView,
+                // The following fields will be overwritten in `calculateAndSetRewards` below
+                totalRewards: UFix64(0.0),
```
turbolent, line 638: Remove unnecessary conversion:

```suggestion
                totalRewards: 0.0,
```

contracts/epochs/FlowEpoch.cdc
```diff
@@ -432,6 +559,97 @@ access(all) contract FlowEpoch {
...
+                collectorClusters: [],
+                clusterQCs: [],
+                dkgKeys: dkgPubKeys)
```
turbolent, line 641:
```suggestion
                dkgKeys: dkgPubKeys
            )
```

lib/go/test/epoch_test_helpers.go
```diff
@@ -712,6 +787,55 @@ func verifyEpochCommit(
...
 }
 
+// verifyEpochRecover verifies that the EpochRecover event values are equal to the provided expected values
+func verifyEpochRecover(
+	t *testing.T,
+	adapter *adapters.SDKAdapter,
+	epochAddress flow.Address,
+	expectedRecover EpochRecover) {
+	var emittedEvent EpochRecoverEvent
+	evtTypeStr := fmt.Sprintf("A.%s.FlowEpoch.EpochRecover", epochAddress.String())
```
turbolent, line 797: Use a `common.AddressLocation` and `common.Location.TypeID(...)` instead

turbolent (commented):
Mainly reviewed the Cadence code from a code quality perspective. I can't say anything in regards to the business logic

contracts/epochs/FlowEpoch.cdc
jordanschalm: > The Protocol State is the ultimate source of truth

I don't completely agree with that mindset. It's true that the Protocol State is the source of truth for the network, but in the steady state (ignoring sporks for a moment) its state is populated by trusted service events from the smart contracts. Ultimately the smart contracts are determining t

When the system contract emits the `EpochRecover` event it can either:
1. Immediately transition into the `RecoveryEpoch`
2. Wait until the start view, then transition into the `RecoveryEpoch`

Intuitively (2) seems like the better option, but with the context that we are currently already in a bad state (EFM), I'm not fully convinced it is here. Neither are great options, but (1) is less complex and parallels how `resetEpoch` works.

contracts/epochs/FlowEpoch.cdc
```diff
@@ -432,6 +559,97 @@ access(all) contract FlowEpoch {
...
+
+            /// Save the new epoch meta data, it will be referenced as 
+            /// the epoch progresses through each phase.
+            FlowEpoch.saveEpochMetadata(newEpochMetadata)
+
+            /// Calculate rewards for the current epoch
+            /// and set the payout for the next epoch
+            FlowEpoch.calculateAndSetRewards()
+            /// Start a new Epoch, which increments the current epoch counter
+            FlowEpoch.startNewEpoch()
```
jordanschalm, line 651: It's possible that the `EpochRecover` will be rejected by the Protocol State, in which case we would need to send a new `recoverEpoch` transaction.

We should make sure this function is decoupled from reward payout so we can, if necessary, re-try this process without double-paying rewards.

contracts/epochs/FlowEpoch.cdc
```diff
@@ -117,9 +117,9 @@ access(all) contract FlowEpoch {
         /// EpochSetup phase. Deadlines are specified in terms of a consensus view number.
         /// When a DKG participant observes a finalized and sealed block with view greater
         /// than the given deadline, it can safely transition to the next phase.
-        DKGPhase1FinalView: UInt64,
-        DKGPhase2FinalView: UInt64,
-        DKGPhase3FinalView: UInt64,
+        dkgPhase1FinalView: UInt64,
+        dkgPhase2FinalView: UInt64,
+        dkgPhase3FinalView: UInt64,
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-core-contracts/pull/420/commits/2e21e417cbfb34e570ddfe1d43ae6e34d11333ef

contracts/epochs/FlowEpoch.cdc
```diff
@@ -147,6 +147,134 @@ access(all) contract FlowEpoch {
...
+        /// for the upcoming epoch.
+        randomSource: String,
+
+        /// The deadlines of each phase in the DKG protocol to be completed in the upcoming
+        /// EpochSetup phase. Deadlines are specified in terms of a consensus view number. 
+        /// When a DKG participant observes a finalized and sealed block with view greater 
+        /// than the given deadline, it can safely transition to the next phase. 
+        dkgPhase1FinalView: UInt64,
+        dkgPhase2FinalView: UInt64,
+        dkgPhase3FinalView: UInt64,
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-core-contracts/pull/420/commits/2e21e417cbfb34e570ddfe1d43ae6e34d11333ef

contracts/epochs/FlowEpoch.cdc
```diff
@@ -147,6 +147,134 @@ access(all) contract FlowEpoch {
...
+        /// The DKG public keys passed in the recoverEpoch transaction. These are re-used from the
+        /// last successful DKG.
+        dkgPubKeys: [String],
+    )
+
+    /// Contains metadata about the recovery epoch, this data 
+    /// is stored at the storage path /storage/recoverEpochMetadata.
+    /// This struct is a 1:1 copy of the event EpochRecover event and is
+    /// used to populate all the fields of the event when the heartbeat 
+    /// detects a new RecoverEpochMetadata stored.
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-core-contracts/pull/420/commits/9c60f6f6fdde6ace515061f328febe14b1966009

contracts/epochs/FlowEpoch.cdc
kc1116: https://github.com/onflow/flow-core-contracts/pull/420/commits/5b08984f38e26363d4777233f5fb3077f9f6f406

contracts/epochs/FlowEpoch.cdc
kc1116: https://github.com/onflow/flow-core-contracts/pull/420/commits/5b08984f38e26363d4777233f5fb3077f9f6f406

lib/go/test/epoch_test_helpers.go
kc1116: https://github.com/onflow/flow-core-contracts/pull/420/commits/5edced7a617bb1ecfd358e6e65d25f94acc3fdde

contracts/epochs/FlowClusterQC.cdc
```diff
@@ -258,6 +258,24 @@ access(all) contract FlowClusterQC {
         }
     }
 
+    /// Represents the quorum certificate vote data for a signer 
+    /// of the certificate.
+    access(all) struct ClusterQCVoteData {
```
jordanschalm, line 263: In order to match the [current `util` command output](https://github.com/onflow/flow-go/blob/70f9aabc9ec76bd1a1c23b2bf99880fb2ad4dc2e/cmd/util/cmd/common/clusters.go#L165-L184), I think this structure needs to have the form:
```cadence
struct ClusterQCVoteData {
  voteData: String // hex-encoded aggregated signature
  voterIDs: [String]
}
```

lib/go/test/flow_epoch_test.go
```diff
@@ -1516,3 +1517,203 @@ func TestEpochReset(t *testing.T) {
...
+		}
+
+		args := []cadence.Value{
+			cadence.NewUInt64(startView),
+			cadence.NewUInt64(stakingEndView),
+			cadence.NewUInt64(endView),
+			cadence.NewUInt64(targetDuration),
+			cadence.NewUInt64(targetEndTime),
+			cadence.NewArray(collectorClusters),        // collectorClusters
+			cadence.NewArray(make([]cadence.Value, 0)), // clusterQCVoteData
```
jordanschalm, line 1627: Given that this seems to work (3 clusters and 0 QCs), perhaps we should add a sanity check to `recoverEpoch` that `len(clusters) == len(clusterQCs)`

transactions/epoch/admin/recover_epoch.cdc
```diff
@@ -0,0 +1,48 @@
...
+        let epochAdmin = signer.storage.borrow<&FlowEpoch.Admin>(from: FlowEpoch.adminStoragePath)
+            ?? panic("Could not borrow epoch admin from storage path")
+
+        if initNewEpoch {
+            epochAdmin.recoverNewEpoch(startView: startView,
+                    stakingEndView: stakingEndView,
+                    endView: endView,
+                    targetDuration: targetDuration,
+                    targetEndTime: targetEndTime,
+                    clusterAssignments: clusterAssignments  ,
```
jordanschalm, line 32:
```suggestion
                    clusterAssignments: clusterAssignments,
```
Likewise on line 42

transactions/epoch/admin/recover_epoch.cdc
```diff
@@ -0,0 +1,48 @@
...
+import FlowIDTableStaking from "FlowIDTableStaking"
+import FlowClusterQC from "FlowClusterQC"
+
+// The recoverEpoch transaction creates and starts a new epoch in the FlowEpoch smart contract
+// to which will force the network exit EFM. The recoverEpoch service event will be emitted 
+// and processed by all protocol participants and each participant will update their protocol 
+// state with the new Epoch data.
+// This transaction should only be used with the output of the bootstrap utility:
+//   util epoch efm-recover-tx-args
+transaction(startView: UInt64,
```
jordanschalm, line 11: We should include the `epochCounter` of the `RecoveryEpoch` here and use it for safety checks in the `recoverNewEpoch` implementation. 

In `util epoch efm-recover-tx-args`, we know exactly what the `RecoveryEpoch` epoch counter must be to be accepted by the Protocol State (current+1). If we don't pass that counter value here, then we're relying on the Protocol/SmartContract counters being synchronized in a particular way for correctness. In EFM, that is not guaranteed.

Currently we set the `RecoveryEpoch` counter as follows:
- if `initNewEpoch` then `RecoveryEpoch.counter` = currentEpochCounter+1
- if `!initNewEpoch` then `RecoveryEpoch.counter` = currentEpochCounter

I would change this to:
- pass in `recoveryEpochCounter` as an argument
- change `initNewEpoch` to be `unsafeAllowOverwrite`

In `recoverNewEpoch`, check that `recoveryEpochCounter == currentEpochCounter+1`.
- If that passes, then continue
- If that fails, then only continue if `unsafeAllowOverwrite` is set to true.

lib/go/test/flow_epoch_test.go
```diff
@@ -1516,3 +1517,203 @@ func TestEpochReset(t *testing.T) {
...
+			endView:               endView,
+			stakingEndView:        stakingEndView,
+			totalRewards:          "0.0",
+			rewardsBreakdownArray: 0,
+			rewardsPaid:           false,
+			collectorClusters:     nil,
+			clusterQCs:            nil,
+			dkgKeys:               dkgPubKeys,
+		}
+		verifyEpochMetadata(t, b, env, expectedMetadata)
```
jordanschalm, line 1663: Can we also validate that `curEpochCounter = startEpochCounter+1`

lib/go/test/flow_epoch_test.go
```diff
@@ -1516,3 +1517,203 @@ func TestEpochReset(t *testing.T) {
...
+			[]sdkcrypto.Signer{IDTableSigner},
+			false,
+		)
+
+		expectedRecoverEvent.firstView = newStartView
+		expectedRecoverEvent.finalView = newEndView
+		expectedRecoverEvent.dkgPhase1FinalView = newStartView + numStakingViews + numDKGViews - 1
+		expectedRecoverEvent.dkgPhase2FinalView = newStartView + numStakingViews + (2 * numDKGViews) - 1
+		expectedRecoverEvent.dkgPhase3FinalView = newStartView + numStakingViews + (3 * numDKGViews) - 1
+		verifyEpochRecover(t, adapter, idTableAddress, expectedRecoverEvent)
```
jordanschalm, line 1717: Can we also validate: 
- `curEpochCounter = startEpochCounter+1`
- check `epochMetadata` in addition to the EpochRecover event

contracts/epochs/FlowEpoch.cdc
```diff
@@ -432,6 +498,199 @@ access(all) contract FlowEpoch {
...
+                dkgPubKeys: dkgPubKeys,
+            )
+        }
+
+        /// Performs sanity checks for the provided epoch configuration. It will ensure the following;
+        /// - There is a valid phase configuration.
+        /// - All nodes in the node ids list have a weight > 0.
+        /// If the configuration is a valid configuration the the staking auction,
+        /// qc voting and dkg will be ended depending on the current epoch phase.
+        access(all) fun recoverEpochPreChecks(startView: UInt64,
```
jordanschalm, line 550: I would consider renaming this function (or moving the latter half that stops voting, DKG, or the staking auction to a separate function). 

To me, "sanity checks" implies that we are validating inputs but not changing state. Currently the function is both validating inputs and modifying state.

lib/go/test/flow_epoch_test.go
```diff
@@ -1516,3 +1517,203 @@ func TestEpochReset(t *testing.T) {
 		assertEqual(t, CadenceUFix64("249999.99300000"), result)
 	})
 }
+
+func TestEpochRecover(t *testing.T) {
```
jordanschalm, line 1521: Could you add documentation for the test here?

Currently, I think we are testing:
- execute an epoch recover where `recoveryEpochCounter == currentEpochCounter+1`
- execute an epoch recover where `recoveryEpochCounter == currentEpochCounter`

After we explicitly include the `recoveryEpochCounter`, I would also suggest testing the case where:
- `recoveryEpochCounter != currentEpochCounter+1` and the `unsafeAllowOverwrite` flag is not set

jordanschalm (commented):
Took another pass - my main change request is to make sure that the expected post-recovery epoch counter that we can compute in the `efm-recover-tx-args` utility is passed explicitly to the transaction. I think this is something we overlooked previously, but including it would eliminate a class of potential problems.

lib/go/test/flow_epoch_test.go
kc1116: https://github.com/onflow/flow-core-contracts/pull/420/commits/1e41ecf0ca8519eec08751639a3ad3928072c8a0

kc1116: Not sure if I follow the last point, if we have 2 separate funcs one to recover new epoch and one to recover current epoch we dont need to unsafeAllowOverwrite we dont pass unsafeAllowOverwrite to any of the FlowEpoch funcs we only use it in the transaction to decide which func `recoverNewEpoch` or `recoverCurrentEpoch`   to call.

lib/go/test/flow_epoch_test.go
kc1116: https://github.com/onflow/flow-core-contracts/pull/420/commits/614f489fb0448a5278b4eec15cee7742701370ae

lib/go/test/flow_epoch_test.go
kc1116: https://github.com/onflow/flow-core-contracts/pull/420/commits/614f489fb0448a5278b4eec15cee7742701370ae

lib/go/test/flow_epoch_test.go
```diff
@@ -1516,3 +1517,203 @@ func TestEpochReset(t *testing.T) {
...
+	)
+
+	// Register a DKG Participant
+	tx = createTxWithTemplateAndAuthorizer(b, templates.GenerateEpochRegisterDKGParticipantScript(env), addresses[1])
+	signAndSubmit(
+		t, b, tx,
+		[]flow.Address{addresses[1]},
+		[]sdkcrypto.Signer{signers[1]},
+		false,
+	)
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-core-contracts/pull/420/commits/61f07b634d3ea7446700059a4194330c6ac06ef7

lib/go/test/flow_epoch_test.go
kc1116: https://github.com/onflow/flow-core-contracts/pull/420/commits/a78b7d7ce5e271717b04aaa6f289612b01e4d6ea

contracts/epochs/FlowEpoch.cdc
```diff
@@ -432,6 +498,199 @@ access(all) contract FlowEpoch {
...
+            counter: FlowEpoch.proposedEpochCounter(),
+            seed: randomSource,
+            startView: startView,
+            endView: endView,
+            stakingEndView: stakingEndView,
+            // The following fields will be overwritten in `calculateAndSetRewards` below
+            totalRewards: 0.0,
+            collectorClusters: [],
+            clusterQCs: [],
+            dkgKeys: dkgPubKeys)
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-core-contracts/pull/420/commits/19a3795e676c9d4825a775c827859c2b772ec7ae, not quite sure what you mean I am following the indent format in the file.

transactions/epoch/admin/recover_epoch.cdc
kc1116: https://github.com/onflow/flow-core-contracts/commit/1e41ecf0ca8519eec08751639a3ad3928072c8a0

contracts/epochs/FlowEpoch.cdc
kc1116: https://github.com/onflow/flow-core-contracts/pull/420/commits/802d8284a8709fe3a118256d09baa070661f7b8b

contracts/epochs/FlowEpoch.cdc
```diff
@@ -432,6 +497,234 @@ access(all) contract FlowEpoch {
...
+            // sanity check we must receive qc vote data for each cluster
+            assert(
+                numOfClusterAssignments == numOfClusterQCVoteData, 
+                message: "number of cluster assignments does not match number of cluster qc vote data"
+            )
+        }
+        
+        /// Stops epoch components. If the configuration is a valid configuration the staking auction,
+        /// qc voting and dkg will be ended depending on the current epoch phase.
+        access(all) fun stopEpochComponents()
```
joshuahannan, line 576: `access(self)`

contracts/epochs/FlowEpoch.cdc
```diff
@@ -432,6 +497,234 @@ access(all) contract FlowEpoch {
...
+        }
+
+        /// Ends the currently active epoch and restarts it with the provided configuration.
+        /// This function is intended to update the current epoch configuration when a recovery
+        /// transaction needs to be submitted more than once. This function differs 
+        /// from recoverNewEpoch because it does not increment the epoch counter. It also 
+        /// does not calculate and set rewards avoiding double paying rewards for the same epoch.
+        /// This meta data will be emitted in the EpochRecover service event. This function is used 
+        /// within sporks to recover the network from Epoch Fallback Mode (EFM).
+        access(all) fun recoverCurrentEpoch(startView: UInt64,
```
joshuahannan, line 672: Are we sure we shouldn't combine this and `recoverNewEpoch` into a single function? Might save some lines of code and we could potentially just detect which one we need. I don't know for sure though

kc1116: Yes that how it initially was but we decided to make them 2 distinct functions so that it is more obvious the difference in behavior.

contracts/epochs/FlowEpoch.cdc
kc1116: https://github.com/onflow/flow-core-contracts/pull/420/commits/1f635156b348a418c12429e1664087786365c5c9

lib/go/test/flow_epoch_test.go
jordanschalm: Once we are passing in the `recoveryEpochCounter`, we must use that value as the epoch counter for the recovery epoch 100% of the time, otherwise we will produce an incompatible recovery epoch state. `recoveryEpochCounter` is the source of truth.

In the commit which was merged to the feature branch, the logic is:
1. if `unsafeAllowOverwrite` is set, then use `recoveryEpochCounter` as the counter for the recovery epoch
2. otherwise, ignore the `recoveryEpochCounter` input and assume that the the smart contract and protocol state epoch counters are synchronized

I think conditional branch 2 is unsafe and we should respect the `recoveryEpochCounter` codepath in all cases. Sorry if I didn't explain this properly in the previous review. The purpose of including `recoveryEpochCounter` at all is to make sure it is used as the epoch counter for the recovery epoch being constructed in the transaction, in all circumstances.

contracts/epochs/FlowEpoch.cdc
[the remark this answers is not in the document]
jordanschalm: Line 624 is starting a multi-line expression, constructing an instance of `EpochMetadata`. 

My suggestion is to indent the lines within that expression:
```
            let newEpochMetadata = EpochMetadata(
                /// Increment the epoch counter when recovering with a new epoch
                counter: FlowEpoch.proposedEpochCounter(),
                seed: randomSource,
                startView: startView,
                endView: endView,
                stakingEndView: stakingEndView,
                // The following fields will be overwritten in `calculateAndSetRewards` below
                totalRewards: 0.0,
                collectorClusters: [],
                clusterQCs: [],
                dkgKeys: dkgPubKeys)
```

Similar to what we do on, for example, [line 646-658](https://github.com/onflow/flow-core-contracts/pull/420/files#diff-1da2c6b9c85ee2833c92c917dd4f57655a9f505ff07e2d8088977b6bcdb1969bR646-R658)

kc1116:
Reopening

PR #319, opened 2022-11-02
FEATURE: Cadence 1.0 Core Contracts Changes

joshuahannan:
Includes changes related to Cadence 1.0

Many functions have been changed to `view` for new view function requirements
Updates casting to use correct `as` syntax
changes `pub` to `access(all)`
Adds entitlements
Uses the new fungible token standard and `Burner` contract
Removes custom destructors

PR #429, opened 2024-05-09
Add entitlement to LockedTokenManager

joshuahannan:
Adds entitlements to `LockedTokens.LockedTokenManager` to prevent accessing privileged methods via reference downcasting

Need to make sure that these will still be valid with the migrations


joshuahannan:
The only entitlement we've added for `LockedTokens.LockedTokenManager` is the `UnlockTokens` entitlement and we use that everywhere there is a `LockedTokenMananger` capability, so that should mean that the migrations will all work because the capabilities will all be migrated to their fully entitled types.

Additionally, I changed a lot of the fields in `LockedTokenManager` to be `access(account)`. This is fine because none of those are accessed in transactions, only in the `LockedTokens` and `FlowStakingCollection` contract

sisyphusSmiling (approved):

sisyphusSmiling:
Approved for merge to #319, only concern is the migrated type as you called out

PR #430, opened 2024-05-15
Add call to dependency audit contract

janezpodhostnik:
ref: https://github.com/onflow/flow-go/issues/5858

This is intended to work together with this FVM PR https://github.com/onflow/flow-go/pull/5908 which is already deployed on testnet, so this can be deployed when it is ready.

This imports the `DependenciesAudit` contract defined here: https://github.com/onflow/contract-updater/pull/33

This is only needed pre-1.0 and the `checkDependencies` function can (and will) be removed post-1.0. At that point the DependenciesAudit contract also becomes useless.  

DO NOT MERGE


sisyphusSmiling (commented):
Contract LGTM though CI is failing

janezpodhostnik:
CI is failing because the generated files were not intentionally not generated, so that this PR would not get accidentally merged.

I don't think this PR should be merged, since we want this feature to go away with 1.0. If we don't merge this PR that would happen automatically when the contract updates to 1.0 during the migration.

sisyphusSmiling:
I see what you're saying, that makes sense

joshuahannan:
@janezpodhostnik is it okay if we close this PR?

PR #435, opened 2024-06-10
Add a new script that checks if payer has balance to pay for tx

illia-malachyn:
This is a part of https://github.com/onflow/flow-go/pull/6004


transactions/FlowServiceAccount/scripts/check_if_payer_has_sufficient_balance.cdc
```diff
@@ -0,0 +1,8 @@
+import FlowFees from 0xFLOWFEESADDRESS
+
+pub fun main(payerAcct: Address, inclusionEffort: UFix64, maxExecutionEffort: UFix64): Bool {
+    let authAcct = getAuthAccount(payerAcct)
+    return FlowFees.verifyPayersBalanceForTransactionExecution(authAcct,
```
illia-malachyn, line 5: I tried to write it like 
`
let result =  verifyPayersBalanceForTransactionExecution(...)
return result.variable
`
but got a runtime error saying that the result was already declared. Do you know what the reason is?

lib/go/test/service_test.go
```diff
@@ -399,6 +399,21 @@ func TestContracts(t *testing.T) {
 
 	})
 
+	t.Run("Should check if payer has sufficient balance to execute tx", func(t *testing.T) {
+		// TODO: Is it better to set up a mock tx and test against it?
```
illia-malachyn, line 403: Can/Should I test it against some mocked tx?

illia-malachyn:
BTW, I noticed that `Should be able to change the conversion and minimum` fails when I run only my test

illia-malachyn:
@janezpodhostnik Can u take a look at it?

lib/go/templates/service_templates.go
```diff
@@ -192,3 +194,9 @@ func GenerateGetExecutionMemoryLimit(env Environment) []byte {
 
 	return []byte(ReplaceAddresses(code, env))
 }
+
+func GenerateCheckIfPayerHasSufficientBalance(env Environment) []byte {
```
peterargue, line 198: How about his naming (here and elsewhere)
```suggestion
func GenerateCheckPayersBalanceForTransactionExecution(env Environment) []byte {
```

lib/go/test/service_test.go
peterargue: yea. I think there should be 2 tests one covering a case when the payer has enough FLOW, and one covering when it doesn't.

It looks like the contract doesn't validate the range of the 2 numeric arguments, so you can probably just use a very large value for `gasLimit` to find one that causes the request to return false.

lib/go/test/service_test.go
```diff
@@ -399,6 +399,21 @@ func TestContracts(t *testing.T) {
...
+
+		// TODO: What account should I use here?
```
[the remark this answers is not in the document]
peterargue: there are a couple accounts already setup in this test, so you should be able to use one of those. Ideally, you want one with a small balance

transactions/FlowServiceAccount/scripts/check_if_payer_has_sufficient_balance.cdc
peterargue: I think this script should just return the same values as `verifyPayersBalanceForTransactionExecution`

lib/go/test/service_test.go
illia-malachyn: I guess I'm running into https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowFees.cdc#L111

lib/go/test/service_test.go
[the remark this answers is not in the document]
illia-malachyn: I run into this check while using some of them https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowFees.cdc#L111. There's nil returned by `executeScriptAndCheck` while using others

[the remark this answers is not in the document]
peterargue: @joshuahannan any ideas on the best way to approach this test?

This PR is adding a new script template to check that a payer has enough flow to cover the tx fees

janezpodhostnik (approved):
Looks good! Thanks

joshuahannan (approved):
Thank you so much for adding this! Great work. :)

lib/go/test/service_test.go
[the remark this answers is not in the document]
joshuahannan: You can fund an account by sending FLOW to it from the service account, which controls the FLOW minter resource. There is a go util in test.go that allows you to mint FLOW

joshuahannan:
Also, I'm merging the `stable-cadence` branch today, so after I do that, you'll need to pull the latest code from master, update everything here to use Cadence 1.0, and then I can review again

lib/go/test/service_test.go
```diff
@@ -401,6 +401,76 @@ func TestContracts(t *testing.T) {
 
 	})
 
+	t.Run("Should check if payer has sufficient balance to execute tx", func(t *testing.T) {
```
illia-malachyn, line 404: @joshuahannan  
I cannot test a new functionality because I'm running into this [check](https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowFees.cdc#L111) all the time. I noticed that this [function](https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowFees.cdc#L109) always returns zero. [This statement](https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowStorageFees.cdc#L175) is zero because [this one](https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowStorageFees.cdc#L133) is always true. 

That's why I submitted a tx `GenerateChangeStorageFeeParametersScript()` in this test. I wanted to set `FlowStorageFees.storageMegaBytesPerReservedFLOW` but it looks like I'm missing sth as I'm still falling through [this check](https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowFees.cdc#L111).

Can you help me with it?

illia-malachyn: I want to be able to adjust script arguments and amount of flow on the account to test that the script works as expected. For now I cannot to this because of the above-described problem.

joshuahannan: Are you saying that the transaction you are sending to set `storageMegaBytesPerReservedFLOW` is failing because of that check? Even if that check returns, the transaction should still be able to execute because it returns `true` for `canExecuteTransaction`. So you should be able to test after setting that field.

Also, I believe there is a way to start the emulator with transaction and storage fees turned on, so you might be able to just do that instead. Let me see if I can find that

joshuahannan: As you can see [here](https://github.com/onflow/flow-core-contracts/blob/master/lib/go/test/test.go#L85), we normally start the emulator with storage fees disabled, but you can probably set that to true and then it will be enabled and you won't have to set anything yourself.

There is probably an option for transaction fees too

joshuahannan: You can find all your emulator options here: https://github.com/onflow/flow-emulator/blob/master/emulator/blockchain.go#L118

illia-malachyn: > Are you saying that the transaction you are sending to set `storageMegaBytesPerReservedFLOW` is failing because of that check? Even if that check returns, the transaction should still be able to execute because it returns `true` for `canExecuteTransaction`. So you should be able to test after setting that field.
> 
> Also, I believe there is a way to start the emulator with transaction and storage fees turned on, so you might be able to just do that instead. Let me see if I can find that

I'm saying that `canExecuteTransaction` value is always `true` no matter what execution effort, gas limit or payer balance is

illia-malachyn: When I set `StorageLimitEnabled = true` I can see that the values of `VerifyPayerBalanceResult` start changing and I'm no longer running into that check, which is good. 
However, the `canExecuteTransaction` is still always equal to `true` as `requiredBalance` is always bigger or equal to the actual payer's balance. Even if I don't mint any tokens, account's balance is "enough to pay for tx". I tried to set both `inclusionEffort` and `gasLimit` to different values and it doesn't change the behavior. Is it expected?

janezpodhostnik: You are probably hitting this check: 
```
if minimumRequiredBalance == UFix64(0) {
            // If the required balance is zero exit early.
```
https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowFees.cdc#L111

In the FVM you need both  context options 
```
fvm.WithTransactionFeesEnabled(true),
fvm.WithAccountStorageLimit(true),
```
and bootstrap options
```
fvm.WithTransactionFee(fvm.DefaultTransactionFees),
fvm.WithStorageMBPerFLOW(fvm.DefaultStorageMBPerFLOW),
fvm.WithMinimumStorageReservation(fvm.DefaultMinimumStorageReservation),
fvm.WithAccountCreationFee(fvm.DefaultAccountCreationFee),
```

Not exactly sure how to set that up in a test in the core-contracts but I believe you have to set the emulator options in which case  you can use the corresponding emulator options:
```
WithStorageLimitEnabled
WithMinimumStorageReservation // this covers both WithMinimumStorageReservation and WithAccountCreationFee
WithStorageMBPerFLOW
WithTransactionFeesEnabled
```

illia-malachyn: Hey @janezpodhostnik. I did set up the options you recommended though the payer has still enough balance to pay for tx. I tried to execute some simple transaction and set my custom-created account as a tx payer but I'm running into errors saying it doesn't have sufficient signatures. Can u please take a look at it? I pushed a commit with the latest changes.

illia-malachyn: Just to be clear: I want to test the case where the payer has sufficient balance and insufficient one to pay for tx.
The issue is that the balance is always sufficient even if I don't mint any tokens and set inclusion effort to a big value.
I decided to execute some simple transaction (as a payer) to reduce the amount of Flow on the account so that I have insufficient balance and can test this case.

For now, I have a problem with signatures. However, I'd like to hear any advice from you If the thing I'm making does make sense.

illia-malachyn: @joshuahannan can you please take a look into this while Janez is off?

lib/go/test/service_test.go
```diff
@@ -401,6 +405,99 @@ func TestContracts(t *testing.T) {
...
+		err = tx.AddArgument(cadence.Int32(15))
+		require.NoError(t, err)
+
+		txRes := signAndSubmit(
+			t, b, tx,
+			[]flow.Address{accAddress},
+			[]crypto.Signer{accSigner},
+			false,
+		)
+		require.True(t, txRes.Succeeded())
```
janezpodhostnik, line 469: As it is now the transaction is set up with the wrong signers/authorisers. You can try changing this to:

```suggestion
			tx := flow.NewTransaction().
			SetScript(txCode).
			SetComputeLimit(9999).
			SetProposalKey(accAddress, 0, 0).
			SetPayer(accAddress).
			AddAuthorizer(accAddress)

		err = tx.AddArgument(cadence.Int32(15))
		require.NoError(t, err)

		err = tx.SignEnvelope(accAddress, 0, accSigner)
		assert.NoError(t, err)

		// this transaction should fail and be reverted, but the fees will still be paid
		// which will push the balance below the minimum account balance
		// calling VerifyPayerBalanceForTxExecution after this will return false.
		txRes := Submit(t, b, tx, true)
		require.True(t, txRes.Reverted())
```


The key thing to know is that the only way to get the account under the minimum account balance is to send a transaction (while the account is very close to the minimum account balance) that will fail, but some fees will still be deducted, which will result in the account being below minimum account balance.

This will also then mean that `VerifyPayerBalanceForTxExecution` will start returning false for that account.

lib/go/test/service_test.go
```diff
@@ -401,6 +405,99 @@ func TestContracts(t *testing.T) {
...
+		//mintTokensForAccount(t, b, env, accAddress, "15.0")
+
+		//TODO: remove this. this is for debug only
+		acc, err := adapter.GetAccount(context.Background(), accAddress)
+		require.NotNil(t, acc)
+		require.NoError(t, err)
+
+		// set up args
+		cadenceAddress := cadence.NewAddress(accAddress)
+		inclusionEffort := cadence.UFix64(20)
```
janezpodhostnik, line 481:
```suggestion
		inclusionEffort := cadence.UFix64(100_000_000) 
```
this should be `1.0 UFix64`

lib/go/test/service_test.go
illia-malachyn: Thank you so much!

joshuahannan (approved):

joshuahannan:
@illia-malachyn Thank you for adding this! Is this PR ready to be merged?

illia-malachyn:
@joshuahannan Yes, you can merge it

PR #441, opened 2024-08-07
Add node ID to index mapping to DKG data model

jordanschalm:
Ref: https://github.com/onflow/flow-go/issues/6213

### General Context
This PR modifies the DKG result data model to include an explicit mapping from node ID to DKG index. Currently the submission is an ordered list of keys, with the mapping being implicit based on the ordering. To support EFM Recovery we need to explicitly encode the index mapping and include it in the `EpochCommit` event. See https://github.com/onflow/flow-go/issues/6213 for details.

<details>
<summary>📝 Notes about implementation options from original draft</summary>

I broadly see two approaches to implement this:
1. Add contract fields (storage slots, because of upgrade limitations) which includes only the new index mapping, conceptually extending the existing fields `finalSubmissionByNodeID` and `uniqueFinalSubmissions`. 
  The contract would use the new and existing fields together, and would need to maintain consistency between them.
2. Add one contract field which tracks the entire submission structure (now including both a key list and index mapping), conceptually replacing the existing fields `finalSubmissionByNodeID` and `uniqueFinalSubmissions`.
  The existing fields would be deprecated (we can't remove them because of upgrade limitations, but they would not be referenced by any code in the contract).

After noodling around a bit with both, I'm inclined to go with Option 2, because it seems easier to work with and reason about the submission state when it's encapsulated in one field (and type). 

</details>

### Changes
- Add type `ResultSubmission` that encapsulates one complete DKG submission,  (replaces `[String?]`)
- Add type `SubmissionTracker` that encapsulates the state for all DKG submissions for one DKG instance
- Deprecate existing contract fields which track final submissions as `[String]`
- Adds `dkgGroupKey` and `dkgIdMapping` fields to `EpochCommit` (previously the group key was by convention the first entry in the single `dkgPubKeys` entry


joshuahannan:
I'm also leaning towards option 2. We actually can remove fields in upgrades. It doesn't delete the state, but it removes the ability to access it anywhere, so that is an option. Might be good to keep the field though for the small chance that we might actually need it again in the future

contracts/epochs/FlowDKG.cdc
```diff
@@ -357,6 +426,7 @@ access(all) contract FlowDKG {
     }
 
     /// Get the count of the final submissions array
+    // TODO: Should this be a view function?
```
joshuahannan, line 429: yes, probably should be view

joshuahannan (commented):
This looks like a good start!

contracts/epochs/FlowDKG.cdc
```diff
@@ -100,6 +92,81 @@ access(all) contract FlowDKG {
...
+            self.pubKeys = pubKeys
+            self.idMapping = idMapping
+        }
+
+        access(all) fun equals(other: FlowDKG.ResultSubmission): Bool {
+            if self.groupPubKey != other.groupPubKey {
+                return false
+            }
+            // TODO: validate equality check on list is safe
+            if self.pubKeys != other.pubKeys {
```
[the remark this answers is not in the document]
jordanschalm: Equivalent `ResultSubmission` would need to have them in the same order, because the index is meaningful in the DKG.

jordanschalm:
> I'm also leaning towards option 2. We actually can remove fields in upgrades. It doesn't delete the state, but it removes the ability to access it anywhere, so that is an option. Might be good to keep the field though for the small chance that we might actually need it again in the future

Sounds good. We'll move ahead in that direction then - thanks for the review! I agree keeping the deprecated fields (at least at first) is the better move.

contracts/epochs/FlowDKG.cdc
```diff
@@ -100,6 +90,230 @@ access(all) contract FlowDKG {
...
+    }
+
+    // ResultSubmission represents a result submission from one DKG participant.
+    // Each submission includes a group public key and an ordered list of participant public keys.
+    // A submission may be empty, in which case all fields are nil - this is used to represent a local DKG failure.
+    // All non-empty submission will have all non-nil, valid fields.
+    // By convention, all keys are encoded as lowercase hex strings, though it is not enforced here.
+    //
+    // INVARIANTS:
+    //  (1) either all fields are nil (empty submission) or no fields are nil
```
jordanschalm, line 147: Previously, the data model for submissions was `[String?]`, so some keys could be nil and some could be non-nil. If any keys were nil, we considered the submission invalid ("empty" in the terminology used in this PR). 

This PR changes the data model so that either all fields are nil, or none are, and enforces this in the constructor.

contracts/epochs/FlowDKG.cdc
```diff
@@ -119,6 +333,7 @@ access(all) contract FlowDKG {
 
         /// Posts a whiteboard message to the contract
         access(all) fun postMessage(_ content: String) {
+            // TODO: DKG enabled?
```
jordanschalm, line 336: It seems like the smart contract allows posting whiteboard messages and sending final submissions even when the DKG is disabled 🤔. I don't think this matters functionally because:
- the DKG cannot be ended (winning final submission retrieved) if it is disabled
- when enabled/started, all state including whiteboard messages and submissions are wiped

For clarity though, I think we should disallow interacting with the contract while it is disabled.

contracts/epochs/FlowDKG.cdc
```diff
@@ -279,36 +442,9 @@ access(all) contract FlowDKG {
         }
     }
 
-    /// Checks if two final submissions are equal by comparing each element
-    /// Each element has to be exactly the same and in the same order
-    access(account) fun submissionsEqual(_ existingSubmission: [String?], _ submission: [String?]): Bool {
```
jordanschalm, line 284: Replaced with `ResultSubmission.equals`

contracts/epochs/FlowDKG.cdc
```diff
@@ -139,57 +354,9 @@ access(all) contract FlowDKG {
         /// Sends the final key vector submission. 
         /// Can only be called by consensus nodes that are registered
         /// and can only be called once per consensus node per epoch
-        access(all) fun sendFinalSubmission(_ submission: [String?]) {
```
jordanschalm, line 142: Replaced with `SubmissionTracker` methods

contracts/epochs/FlowDKG.cdc
```diff
@@ -398,53 +522,55 @@ access(all) contract FlowDKG {
...
+            return tracker
+        }
+        // The singleton SubmissionTracker has not been created yet - create it and return a reference.
+        // This codepath should be executed at most once per FlowDKG instance and only if it was upgraded from an older version.
+        self.account.storage.save(SubmissionTracker(), to: /storage/flowDKGFinalSubmissionTracker)
+        return self.mustBorrowSubmissionTracker()
+    }
+
+    // Borrows the singleton SubmissionTracker from storage; panics if none exists.
+    access(contract) view fun mustBorrowSubmissionTracker(): &FlowDKG.SubmissionTracker {
```
jordanschalm, line 543: This can be used to read from `SubmissionTracker` in a view-only context.

lib/go/test/flow_dkg_test.go
```diff
@@ -19,10 +19,6 @@ import (
 	"github.com/onflow/flow-core-contracts/lib/go/templates"
```
jordanschalm, line 19: These tests are updated to the point where they are passing, however there are still several places (noted as TODO comments) where additional updates should be made prior to merging. In particular cases where the test is expecting a failure, but the failure occurs for the wrong reason (like passing in the old transaction arguments).

jordanschalm (commented):
Adding some notes for reviewers

contracts/epochs/FlowDKG.cdc
```diff
@@ -100,6 +90,230 @@ access(all) contract FlowDKG {
...
+    // Checks whether the ResultSubmission constructor arguments constitute a valid empty submission.
+    // A valid empty submission has all fields nil, and is used when the submittor locally failed the DKG.
+    // All non-empty submissions must have all non-nil fields.
+    access(all) view fun isValidNilSubmission(groupPubKey: String?, pubKeys: [String]?, idMapping: {String:Int}?): Bool {
+        // If any fields are nil, then this represents a empty submission and all fields must be nil
+        if groupPubKey == nil && pubKeys == nil && idMapping == nil {
+            return true
+        }
+        // Otherwise, all fields must be non-nil
+        return groupPubKey != nil && pubKeys != nil && idMapping != nil
```
joshuahannan, line 102: I'm a little confused. The comments make it seem like this function is checking if any submission is valid, but the function name is `isValidNilSubmission`, which makes me think that this should only verify that the all-`nil` case is valid. Maybe it should be renamed to `isValidSubmission`? Also, we can simplify the conditional I think:
```
        return (groupPubKey == nil && pubKeys == nil && idMapping == nil)
        || (groupPubKey != nil && pubKeys != nil && idMapping != nil)
```

contracts/epochs/FlowDKG.cdc
```diff
@@ -100,6 +90,230 @@ access(all) contract FlowDKG {
...
+             if key.length != FlowDKG.submissionKeyLength {
+                 return false
+             }
+        }
+        return true
+    }
+
+    // Checks that an id mapping (part of ResultSubmission) contains one entry per public key.
+    access(all) view fun isValidIDMapping(pubKeys: [String]?, idMapping: {String: Int}?): Bool {
+        if pubKeys == nil {
```
joshuahannan, line 133: Should this be `pubKeys == nil && idMapping == nil`?

contracts/epochs/FlowDKG.cdc
```diff
@@ -100,6 +90,230 @@ access(all) contract FlowDKG {
...
+        access(all) let pubKeys: [String]?
+        // A mapping from node ID to DKG index.
+        // There must be exactly one key per authorized DKG participant.
+        // The set of values should form the set {0, 1, 2, ... n-1}, where n is the number of
+        // authorized DKG participant, however this is not enforced here.
+        access(all) let idMapping: {String:Int}?
+
+        init(groupPubKey: String?, pubKeys: [String]?, idMapping: {String:Int}?) {
+            pre {
+                FlowDKG.isValidNilSubmission(groupPubKey: groupPubKey, pubKeys: pubKeys, idMapping: idMapping): "invalid empty submission"
```
joshuahannan, line 163:
```suggestion
                FlowDKG.isValidNilSubmission(groupPubKey: groupPubKey, pubKeys: pubKeys, idMapping: idMapping): 
                "FlowDKG.ResultSubmission.init: Invalid DKG Result Submission! "
                .concat("DKG fields must all be `nil` or all be non-`nil`")
```
I'm going through all the core contracts repos and updating the error messages to be more helpful, so it would be good to start doing this for all our PRs from now on. Basically going off the guidelines in [this issue](https://github.com/onflow/docs/issues/795). Can you do something similar for the other panics in this PR?

contracts/epochs/FlowDKG.cdc
```diff
@@ -100,6 +90,230 @@ access(all) contract FlowDKG {
...
+                self.authorized[nodeID] = true
+            }
+        }
+
+        // Adds the result submission for the DKG participant identified by nodeID.
+        // The DKG participant must be authorized for the current epoch, and each participant may submit only once.
+        // CAUTION: This method should only be called by Participant, which enforces that Participant.nodeID is passed in.
+        access(all) fun addSubmission(nodeID: String, submission: ResultSubmission) {
+            pre {
+                self.authorized[nodeID] != nil: "must be authorized for this DKG instance"
```
joshuahannan, line 255:
```suggestion
                self.authorized[nodeID] != nil: 
                "FlowDKG.addSubmission: Cannot add Result Submission for DKG for epoch "
                .concat(FlowEpoch.proposedEpochCounter().toString())
                .concat(". The node  with id ").concat(nodeID)
                .concat(" that is submitting the Result Submission is not authorized for this DKG instance."
```
Same comment about error messages here.

contracts/epochs/FlowDKG.cdc
joshuahannan: That is fine with me. Feel free to add those pre-conditions to the functions if you want!

contracts/epochs/FlowDKG.cdc
```diff
@@ -398,53 +522,55 @@ access(all) contract FlowDKG {
...
+        }
+        // The singleton SubmissionTracker has not been created yet - create it and return a reference.
+        // This codepath should be executed at most once per FlowDKG instance and only if it was upgraded from an older version.
+        self.account.storage.save(SubmissionTracker(), to: /storage/flowDKGFinalSubmissionTracker)
+        return self.mustBorrowSubmissionTracker()
+    }
+
+    // Borrows the singleton SubmissionTracker from storage; panics if none exists.
+    access(contract) view fun mustBorrowSubmissionTracker(): &FlowDKG.SubmissionTracker {
+        return self.account.storage.borrow<&SubmissionTracker>(from: /storage/flowDKGFinalSubmissionTracker)!
```
joshuahannan, line 544: Should we add a `?? panic()` here?

transactions/dkg/send_empty_final_submission.cdc
```diff
@@ -0,0 +1,20 @@
...
+// This transaction is used when the participant locally failed the DKG process, and are
+// recording this by submitting an empty result. For non-empty submissions, use "send_final_submission".
+transaction() {
+
+    let dkgParticipant: &FlowDKG.Participant
+    let submission: FlowDKG.ResultSubmission
+
+    prepare(signer: auth(BorrowValue) &Account) {
+        self.dkgParticipant = signer.storage.borrow<&FlowDKG.Participant>(from: FlowDKG.ParticipantStoragePath)
+            ?? panic("Cannot borrow dkg participant reference")
```
joshuahannan, line 13:
```suggestion
            ?? panic("Cannot borrow DKG Participant Reference from path "
                    .concat(FlowDKG.ParticipantStoragePath)
                    .concat(". The signer needs to make sure their account is initialized with the DKG Participant resource."))
```

It would be good to start updating the error messages here too. This would apply to most of the DKG transactions here too

joshuahannan (approved):
Looks pretty good! I just had a few comments. I appreciate converting some of the tests to the Cadence testing framework. Still will need to make sure all the important transactions are tested in Go though because we need to make sure the templates package is still working properly and returning valid transactions.

contracts/epochs/FlowDKG.cdc
```diff
@@ -100,6 +90,230 @@ access(all) contract FlowDKG {
...
+        }
+
+        // Returns true if this ResultSubmission instance represents an empty submission.
+        // Since the constructor enforces invariant (1), we only need to check one field.
+        access(all) view fun isEmpty(): Bool {
+            return self.groupPubKey == nil
+        }
+
+        // Checks whether the input is equivalent to this ResultSubmission.
+        // Submissions must have identical keys, in the same order, and identical key mappings.
```
tarakby, line 180:
```suggestion
        // Submissions must have identical keys, in the same order, and identical index mappings.
```

contracts/epochs/FlowDKG.cdc
jordanschalm: We could check both fields, but we're relying on Invariant (1) enforced by the constructor precondition: either all fields are nil, or none are. (see comment on subsequent line)

contracts/epochs/FlowDKG.cdc
jordanschalm: The use-case for the function is in the constructor pre-condition, so we end up passing both empty and non-empty submissions to the function -- hence the weird naming. I agree it's confusing.

I don't love `isValidSubmission`, because not all the validity checks for a submission are contained in this function (you could pass `isValidSubmission` with an invalid submission). I renamed it to refer to the invariant it is enforcing in https://github.com/onflow/flow-core-contracts/pull/441/commits/07090b8ff8a8fb8bd0dc97f350b2a7c22c22e5a1.

contracts/epochs/FlowDKG.cdc
jordanschalm: I updated all the error messages in the contract, in https://github.com/onflow/flow-core-contracts/pull/441/commits/ed047b9ea0c205bfa841857148169f595dbf163c

joshuahannan (approved):

contracts/epochs/FlowDKG.cdc
joshuahannan: good point. That is fine

lib/go/test/epoch_test_helpers.go
```diff
@@ -724,3 +741,113 @@ func expectedTargetEndTime(timingConfig cadence.Value, targetEpoch uint64) uint6
 
 	return refTimestamp + duration*(targetEpoch-refCounter)
 }
+
+func DKGPubKeyFixture() string {
```
jordanschalm, line 745: TODO: docs

joshuahannan (approved):

PR #448, opened 2024-09-11
Fixes `fillNodeRoleSlots()` to not remove from array while it is being accessed

### LINKED ISSUE #443: Fix bug in `fillNodeRoleSlots()` that removes nodes from the same list it is indexing into
### Problem

In `FlowIDTableStaking.Admin.fillNodeRoleSlots()`, there is [a loop](https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowIDTableStaking.cdc#L1198-L1203) that removes nodes from the same list it indexes to.
This can cause an error when ending the staking auction or resetting the epoch if the random index chosen is out of bounds.

### Steps to Reproduce 

TBD

### Acceptance Criteria

This may have been already fixed before and we lost the commit. Need to find out if that is somewhere and integrate it with the current code.
When running reset epoch X number of times (high number) with nodes registered but not staked, should never fail with out of bounds index.
  
### Context

This causes the reset epoch and end staking auction to fail randomly sometimes

joshuahannan:
Closes #443 

Fixes a bug that was causing the system chunk to fail when the end staking auction was executing. 

The code is in the loop in `fillNodeRoleSlots()` function that refunds candidate nodes who didn't get chosen for staking. The loop was removing items from the `candidateNodesForRole` dictionary in the same loop as the dictionary being accessed. The index that was being accessed was randomly generated, so it would sometimes cause an out of bounds error. This likelihood of the error increases as the number of candidate nodes increases.

The fix moves the code to remove from the candidate node list to after the loop has completed

Still working on a test for the change

### DIFF
--- a/lib/go/templates/idtable_staking_templates.go
+++ b/lib/go/templates/idtable_staking_templates.go
@@ -576,3 +576,17 @@ func GenerateGetApprovedNodesScript(env Environment) []byte {
 
 	return []byte(ReplaceAddresses(code, env))
 }
+
+func GenerateEndStakingTestScript(env Environment) []byte {
+	code := `
+		import FlowIDTableStaking from "FlowIDTableStaking"
+		
+		access(all) fun main() {
+			let acct = getAuthAccount<auth(BorrowValue) &Account>("FlowIDTableStaking")
+			let adminRef = acct.storage.borrow<&FlowIDTableStaking.Admin>(from: FlowIDTableStaking.StakingAdminStoragePath)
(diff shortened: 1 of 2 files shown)


janezpodhostnik (approved):

jordanschalm (approved):

sisyphusSmiling (approved):

janezpodhostnik:
@joshuahannan should this be merged now?

joshuahannan:
yes, I'll merge

PR #451, opened 2024-10-01
Updates ft and nft dependencies and basic transactions

joshuahannan:
* Updates the FT and NFT dependencies to include the latest transactions and contracts with improved error messages
* Updates the flow token transactions and basic account transactions to have better error messages


transactions/accounts/add_key.cdc
```diff
@@ -4,9 +4,20 @@ transaction(key: String, signatureAlgorithm: UInt8, hashAlgorithm: UInt8, weight
...
+                .concat(" but the options are either 1 (ECDSA_P256), 2 (ECDSA_secp256k1), or 3 (BLS_BLS12_381).")
+			hashAlgorithm >= 1 && hashAlgorithm <= 6:
+                "Cannot add Key: Must provide a hash algorithm raw value that corresponds to "
+                .concat("one of of the available hash algorithms for Flow keys.")
+                .concat("You provided ").concat(hashAlgorithm.toString())
+                .concat(" but the options are 1 (SHA2_256), 2 (SHA2_384), 3 (SHA3_256), ")
+                .concat("4 (SHA3_384), 5 (KMAC128_BLS_BLS12_381), or 6 (KECCAK_256).")
+			weight <= 1000.0:
+                "Cannot add Key: The key weight must be between 0 and 1000."
+                .concat(" You provided ").concat(weight.toString()).concat(" which is invalid.")
```
sisyphusSmiling, line 20: These updated error messages are really helpful, but I'm starting to wonder if there's a way we can standardize some of them as they're getting pretty long. Maybe something like `FlowCoreErrors` and in this instance `invalidSigAlgo(_ provided: UInt8): String` and `invalidHashAlgo(_ provided: UInt8): String`. Modifiers would be nice, but that's a whole new language feature.

transactions/accounts/revoke_key.cdc
```diff
@@ -3,7 +3,9 @@ transaction(keyIndex: Int) {
 		if let key = signer.keys.get(keyIndex: keyIndex) {
 			signer.keys.revoke(keyIndex: keyIndex)
 		} else {
-			panic("No key with the given index exists on the authorizer's account")
+			panic("Cannot revoke key: No key with the index "
+                .concat(keyIndex.toString())
+                .concat(" exists on the authorizer's account."))
```
sisyphusSmiling, line 8: np: Not sure if it's just GitHub but spacing looks off here

sisyphusSmiling (approved):

transactions/accounts/add_key.cdc
joshuahannan: It is a little awkward right now, but once we have better string formatting in Cadence, then I think it won't be as bad. I think having a separate contract for it is a little unnecessary.

PR #456, opened 2024-10-17
Add Crypto contract

turbolent:
Work towards https://github.com/onflow/cadence/issues/3135

Test case is adapted from https://github.com/onflow/flow-go-sdk/blob/7eb6328b774496418310a8487a893c070c9e5367/examples/verify_signature/user_signature/main.go.

The `Crypto` contract code and the Cadence-based tests were originally located in the Cadence codebase, and are going to be removed from there in https://github.com/onflow/cadence/pull/3619. The contract code and test code was moved as-is.


contracts/flow.json
joshuahannan, line 1: Can this be put in the main `flow.json` instead?

joshuahannan (approved):
Might be nice to include some more comments in the contract and tests

turbolent:
@joshuahannan Yes, it would be great to improve the contracts. Once we have put it on chain it should be much easier

Issue #426, opened 2024-05-07
Require proof of possession of the staking private key from staking nodes 


tarakby:
## Context

(based on the old private repo issue https://github.com/dapperlabs/flow-go/issues/6284)

Proof of possession of the staking private key (PoP) is a required info from all staking nodes, to prevent BLS rogue key attacks. 
This issue is about updating the core contract node registration to require a valid PoP in all staking requests.

## Definition of Done

* Add a new proof of possession parameter to the `registerNode` method
* Verify the PoP in the register method, if it doesn't verify properly, revert
* Update LockedTokens and StakingCollection to use this new parameter also
* Add tests for the transaction

PR #425, opened 2024-05-06
Add PoP of staking private key for node registration

tarakby:
rebase of https://github.com/onflow/flow-core-contracts/pull/301 against the `stable-cadence` branch, closes https://github.com/onflow/flow-core-contracts/issues/426

Adds a PoP parameter to the `addNodeRecord` method in the staking contract. 
In the `NodeRecord` init function, PoP is verified against the staking public key.

Also updates the locked tokens contract, staking collection, and all of their transactions to include this parameter

PR #457, opened 2024-10-17
Adds Proof of Possession for Node Staking Keys

joshuahannan:
Based on https://github.com/onflow/flow-core-contracts/pull/425, closes https://github.com/onflow/flow-core-contracts/issues/426
Moved to a new branch because of Git issues

Adds a PoP parameter to the `addNodeRecord` method in the staking contract. 
In the `NodeRecord` init function, PoP is verified against the staking public key.

Also updates the locked tokens contract, staking collection, and all of their transactions to include this parameter


transactions/stakingCollection/register_node_old.cdc
```diff
@@ -0,0 +1,52 @@
...
+            machineAccountKeySignatureAlgorithm: UInt8, 
+            machineAccountKeyHashAlgorithm: UInt8) {
+
+    let stakingCollectionRef: auth(FlowStakingCollection.CollectionOwner) &FlowStakingCollection.StakingCollection
+
+    prepare(account: auth(BorrowValue) &Account) {
+        self.stakingCollectionRef = account.storage.borrow<auth(FlowStakingCollection.CollectionOwner) &FlowStakingCollection.StakingCollection>(from: FlowStakingCollection.StakingCollectionStoragePath)
+            ?? panic(FlowStakingCollection.getCollectionMissingError(nil))
+
+        if let machineAccount = self.stakingCollectionRef.registerNode(
```
tarakby, line 23: doesn't `registerNode` expect the PoP parameter? I guess we need `registerNodeOld` no?

tarakby: That makes sense, thanks for clarifying. This old transaction would break after the contract upgrade which is expected

tarakby:
Ledger app changes will be based on the current [version](https://github.com/onflow/flow-core-contracts/pull/457/commits/ea2ba5a36920264022c8c2a52f5ef3dea92265b6) of the PR. 

This PR will remain open till we approach upgrading the contract on Mainnet.

lib/go/templates/cmd/manifest/manifest.go
```diff
@@ -829,7 +829,7 @@ func generateManifest(env templates.Environment) *manifest {
 	))
 
 	m.addTemplate(generateTemplate(
-		"SCO.17", "Register Node with PoP",
+		"SCO.17", "Register Node",
```
tarakby, line 832: A wallet would use `SCO.03` or `SCO.17` but not both at the same time. `SCO.17` is just the new "Register Node" transaction.

PR #459, opened 2024-10-21
Adds functions and tx to reclaim leases

joshuahannan:
When we did the Crescendo upgrade for LockedTokens, we made the staking object fields `access(contract)` so that they couldn't be accessed with downcasting from a reference, but this blocked a method that the foundation used to recover leases from node operators we were assisting. 

This adds a function that can be accessed from the owner of the `LockedTokenManager` to borrow a reference to the object  to be able to do staking operations. Also adds a test for it.


lib/go/test/flow_lockedtokens_test.go
```diff
@@ -627,6 +627,27 @@ func TestLockedTokensStaker(t *testing.T) {
...
+
+		signAndSubmit(
+			t, b, tx,
+			[]flow.Address{joshSharedAddress},
+			[]crypto.Signer{adminSigner},
+			false,
+		)
+
+		// Check balance of locked account
+		result := executeScriptAndCheck(t, b, templates.GenerateGetFlowBalanceScript(env), [][]byte{jsoncdc.MustEncode(cadence.Address(joshSharedAddress))})
```
Kay-Zee, line 648: shouldn't this be checking that the admin account is +10 flow?

Also, if it's easy, it'd be nice to check that the Unstaked bucket is -10 flow here as well

Kay-Zee (approved):

lib/go/test/flow_lockedtokens_test.go
joshuahannan: yes, I'll add that

joshuahannan: Added the balance check. I can't check the unstaked bucket because we're just using a dummy contract to test here, but it should be fine

joshuahannan:
@nvdtf @sisyphusSmiling Can one of you take a look here? We'd like to do this upgrade soon. Thanks!

transactions/lockedTokens/admin/recover_lease_tokens.cdc
```diff
@@ -0,0 +1,40 @@
...
+        // Get a reference to the signer's locked token manager
+        let tokenManagerRef = signer.storage.borrow<auth(FungibleToken.Withdraw, LockedTokens.RecoverLease) &LockedTokens.LockedTokenManager>(from: LockedTokens.LockedTokenManagerStoragePath)
+            ?? panic("The signer does not store a LockedTokenManager object at the path "
+                    .concat(LockedTokens.LockedTokenManagerStoragePath.toString()))
+
+        let nodeRef = tokenManagerRef.borrowNodeForLease()
+            ?? panic("Could not borrow a reference to a node in the LockedTokenManager of the signer's account")
+
+        // Withdraw enough tokens to pay for fees, assuming there are some rewards in the rewards bucket
+        tokenManagerRef.deposit(from: <-nodeRef.withdrawRewardedTokens(amount: 0.0001)!)
```
sisyphusSmiling, line 21: Is this reimbursing the fee incurred for running this transaction?

sisyphusSmiling (approved):

transactions/lockedTokens/admin/recover_lease_tokens.cdc
joshuahannan: yeah, most of these accounts have only the minimum in their main vaults so they need to be topped up a little bit to pay for the tx fee

Issue #460, opened 2024-10-22
Update the crypto algos during account creation


tarakby:
1. Looking at the[ account creation transaction](https://github.com/onflow/flow-core-contracts/blob/master/transactions/accounts/create_new_account.cdc), the algorithm checks and the error messages need to be updated to avoid confusion: although Cadence supports many signature and hash algorithms, [not all of them can be used to create a Flow account](https://developers.flow.com/build/basics/accounts#signature-and-hash-algorithms). The transaction check should be updated to include only the supported algorithms by a Flow account :
- signature: ECDSA_P256 and ECDSA_secp256k1
- hash: SHA2_256 and SHA3_256
In the current code, any other combination would pass the sanity check but the account creation would fail later.

2. Since the transaction creates a new account with a single key, the key must be a full-weight one (otherwise the new account is locked). Therefore it may be not useful to accept the key weight as an input. The transaction could hardcode the weight to 1000.

tarakby:
@joshuahannan I'm happy to make any eventual changes, but I wanted to discuss the issue above first.

joshuahannan:
Thanks for flagging. I didn't realize that. I'm fine if you want to update the error messages. You'll also need to update:
* https://github.com/onflow/flow-core-contracts/blob/master/transactions/accounts/add_key.cdc
* https://github.com/onflow/flow-core-contracts/blob/master/transactions/stakingCollection/create_machine_account.cdc
* https://github.com/onflow/flow-core-contracts/blob/master/transactions/stakingCollection/register_node.cdc

tarakby:
How about the point (2) in the issue description, I think that would add a breaking change to the transaction. Should I avoid it?

joshuahannan:
Oh yeah, sorry for missing that. I think we could remove the weight argument. We're already introducing breaking changes. Might as well make sure everything else is as we want it

PR #461, opened 2024-10-23
Update crypto algorithm checks for account keys

tarakby:
closes https://github.com/onflow/flow-core-contracts/issues/460

Account keys can only use:
 - ECDSA on P256 and secp256k1
 - sha2 and sha3 with 256 bits
 All other combinations should be rejected early by the transactions.


joshuahannan (approved):
Looks good! You'll also need to update `/lib/go/templates/cmd/mainifest/manifest.go` to remove the weight argument from the sample arguments for the create account transaction. And then this will affect the ledger transactions, so make sure this gets pulled into the feature branch as soon as you merge it

tarakby:
Good point 👌🏼  I removed the weight argument in https://github.com/onflow/flow-core-contracts/pull/461/commits/7223cd9a1c8ddf219c78ebc0180d2ea331fce3d0

PR #464, opened 2024-10-28
Add `ProtocolStateVersionUpgrade` service event

jordanschalm:
This PR adds a minimal version of a service event to upgrade the protocol state version (see also https://github.com/onflow/flow-go/pull/5428). This PR replaces https://github.com/onflow/flow-core-contracts/pull/419 and targets the EFM Recovery branch, as we intend to deploy this change alongside EFM Recovery.

## Context
The plan is to deploy this changes alongside EFM Recovery (see [branch](https://github.com/onflow/flow-core-contracts/tree/feature/efm-recovery)

## Outstanding TODOs
- [x] Alongside https://github.com/onflow/flow-go/issues/5828, remove storage-then-emit-on-heartbeat logic
- [ ] Re-enable `TestProtocolVersionUpgrade` integration test in `flow-go` before merging.


contracts/NodeVersionBeacon.cdc
```diff
@@ -280,6 +285,13 @@ access(all) contract NodeVersionBeacon {
 
             emit NodeVersionBoundaryFreezePeriodChanged(freezePeriod: newFreezePeriod)
         }
+
+        /// Emits the given protocol state version upgrade event. If the version and active are
+        /// valid, this will cause the Protocol State to upgrade its model version when the
+        /// event is incorporated.
+        access(all) fun emitProtocolStateVersionUpgrade(newProtocolVersion: UInt64, activeView: UInt64) {
```
joshuahannan, line 292: Is there any validation we can do on these inputs? Like the view needs to be in the future, version needs to be greater than the current version, etc?

joshuahannan (approved):

contracts/NodeVersionBeacon.cdc
jordanschalm: We can check that view needs to be in the future (but the actual acceptance criteria is more involved). We can't check the version is greater than current, because the current version only exists in the protocol state. 

The complete validation is done in the Protocol State Machine, where the service event is processed. Even if we partially replicate the validation here, we still need to wait for the service event to be processed at the Protocol level before we know it was successful. 

Since emitting these events is not state-changing from the smart contract perspective (you can emit as many as you want, it doesn't change any on-chain state), I'm inclined to keep all the validation only in the Protocol State machine. I'll add some documentation to explain this.

PR #467, opened 2024-11-26
Add new LinearCodeAddressGenerator contract

turbolent:
Port the linear code address generator from https://github.com/onflow/flow-go/blob/master/model/flow/address.go to Cadence.

We'll likely need this for the migration code of https://github.com/onflow/cadence/issues/3584, which will be implemented as a Cadence contract.


contracts/LinearCodeAddressGenerator.cdc
```diff
@@ -0,0 +1,112 @@
...
+    let generatorMatrixRows: [UInt64; 45]
+
+    /// Columns of the parity-check matrix H of the [64,45]-code used for Flow addresses.
+    /// H is a (n x p) matrix with coefficients in GF(2), each column is converted into
+    /// a big endian integer representation of the GF(2) column vector.
+    /// H is used to verify a code word is a valid account address.
+    access(all)
+    let parityCheckMatrixColumns: [UInt64; 64]
+
+    init() {
```
sisyphusSmiling, line 32: I see that the `LinearAddressGenerator` in the reference implementation has an [`index` field](https://github.com/onflow/flow-go/blob/master/model/flow/address.go#L145). I'm assuming the field isn't necessary here since the the `index` state isn't tracked at the execution level, but am just making sure.

contracts/LinearCodeAddressGenerator.cdc
```diff
@@ -0,0 +1,112 @@
...
+    /// Returns true if the given address is valid, for the given chain code word.
+    access(all)
+    fun isValidAddress(_ address: Address, chainCodeWord: UInt64): Bool {
+
+        let address = UInt64.fromBigEndianBytes(address.toBytes())!
+        var codeWord = chainCodeWord ^ address
+
+        if codeWord == 0 {
+            return false
+        }
```
sisyphusSmiling, line 97: Looks like what follows below is from [`isValidCodeWord`](https://github.com/onflow/flow-go/blob/master/model/flow/address.go#L302-L312), but I couldn't find the reference implementation for this block. Is there one or is this a net new check?

sisyphusSmiling (commented):
Left a couple clarifying questions. My main question is whether `index` needs to be tracked in contract or some struct's state or if that's relegated to some lower-level state.

contracts/LinearCodeAddressGenerator.cdc
```diff
@@ -0,0 +1,112 @@
...
+    access(all)
+    let codeWordMainnet: UInt64
+
+    /// The code word for addresses on Flow Testnet.
+    access(all)
+    let codeWordTestnet: UInt64
+
+    /// The code word for addresses on transient networks, like the Flow Emulator.
+    access(all)
+    let codeWordTransient: UInt64
```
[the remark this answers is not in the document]
turbolent: Good idea! Refactored in a818f5a36efdbbf257b6afaf1ae941705102ef88

contracts/LinearCodeAddressGenerator.cdc
turbolent: This Cadence function basically is a port of [`linearCodeImpl.IsValid`](https://github.com/onflow/flow-go/blob/8a3055c826a60f7a71c7a829d5d9e967c04940b0/model/flow/chain.go#L142-L152).

contracts/LinearCodeAddressGenerator.cdc
turbolent: The Cadence `LinearAddressGenerator` just implements the index <> address code, i.e. is more a port of of just [`linearCodeImpl`](https://github.com/onflow/flow-go/blob/8a3055c826a60f7a71c7a829d5d9e967c04940b0/model/flow/chain.go#L128). 

`linearCodeAddressGenerator` in the Go code is a stateful sequence/iterator, which isn't needed, but we could add a type for it to the contract once needed.

sisyphusSmiling (approved):

sisyphusSmiling:
Approved but CI failing. Looks like the go assets need to be updated as well.

tarakby:
I'm back from time off and looking at this today

contracts/LinearCodeAddressGenerator.cdc
```diff
@@ -0,0 +1,105 @@
...
+
+    /// Columns of the parity-check matrix H of the [64,45]-code used for Flow addresses.
+    /// H is a (n x p) matrix with coefficients in GF(2), each column is converted into
+    /// a big endian integer representation of the GF(2) column vector.
+    /// H is used to verify a code word is a valid account address.
+    access(all)
+    let parityCheckMatrixColumns: [UInt64; 64]
+
+    init() {
+        self.codeWords = {
```
tarakby, line 24: these are invalid code words rather than code words, maybe it's better to add the "invalid" prefix for clarity.

tests/LinearCodeAddressGenerator_test.cdc
```diff
@@ -0,0 +1,112 @@
...
+     let actual = generateAddresses(
+        count: 10,
+        chainCodeWord: chainCodeWord
+    )
+
+    Test.assertEqual(expected, actual)
+
+    for address in actual {
+        Test.assert(
+            LinearCodeAddressGenerator.isValidAddress(
```
tarakby, line 46: An `isValidAddress` implementation that always returns true would pass this test. 
It would be good to add another assertion that
```
LinearCodeAddressGenerator.isValidAddress(
   address,
   chainCodeWord: otherChainCodeWord
)
```
returns `false`. 
We could set `otherChainCodeWord` to one chain different than `chainCodeWord`.

tarakby (approved):
Nice!

contracts/LinearCodeAddressGenerator.cdc
turbolent: I found that terminology really confusing from a user-perspective. Though as far as I understand the codewords are technically invalid and mixed into the value, they are the values that should be used by the user. If a user sees "invalid" they think "should not use". Maybe we can hide this implementation detail?

tests/LinearCodeAddressGenerator_test.cdc
turbolent: Good idea! Added a test for the other chain code words in e5496c6b2dcae9769130a115b7be448878235668

contracts/LinearCodeAddressGenerator.cdc
tarakby: You're right, there is only one space of valid code words (which is the address space we use for mainnet). 
To get a different disjoint space of addresses, we get the valid code word space and mix it with an invalid value (value that is not a code word). For each invalid value, we get a new space of addresses (none of the new space addresses is a valid code word).
The fact that the value is invalid is important to get disjoint spaces of addresses, though I agree it is an implementation detail. Users should only use addresses and shouldn't care about code words.

On the Go side, code words are hidden from the package users. On Cadence, I'm not sure how the contract is going to be used. I don't think any user should access these values directly so they can be hidden (made private, non exported maybe?) to avoid confusion, same as the rest of the linear codes stuff. 
However, since the contract is using correct names for the linear code things (matrix, parity, linearity..), it makes sense IMO to keep the mention that these are invalid code words.

turbolent: I've hid all the implementation details (matrix, partity, etc.) and introduced the notion of a chain in [26ba620](https://github.com/onflow/flow-core-contracts/pull/467/commits/26ba6207e7feb0d72d8c599759ffa51eef50a5ce). So the only public API is now:

- `Chain` type
- `fun address(at index: UInt64, chain: Chain): Address`
- `fun isValidAddress(_ address: Address, chain: Chain): Bool`

contracts/LinearCodeAddressGenerator.cdc
```diff
@@ -0,0 +1,122 @@
+
+/// LinearCodeAddressGenerator allows generating and validating Flow account addresses.
+access(all)
+contract LinearCodeAddressGenerator {
+
+    access(all)
+    enum Chain: UInt8 {
+        access(all)
+	    case Mainnet
```
tarakby, line 9:
```suggestion
        case Mainnet
```

contracts/LinearCodeAddressGenerator.cdc
```diff
@@ -0,0 +1,122 @@
...
+            if word & 1 == 1 {
+                codeWord = codeWord ^ generatorMatrixRow
+            }
+            word = word >> 1
+        }
+
+        return codeWord
+    }
+
+    /// Returns the address at the given index, for the given chain code word.
```
tarakby, line 92:
```suggestion
    /// Returns the address at the given index, for the given chain.
```

contracts/LinearCodeAddressGenerator.cdc
```diff
@@ -0,0 +1,122 @@
...
+            0x42aab, 0x41951, 0x233ce, 0x22a81, 0x21948, 0x1ef60, 0x1deca, 0x1c639,
+            0x1bdd8, 0x1a535, 0x194ac, 0x18c46, 0x1632b, 0x1529b, 0x14a43, 0x13184,
+            0x12942, 0x118c1, 0x0f812, 0x0e027, 0x0d00e, 0x0c83c, 0x0b01d, 0x0a831,
+            0x0982b, 0x07034, 0x0682a, 0x05819, 0x03807, 0x007d2, 0x00727, 0x0068e,
+            0x0067c, 0x0059d, 0x004eb, 0x003b4, 0x0036a, 0x002d9, 0x001c7, 0x0003f
+        ]
+    }
+
+    access(self)
+    fun codeWord(forChain chain: Chain): UInt64 {
```
tarakby, line 61: now that this is private, can we rename to `invalidCodeWord` ?

tarakby (approved):

bluesign:
I think address(0) should fail

tarakby:
> Nit: I think address(index: 0) should fail

right thanks @bluesign. 
Actually, `address(index)` should only work for `index` in the range `[1 .. 2^45 - 1]`, and should fail otherwise.

contracts/LinearCodeAddressGenerator.cdc
turbolent: Renamed in aecc97425daa7fa118201ce0888c2b5a2319a795

turbolent:
@tarakby Added a check for the index in 7fb76ed7191ecad7ad7b100c35196e5424a5b3dc

tarakby (approved):
Sweet! Thank you @turbolent, feel free to merge

contracts/LinearCodeAddressGenerator.cdc
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        return codeWord
+    }
+
+    /// Returns the address at the given index, for the given chain.
+    access(all)
+    fun address(at index: UInt64, chain: Chain): Address? {
+        // The index must be in the range [1 .. 2^45 - 1]
+        // TODO: Use Cadence exponentiation operator, once available
```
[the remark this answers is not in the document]
turbolent: Good idea, replaced in [9ae9a32](https://github.com/onflow/flow-core-contracts/pull/467/commits/9ae9a32956848577bc5381ba1b9997d8ab4d22dc)

contracts/LinearCodeAddressGenerator.cdc
joshuahannan, line 1: Can you add a bit more docs comments to this contract to make it more clear what the purpose of each part is? Thank you!

joshuahannan (approved):

contracts/LinearCodeAddressGenerator.cdc
turbolent: The public functions have docstrings, and the private functions are implementation details. I only ported the Go code over, I'm not an expert in the details of how the underlying code works.

joshuahannan:
okay. Can you add a little blurb about it in the README with intended import addresses also like there is for other contracts? Once that is added, I'm okay with you merging it

turbolent:
@joshuahannan Added the contract to the README. I don't know where the code will be deployed, we can add a table once it is

joshuahannan:
yeah, I think the service account makes sense. I'm fine with merging this and deploying it on testnet as soon as you think it is ready.

turbolent:
@joshuahannan It's ready from my side. It would be nice to merge and close this PR, as it's been open for a while.

joshuahannan (approved):
Where is this code used? Does it need to be deployed to flow-go bootstrapping? If so, we should create a getter for it in the `lib/go/contracts` package. I can do that if needed

turbolent:
> Where is this code used? Does it need to be deployed to flow-go bootstrapping?

The code is not used anywhere. As I mentioned above, I had originally created it / ported it from the Go code for the migration of on-the-fly account storage format in flow-go, but we did not end up needing it after all.
It is not part of the bootstrapped contracts and does not need a getter, but that could be done if someone needs it.

I still think it is useful to have the logic available on-chain, as a system contract that can be trusted.

joshuahannan:
Sounds like it probably should be included then so we can deploy it as part of flow-go bootstrapping and probably deploy it to testnet and mainnet. I'll add that

PR #476, opened 2025-02-11
Use standard import placeholders

joshuahannan:
Updates all the contracts, transactions and scripts to use the standard import placeholder format `import "...."` instead of the old format.

I know there are a lot to review, but the changes are exactly the same for every file, so it shouldn't take too long to review

The `flow-evm bridge` is now using the core contracts templates package to replace import placeholders and uses the current format, so this needs to be updated to support that.

I'll work on updating `flow-go` transaction as well.

Flag: Tests are failing because `flow-go` relies on the old package that replaces addresses based on the old format. To get tests to pass, we'll need a new version of the emulator with this change.


joshuahannan:
Once I get an approval on this, I will tag a pre-release based off this branch so I can update flow-go

janezpodhostnik (approved):
I spot checked some files and it looks good to me, but I didn't go through everything. Please let me know if there are any files I need to specifically check.

PR #484, opened 2025-06-19
Implementation for FLIP 332: Stricter Validation of Node Network Addresses

vishalchangrani:
FLIP PR: https://github.com/onflow/flips/pull/333
FLIP Issue: https://github.com/onflow/flips/pull/332


contracts/FlowIDTableStaking.cdc
```diff
@@ -1822,6 +1823,67 @@ access(all) contract FlowIDTableStaking {
...
+            }
+            if isIP {
+                return false
+            }
+        }
+
+        // Check if domain has at least one dot and valid characters
+        if !domain.contains(".") {
+            return false
+        }
```
bluesign, line 1866: here better way is to check last segment is not numeric.

vishalchangrani: excellent suggestion!

vishalchangrani: I have update the changes to be more robust.

contracts/FlowIDTableStaking.cdc
```diff
@@ -166,7 +166,7 @@ access(all) contract FlowIDTableStaking {
                 FlowIDTableStaking.isValidNodeID(id): "The node ID must have only numbers and lowercase hex characters"
                 FlowIDTableStaking.nodes[id] == nil: "The ID cannot already exist in the record"
                 role >= UInt8(1) && role <= UInt8(5): "The role must be 1, 2, 3, 4, or 5"
-                networkingAddress.length > 0 && networkingAddress.length <= 510: "The networkingAddress must be less than 510 characters"
```
vishalchangrani, line 169: moved this check inside of the `isValidateNetworkingAddress` function.

contracts/FlowIDTableStaking.cdc
```diff
@@ -1822,6 +1822,91 @@ access(all) contract FlowIDTableStaking {
         return true
     }
 
+    /// Validates that a networking address is properly formatted
+    /// Requirements:
+    /// 1. Must not be an IP address
+    /// 2. Must contain a port number after a colon
+    /// 3. Must be a valid domain name format
+    access(contract) view fun isValidNetworkingAddress(address: String): Bool {
```
joshuahannan, line 1830: I feel like this function should be in a string utils contract and not in the staking contract. What do you think? We have one deployed for the bridge that we could potentially put it in

vishalchangrani: Can you point me to it? I can't find the `StringUtils` contract.

joshuahannan: https://github.com/onflow/flow-evm-bridge/blob/main/cadence/contracts/utils/StringUtils.cdc

vishalchangrani, line 1830: what do you guys think about making this `access(all)` instead `access(contract)` so that any client side apps like FlowPort can use it to verify the networking address before submitting the transaction.
The function does not modify state so should be safe to expose.

joshuahannan: yes, we should make this `access(all)`

vishalchangrani: changed it to `access(all)`.

contracts/FlowIDTableStaking.cdc
```diff
@@ -1822,6 +1822,91 @@ access(all) contract FlowIDTableStaking {
...
     }
 
+    /// Validates that a networking address is properly formatted
+    /// Requirements:
+    /// 1. Must not be an IP address
+    /// 2. Must contain a port number after a colon
+    /// 3. Must be a valid domain name format
+    access(all) view fun isValidNetworkingAddress(address: String): Bool {
+        // Check length
+       if  address.length == 0 && address.length > 510 {
```
turbolent, line 1832: This condition is always false. Did you mean
```suggestion
       if  address.length == 0 || address.length > 510 {
```

contracts/FlowIDTableStaking.cdc
```diff
@@ -1822,6 +1822,91 @@ access(all) contract FlowIDTableStaking {
...
+            return false
+        }
+
+        // Check if domain contains only letters, digits and hyphens
+        let validChars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789.-".utf8
+        for char in domain.utf8 {
+            if !validChars.contains(char) {
+                return false
+            }
+        }
```
turbolent, line 1862: This might be quite inefficient: The lookup array of valid chars is re-built on each call of the function, and the loop ends up effectively performing a nested linear scan: first over all characters of the domain, and then again nested over all valid characters.

bluesign: @turbolent  Actually I suggested this one for `readability` mostly, didn't seem to be too inefficient at first sight. ( ~ maybe half efficient ) and I assumed this would be called once ofc.

turbolent: Ah, I see now! Nice 👍 Yeah, this is definitely more readable, the original implementation is not very obvious. I don't know how often this will be called. In case it is a heavily called function, we might want to trade some readability / "straight-forwardness" code for performance.

vishalchangrani:
> @vishalchangrani Are you ready to merge this? After we merge it, I can update flow-go and we can schedule the upgrades

Do we generally first make a release of flow-go and then test it on testnet or the other way round? I was planning on testing this on testnet this week.

contracts/FlowIDTableStaking.cdc
vishalchangrani: This validation will only be called when a new node is registered or when the networking address of an existing node is update. Both of those actions don't happen frequently - once or twice every couple of epoch at most. Hence, I would rather have more readability than performance. Open to suggestions tho.

vishalchangrani:
> @vishalchangrani We try to merge right before we upgrade on testnet and create a release for flow-go before we upgrade on mainnet.

I plan to upgrade on testnet by EOD tomorrow. Will merge this in before that.

PR #486, opened 2025-07-03
Address PR comments

joshuahannan:
Addresses the PR comments from https://github.com/onflow/flow-core-contracts/pull/485


contracts/FlowCallbackScheduler.cdc
```diff
@@ -348,36 +368,26 @@ access(all) contract FlowCallbackScheduler {
...
-            }
-
-            if executionEffort > self.priorityEffortLimit[priority]! {
-                return EstimatedCallback(flowFee: nil, timestamp: nil, error: "Invalid execution effort: greater than available effort for priority")
-            }
-
-            if executionEffort < self.minimumExecutionEffort {
-                return EstimatedCallback(flowFee: nil, timestamp: nil, error: "Invalid execution effort: less than minimum execution effort")
-            }
+            self.checkScheduleArgs(timestamp: timestamp, priority: priority, executionEffort: executionEffort)
```
devbugging, line 390: I believe this means estimate would panic in case of invalid args which I don't think is good idea, since estimate might be used to test args and might fail whole transaction. I intentionally didn't want to panic, but let me know your thinking.

PR #487, opened 2025-07-08
Add correct fee calculation and rearrange schedule conditions

joshuahannan:
* Adds the call to `FlowFees` for proper fee calculations. Emulator fees currently do not include any execution effort cost, so this changed the fee calculations to all be multiples of `0.00001`.
* Updated `schedule()` to call `estimate()` to validate the arguments. This makes it so we don't have to write separate unit tests for the same input verification as `estimate()`. As part of this, I updated estimate to not panic on a low priority estimate because it didn't seem necessary to me. 
* Updated error messages to include values


contracts/FlowCallbackScheduler.cdc
```diff
@@ -375,24 +362,21 @@ access(all) contract FlowCallbackScheduler {
             priority: Priority,
             executionEffort: UInt64
         ): EstimatedCallback? {
```
devbugging, line 364: this shouldn't have to be an optional anymore since EstimateCallback now has error field

contracts/FlowCallbackScheduler.cdc
```diff
@@ -490,38 +474,39 @@ access(all) contract FlowCallbackScheduler {
...
-            return val < 0 ? 0 : val
-        }
-
         /// calculate fee by converting execution effort to a fee in Flow tokens.
-        access(self) view fun calculateFee(executionEffort: UInt64, priority: Priority): UFix64 {
-            // todo: change to match the actual fee calculation
-            let baseFee = UFix64(executionEffort) / 10000.0
+        access(all) view fun calculateFee(executionEffort: UInt64, priority: Priority): UFix64 {
+            // Use the official FlowFees calculation
+            let baseFee = FlowFees.computeFees(inclusionEffort: 1.0, executionEffort: UFix64(executionEffort))
```
devbugging, line 486: 👌

contracts/FlowCallbackScheduler.cdc
```diff
@@ -124,26 +124,24 @@ access(all) contract FlowCallbackScheduler {
...
-            }
-
-            if newStatus == Status.Executed && self.status != Status.Processed {
-                panic("Invalid status: Callback with id \(self.id) cannot be marked as Processed until after it is Scheduled")
-            }
-
-            if newStatus == Status.Processed && self.status != Status.Scheduled {
-                panic("Invalid status: Callback with id \(self.id) cannot be marked as Processed until after it is Scheduled")
+            pre {
+                self.status != Status.Executed && self.status != Status.Canceled:
```
devbugging, line 128: 😍

contracts/FlowCallbackScheduler.cdc
```diff
@@ -375,24 +362,21 @@ access(all) contract FlowCallbackScheduler {
             priority: Priority,
             executionEffort: UInt64
         ): EstimatedCallback? {
-            /// low priority callbacks are not supported
```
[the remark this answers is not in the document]
joshuahannan: The `EstimatedCallback` struct it returns is still mostly useless because it has the the special timestamp value, so we can just leave it up to the caller if they want to panic or not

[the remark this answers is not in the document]
joshuahannan: okay. I updated it in [77b264d](https://github.com/onflow/flow-core-contracts/pull/488/commits/77b264d081203c360afa40a4daf7b1d500b7737e) so that it still returns an error on low priority but wrapped the error checking in schedule to ignore that.

Does that look good?

PR #488, opened 2025-07-09
Adds additional schedule tests

joshuahannan:
* Update `CallbackScheduled` to have a new parameter
* Add error case for when a high priority callback doesn't fit into a timestamp
* Save the slot used amount back to the dictionary
* Add a getter for the slot available effort to help with testing
* Update the callback handler contract to be able to handle more than one callback
* Add more scripts to help with testing
* Add more scheduling tests


contracts/FlowCallbackScheduler.cdc
```diff
@@ -383,6 +390,10 @@ access(all) contract FlowCallbackScheduler {
                 executionEffort: executionEffort
             )
 
+            if scheduledTimestamp == nil {
```
devbugging, line 393: This should only happen if priority is high, for medium and low priority we always scheduled a timestamp.

tests/CallbackScheduler_test.cdc
```diff
@@ -64,10 +87,100 @@ access(all) fun testCallbackSchedulingAndExecution() {
...
+    ).returnValue! as! UInt8
+    Test.assertEqual(UInt8(0), status!)
+
+    // Schedule another callback, medium this time
+    tx = Test.Transaction(
+        code: Test.readFile("../transactions/callbackScheduler/schedule_callback.cdc"),
+        authorizers: [serviceAccount.address],
+        signers: [serviceAccount],
+        arguments: [futureTime, feeAmount, mediumEffort, mediumPriority, testData],
+    )
```
devbugging, line 103: maybe good idea to extract this into helper function `scheduleCallback(timestamp, fee, effort, priority, data)`

joshuahannan: yeah, I think I'll do that

contracts/FlowCallbackScheduler.cdc
```diff
@@ -326,7 +326,7 @@ access(all) contract FlowCallbackScheduler {
                 executionEffort: executionEffort
             )
 
-            if estimate.error != nil {
+            if estimate.error != nil && estimate.timestamp == nil {
```
devbugging, line 329: I think this will bite someone in the ass down the line, but I'm fine with it if you think it's ok. I would just at least add a comment explaining why we do this here.

joshuahannan: okay I'll add that

PR #489, opened 2025-07-09
FEATURE: Transaction scheduling

joshuahannan:
Implements the contracts and tests for the [Scheduled Callbacks FLIP](https://github.com/onflow/flips/pull/331#)


contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,946 @@
...
+
+    /// Interfaces
+
+    /// The callback handler is an interface that defines a single method executeCallback that 
+    /// must be implemented by the contract or resource that would like to schedule the callback. 
+    /// The callback gets executed by the scheduler contract by calling the handler provided to 
+    /// schedule function with Callback entitlement. The arguments are:
+    /// - ID of the scheduled callback (this can be useful for any internal tracking)
+    /// - The data that was passed in during the schedule call
+    access(all) struct interface CallbackHandler {
```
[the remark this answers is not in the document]
joshuahannan: yep, I reported it to the Cadence team and I have updated it in [my latest PR](https://github.com/onflow/flow-core-contracts/pull/500)

contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,946 @@
...
+        priority: UInt8,
+        callbackOwner: Address,
+        fees: UFix64
+    )
+
+    access(all) event CallbackCanceled(
+        id: UInt64,
+        priority: UInt8,
+        callbackOwner: Address
+    )
```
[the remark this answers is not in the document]
joshuahannan: Good point, I'll update those event names

contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,946 @@
...
+    access(all) struct ScheduledCallback {
+        access(self) let scheduler: Capability<auth(CancelCallback) &SharedScheduler>
+        access(all) let id: UInt64
+        access(all) let timestamp: UFix64?
+
+        access(all) view fun status(): Status? {
+            return self.scheduler.borrow()!.getStatus(id: self.id)
+        }
+
+        access(contract) init(
```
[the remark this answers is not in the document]
joshuahannan: I have not tried that but that is a good idea. I'll test it out and report if is possible

contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,946 @@
...
+
+    access(all) fun getSchedulerConfigMetadata(): SchedulerConfig {
+        return self.sharedScheduler.borrow()!.getConfigMetadata()
+    }
+    
+    /// getSizeofData takes a callback's data
+    /// argument and stores it in the contract account's storage, 
+    /// checking storage used before and after to see how large the data is in MB
+    /// If data is nil, the function returns 0.0
+    access(all) fun getSizeofData(_ data: AnyStruct?): UFix64 {
```
sisyphusSmiling, line 919:
```suggestion
    access(all) fun getSizeOfData(_ data: AnyStruct?): UFix64 {
```

contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,946 @@
...
+            switch status {
+                case Status.Executed:
+                    emit CallbackExecuted(
+                        id: callback.id,
+                        priority: callback.priority.rawValue,
+                        callbackOwner: callback.handler.address,
+                        fees: callback.fees.balance
+                    )
+                    // Deposit all the fees into the FlowFees vault
+                    destroy callback.payAndWithdrawFees(multiplierToWithdraw: 0.0)
```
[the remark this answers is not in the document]
joshuahannan: Since `multiplierToWithdraw` is `0.0` here, the amount destroyed here should always be zero because in `payAndWithdrawFees`, we just deposit all the fees not withdrawn to `FlowFees`. We just need to destroy here to make sure there is no loss of resource error.

contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,946 @@
...
+        access(all) let timestamp: UFix64?
+
+        access(all) view fun status(): Status? {
+            return self.scheduler.borrow()!.getStatus(id: self.id)
+        }
+
+        access(contract) init(
+            scheduler: Capability<auth(CancelCallback) &SharedScheduler>,
+            id: UInt64, 
+            timestamp: UFix64?
```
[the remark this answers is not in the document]
joshuahannan: seems like we could change it to not be optional. @devbugging any reason why we shouldn't make this change?

contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,946 @@
...
+        /// calculate fee by converting execution effort to a fee in Flow tokens.
+        access(all) fun calculateFee(executionEffort: UInt64, priority: Priority, data: AnyStruct?): UFix64 {
+            // Use the official FlowFees calculation
+            let baseFee = FlowFees.computeFees(inclusionEffort: 1.0, executionEffort: UFix64(executionEffort))
+            
+            // Scale the execution fee by the multiplier for the priority
+            let scaledExecutionFee = baseFee * self.priorityFeeMultipliers[priority]!
+
+            // Calculate the FLOW required to pay for storage of the callback data
+            let storageFee = FlowStorageFees.storageCapacityToFlow(FlowCallbackScheduler.getSizeofData(data))
```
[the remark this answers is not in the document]
joshuahannan: Considering that all of the callbacks already have to pay at least 2x the regular fee for their transactions, that covers the cost of the internal data more than enough

contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,946 @@
...
+            for id in historicStatuses {
+                let historic = self.historicStatuses[id]
+                if historic!.timestamp < currentTimestamp - self.historicStatusLimit {
+                    self.historicStatuses.remove(key: id)
+                }
+            }
+        }
+
+        /// cancel scheduled callback and return a portion of the fees that were paid.
+        access(CancelCallback) fun cancel(id: UInt64): @FlowToken.Vault {
```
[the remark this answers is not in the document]
joshuahannan: yeah, we basically want to avoid anyone creating the `ScheduledCallback` struct on their own and having the entitlement there makes that a bit safer

contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,946 @@
...
+        /// priority effort limit is the maximum effort per priority in a timeslot
+        access(contract) var priorityEffortLimit: {Priority: UInt64}
+
+        /// minimum execution effort is the minimum effort that can be 
+        /// used for any priority
+        access(contract) var minimumExecutionEffort: UInt64
+
+        /// priority fee multipliers are values we use to calculate the added 
+        /// processing fee for each priority
+        access(contract) var priorityFeeMultipliers: {Priority: UFix64}
```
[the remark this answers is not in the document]
joshuahannan: @devbugging what do you think?

contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,946 @@
...
+            self.error = error
+        }
+    }
+
+    /// Callback data is internal representation of a scheduled callback which contains all the functionality 
+    /// to schedule, process and execute each callback. 
+    access(all) resource CallbackData {
+        access(all) let id: UInt64
+        access(all) let priority: Priority
+        access(all) let executionEffort: UInt64
```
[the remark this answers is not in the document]
joshuahannan: I could see it going both ways though. Yes, technically it is the max execution effort for that specific transaction, but it is also the size of the execution effort chunk for the given timestamp in which case the `max` prefix isn't really accurate

contracts/FlowCallbackScheduler.cdc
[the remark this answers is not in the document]
devbugging: Yes, great idea. I'm pretty sure we will do more inteligent fee calculations, I had that in mind already. But for now lets keep it simple but implement it in a way that allows for flexibility.

contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,946 @@
...
+        ///    how soon after the timestamp the callback will be executed.
+        /// @param: executionEffort: Defines the maximum computational resources allocated to the callback. This also determines 
+        ///    the fee charged. Unused execution effort is not refunded.
+        /// @param: fees: A Vault resource containing sufficient funds to cover the required execution effort.
+        access(contract) fun schedule(
+            callback: Capability<auth(ExecuteCallback) &{CallbackHandler}>,
+            data: AnyStruct?,
+            timestamp: UFix64,
+            priority: Priority,
+            executionEffort: UInt64,
```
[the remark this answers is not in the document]
sisyphusSmiling: > it is a bit different to gasLimit, because you will always pay for what you set in the effort, no matter the actual execution.

Valid point @devbugging. I don't feel very strongly about the change, especially since the code is commented. Another alternative might be `estimatedEffort`, but I don't have strong convictions either way.

contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,946 @@
...
+
+            self.status = newStatus
+        }
+
+        /// payAndWithdrawFees withdraws fees from the callback based on the refund multiplier.
+        /// This action is only allowed for canceled callbacks, otherwise it panics.
+        /// It deposits any leftover fees to the FlowFees vault to be used to pay node operator rewards
+        access(contract) fun payAndWithdrawFees(multiplierToWithdraw: UFix64): @FlowToken.Vault {
+            let amount = self.fees.balance * multiplierToWithdraw
+            let feesToReturn <- self.fees.withdraw(amount: amount) as! @FlowToken.Vault
```
sisyphusSmiling, line 179: Since the withdraw amount could be `0.0` when `multiplierToWithdraw == 0.0`, what do you think about calling `createEmptyVault` to avoid the extra `FungibleToken.Withdraw` event emission?

joshuahannan: That is a good idea. I'll do that

contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,946 @@
...
+        access(contract) view fun calculateScheduledTimestamp(
+            timestamp: UFix64, 
+            priority: Priority, 
+            executionEffort: UInt64
+        ): UFix64? {
+            if priority == Priority.Low {
+                return self.lowPriorityScheduledTimestamp
+            }
+
+            let used = self.slotUsedEffort[timestamp]
```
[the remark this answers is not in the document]
sisyphusSmiling: Maybe I'm missing it, but I don't see where the timestamp is sanitized.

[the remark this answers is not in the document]
joshuahannan: Almost all my changes based on your comments are in these two PRs:
https://github.com/onflow/flow-core-contracts/pull/500
https://github.com/onflow/flow-core-contracts/pull/501

janezpodhostnik (commented):
Great  work so far. I had some comments. Not sure if they were already addressed in the other PRS

contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,1009 @@
...
+                    "Invalid status: Callback with id \(self.id) can only be set as Processed if it is Scheduled"
+            }
+
+            self.status = newStatus
+        }
+
+        /// payAndWithdrawFees withdraws fees from the callback based on the refund multiplier.
+        /// This action is only allowed for canceled callbacks, otherwise it panics.
+        /// It deposits any leftover fees to the FlowFees vault to be used to pay node operator rewards
+        access(contract) fun payAndWithdrawFees(multiplierToWithdraw: UFix64): @FlowToken.Vault {
```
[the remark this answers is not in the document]
joshuahannan: good suggestion! I'll make that change

contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,1150 @@
...
+                // todo: This could get pretty costly if there are a lot of low priority callbacks
+                // in the queue. Figure out how to more efficiently go through the low priority callbacks
+                var lowPriorityEffortAvailable = self.getSlotAvailableEffort(timestamp: timestamp, priority: Priority.Low)
+                if lowPriorityEffortAvailable > 0 {
+                    for lowCallbackID in lowPriorityCallbacks.keys {
+                        let callbackEffort = lowPriorityCallbacks[lowCallbackID]!
+                        if callbackEffort <= lowPriorityEffortAvailable {
+                            lowPriorityEffortAvailable = lowPriorityEffortAvailable - callbackEffort
+                            lowPriorityCallbacks.remove(key: lowCallbackID)
+                            sortedCallbackIDs.append(lowCallbackID)
```
oebeling, line 895: The FLIP says that a low priority callback "is executed opportunistically, **only in blocks after the scheduled timestamp**".

However, it seems this code may process low-priority callbacks before they are due - `lowPriorityCallbacks` is a collection of all low-priority callbacks as they are all stored at the special timeslot `0`.

joshuahannan: oh man, I can't believe I didn't have a test for that. Good catch!

contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,1150 @@
...
+    /// checking storage used before and after to see how large the data is in MB
+    /// If data is nil, the function returns 0.0
+    access(all) fun getSizeOfData(_ data: AnyStruct?): UFix64 {
+        if data == nil {
+            return 0.0
+        } else {
+            let type = data!.getType()
+            if type.isSubtype(of: Type<Number>()) 
+            || type.isSubtype(of: Type<Bool>()) 
+            || type.isSubtype(of: Type<Path>())
```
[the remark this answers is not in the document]
joshuahannan: good point. I didn't consider that! I'll remove `Path` from the list

contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,1150 @@
...
+                    Priority.High: highPriorityEffortReserve,
+                    Priority.Medium: mediumPriorityEffortReserve,
+                    Priority.Low: 0
+                },
+                priorityEffortLimit: {
+                    Priority.High: highPriorityEffortReserve + sharedEffortLimit,
+                    Priority.Medium: mediumPriorityEffortReserve + sharedEffortLimit,
+                    Priority.Low: 5_000
+                },
+                minimumExecutionEffort: 5,
```
oebeling, line 506: nit: FLIP says "The provided execution effort value must be bigger than **10**". Even though this is configurable, would IMO make sense to align the default value and the value mentioned in the FLIP.

contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,1009 @@
...
+
+            let callbackID = self.getNextIDAndIncrement()
+            let callback <- create CallbackData(
+                id: callbackID,
+                handler: callback,
+                data: data,
+                originalTimestamp: sanitizedTimestamp,
+                priority: priority,
+                executionEffort: executionEffort,
+                fees: <- fees,
```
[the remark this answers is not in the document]
bluesign: yeah I think adding ( writing ) some bytes ( considering it is small ) to a large dictionary ( where we may have hypothetical problem ) will be more cheaper in storage fees vs computation required.

contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,1009 @@
...
+            if executionEffort <= available {
+                return timestamp
+            } else if priority == Priority.High {
+                // high priority demands scheduling at exact timestamp or failing
+                return nil
+            }
+
+            // if there is no space left for medium priority we search for next available timestamp
+            // todo: check how big the callstack can grow and if we should avoid recursion
+            return self.calculateScheduledTimestamp(
```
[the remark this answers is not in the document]
bluesign: I think stack depth is 2000, cost to authorizer is negligible ( though to system it can have heavier cost )

contracts/FlowCallbackScheduler.cdc
joshuahannan: good point. We'll update that

joshuahannan: thanks, I'll update that

contracts/FlowCallbackScheduler.cdc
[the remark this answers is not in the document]
janezpodhostnik: With recursion the memory cost is a lot higher which doesn't really translate to fees, but could halt the transaction. the execution cost is also higher, but not by that much, I cant really give you any concrete numbers at this time.

If you feel its cleaner this way we can keep it if the stack isn't going to be too deep most of the time.

contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,1410 @@
...
+                executionEffort: callback.executionEffort,
+                fees: callback.fees,
+                callbackOwner: callback.handler.address,
+                callbackHandlerTypeIdentifier: callbackHandler.getType().identifier,
+                callbackName: callbackHandler.getName(),
+                callbackDescription: callbackHandler.getDescription()
+            )
+
+            // Add the callback to the slot queue and update the internal state
+            self.addCallback(slot: estimate.timestamp!, callback: callback)
```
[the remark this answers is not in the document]
joshuahannan: Good catch, but can't we just make getName and getDescription `view` and that also solves it?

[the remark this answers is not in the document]
bluesign: @joshuahannan  yeah you are right, it is much better.

contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,1410 @@
...
+                    ?? panic("Invalid callback handler: Could not borrow a reference to the callback handler")
+
+                emit PendingExecution(
+                    id: callback.id,
+                    priority: callback.priority.rawValue,
+                    executionEffort: callback.executionEffort,
+                    fees: callback.fees,
+                    callbackOwner: callback.handler.address,
+                    callbackHandlerTypeIdentifier: callbackHandler.getType().identifier,
+                    callbackName: callbackHandler.getName(),
```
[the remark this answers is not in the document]
joshuahannan: yes, you are right. Great catch. I think the two options are either to make them fields instead of functions or to just remove them from the `PendingExecution` event. 

@bjartek What do you think is the best solution here?

contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,1410 @@
...
+        }
+    }
+
+    /// Structs
+
+    /// ScheduledCallback is the struct that the user receives after scheduling a callback.
+    /// It allows them to get the status of their callback and can be passed back
+    /// to the scheduler contract to cancel the callback if it has not yet been executed. 
+    /// It can only be created by the scheduler contract to prevent spoofing.
+    access(all) struct ScheduledCallback {
```
joshuahannan, line 144: I'm curious to hear others thoughts here:

The devrel team created a draft of a [cron job scheduler contract](https://github.com/onflow/scheduledcallbacks-scaffold/blob/mfbz/cron-job-contract/cadence/contracts/CronJobScheduler.cdc) that builds on top of the callbacks contract. Obviously it needs a lot of work anyway, but I noticed that when they schedule a callback, [they don't save the `ScheduledCallback` struct and it just gets lost](https://github.com/onflow/scheduledcallbacks-scaffold/blob/mfbz/cron-job-contract/cadence/contracts/CronJobScheduler.cdc#L157-L164). 

This is a problem because the only way for them to cancel a callback is to have a copy of the ScheduledCallback struct. The fact that the first people who built something with scheduled callbacks made this mistake makes me think that it will probably happen a lot if we don't do anything about it.

Originally we made it a struct because it doesn't really have monetary value and we didn't think the functionality it enabled was necessarily privileged enough to require the awkwardness of dealing with resource syntax and rules.

I'm starting to think that maybe we should change it to a resource instead. In a sense, it does have monetary value because of the fees that are stored in the contract that would be returned if you canceled the callback, and the canceling functionality is also somewhat important to maintain, though I can understand the argument that it might not be a big deal if they lose the ability to cancel it, but I'm not sure.

I'm leaning towards changing the type to a resource to make sure people are forced to store it somewhere or explicitly delete it. Does anyone else have any thoughts?

@bluesign @sisyphusSmiling @bjartek @devbugging @janezpodhostnik

contracts/FlowCallbackScheduler.cdc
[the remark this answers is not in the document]
bjartek: I would rather have it as a field then removing them.

contracts/FlowCallbackScheduler.cdc
bjartek: Have we considered storing a scheduleManager in a users storage and have the schedule method on that? Then you can store it in that I belive?

Otherwise making it a resource makes sense

joshuahannan: I think that is a great idea. We can just make the scheduleManager a resource and keep the ScheduledCallback a struct since they don't ever have to leave the manager resource.

joshuahannan: Should we only allow scheduling callbacks through the schedule manager resource or should we allow people to schedule them directly on the contract also to get a `ScheduledCallback` object back?

bluesign: I am not sure "cancel" is that much needed in general. ( for example in cron context ) 

I am on mobile didn't check the cron contract. But eventually, you would want some "inception" style logic there. 

if I have something running every day at midnight;

- I would have something that schedules to midnight, - then at midnight it schedules the work as a callback
- and then schedules itself once again to following midnight. 

so it can have it own cancel somehow there. ( without using cancel feature here )

bjartek: Making it a resource ensures that it is saved atleast does it not.

But where is it saved? I belive there should be a default place for this. In order to create good tools on top of the spec this is important.

I have not had the time/energy to look at cronjob. However i belive that if the manager lives in the same account se the callback owner you can just borrow it in there can you not?

contracts/FlowCallbackScheduler.cdc
```diff
@@ -0,0 +1,1450 @@
...
+
+            let pendingCallbacks = self.pendingQueue()
+            
+            if pendingCallbacks.length == 0 {
+                return
+            }
+
+            for callback in pendingCallbacks {
+                let callbackHandler = callback.handler.borrow()
+                    ?? panic("Invalid callback handler: Could not borrow a reference to the callback handler")
```
oebeling, line 1228: On the topic of panics that can take down the feature this also seems like a straightforward way to kill the feature. Just revoke the capability or remove the underlying object from storage and we'll crash here and `process()` will revert. 

I think it's likely this could even happen by accident at some point as people move stuff around their storage or clear capabilities they think are unused.

oebeling: I think it fixes this issue - noted a Cadence documentation issue there. You might want to consider a similar thing for `cancel()` if you want to allow cancellation of potentially broken callbacks - though that's way lower priority than bugs like this that could have taken down the whole feature.

contracts/testContracts/TestFlowCallbackHandler.cdc
```diff
@@ -0,0 +1,107 @@
...
+            panic("Invalid status for callback with id \(callback.id)")
+        }
+        self.scheduledCallbacks[callback.id] <-! callback
+    }
+
+    access(all) fun cancelCallback(id: UInt64): @FlowToken.Vault {
+        let callback <- self.scheduledCallbacks.remove(key: id)
+            ?? panic("Invalid ID: \(id) callback not found")
+        return <-FlowCallbackScheduler.cancel(callback: <-callback!)
+    }
```
sisyphusSmiling, line 86: I think comments related to access would be helpful here. In general, a contract dev wouldn't want to expose these methods as public

contracts/FlowTransactionScheduler.cdc
```diff
@@ -0,0 +1,1440 @@
...
+                return timestamp
+            } else if priority == Priority.High {
+                // high priority demands scheduling at exact timestamp or failing
+                return nil
+            }
+
+            // if there is no space left for medium or low priority we search for next available timestamp
+            // todo: check how big the callstack can grow and if we should avoid recursion
+            // todo: we should refactor this into loops, because we could need to recurse 100s of times
+            return self.calculateScheduledTimestamp(
```
devbugging, line 891: We should convert this into a while loop. The next available timestamp can be far in the future, and this might cause a stack overflow.

contracts/FlowTransactionScheduler.cdc
```diff
@@ -0,0 +1,1440 @@
...
+
+    access(all) fun getSchedulerConfigurationDetails(): {SchedulerConfig} {
+        return self.sharedScheduler.borrow()!.getConfigurationDetails()
+    }
+    
+    /// getSizeOfData takes a transaction's data
+    /// argument and stores it in the contract account's storage, 
+    /// checking storage used before and after to see how large the data is in MB
+    /// If data is nil, the function returns 0.0
+    access(all) fun getSizeOfData(_ data: AnyStruct?): UFix64 {
```
devbugging, line 1407: This function can dissable operation of scheduling if the account balance gets too low. We should probably add a comment at the top of the contract that the account holding the resource should have at least X Flow (calculate X) to operate.

contracts/FlowTransactionScheduler.cdc
```diff
@@ -0,0 +1,1440 @@
...
+
+            emit Executed(
+                id: tx.id,
+                priority: tx.priority.rawValue,
+                executionEffort: tx.executionEffort,
+                transactionHandlerOwner: tx.handler.address,
+                transactionHandlerTypeIdentifier: transactionHandler.getType().identifier
+            )
+            
+            transactionHandler.executeTransaction(id: id, data: tx.getData())
```
devbugging, line 1312: nit: I feel with the current naming this should be just `execute()` but I don't feel strongly about it. I just think this is actually not transaction on the contract call level.

contracts/FlowTransactionScheduler.cdc
```diff
@@ -0,0 +1,1440 @@
...
+            } else {
+                let amountToReturn = self.fees * refundMultiplier
+                let amountToKeep = self.fees - amountToReturn
+                let feesToReturn <- FlowTransactionScheduler.withdrawFees(amount: amountToReturn)
+                FlowFees.deposit(from: <-FlowTransactionScheduler.withdrawFees(amount: amountToKeep))
+                return <-feesToReturn
+            }
+        }
+
+        /// getData copies and returns the data field
```
devbugging, line 267: nit: `getData returns the data field`

devbugging, line 267:
```suggestion
        /// getData returns the data field
```

contracts/FlowTransactionScheduler.cdc
```diff
@@ -0,0 +1,1440 @@
...
+
+        /// schedule is the primary entry point for scheduling a new transaction within the scheduler contract. 
+        /// If scheduling the transaction is not possible either due to invalid arguments or due to 
+        /// unavailable slots, the function panics. 
+        //
+        /// The schedule function accepts the following arguments:
+        /// @param: transaction: A capability to a resource in storage that implements the transaction handler 
+        ///    interface. This handler will be invoked at execution time and will receive the specified data payload.
+        /// @param: timestamp: Specifies the earliest block timestamp at which the transaction is eligible for execution 
+        ///    (fractional seconds values are ignored). It must be set in the future.
```
devbugging, line 695:
```suggestion
        ///    (Unix timestamp so fractional seconds values are ignored). It must be set in the future.
```

contracts/FlowTransactionScheduler.cdc
joshuahannan: Why would it disable scheduling? It saves and loads, so after the function is done executing, the storage used is not any different. plus, any fees that are used in scheduling are deposited to the contract account's vault, so that would cover any storage fees anyway

contracts/FlowTransactionScheduler.cdc
joshuahannan: okay, I can update this to `execute()`

joshuahannan: Actually, it seems `execute` is a keyword, so we can't use it

contracts/FlowTransactionScheduler.cdc
```diff
@@ -0,0 +1,1440 @@
...
+
+                // if the slot is now empty remove it from the maps
+                if transactionQueue.keys.length == 0 {
+                    self.slotQueue.remove(key: slot)
+                    self.slotUsedEffort.remove(key: slot)
+
+                    self.sortedTimestamps.remove(timestamp: slot)
+                }
+            }
+
```
devbugging, line 1090: better to extract into a function `garbageCollectSlot(slot)`

contracts/FlowTransactionScheduler.cdc
```diff
@@ -0,0 +1,1440 @@
...
+                        let tx = self.borrowTransaction(id: id)
+                            ?? panic("Invalid ID: \(id) transaction not found during initial processing") // critical bug
+
+                        // Only add scheduled transactions to the queue
+                        if tx.status != Status.Scheduled {
+                            continue
+                        }
+
+                        // this is safeguard to prevent collection growing too large in case of block production slowdown
+                        if totalAvailableEffort.saturatingSubtract(tx.executionEffort) == 0 || totalTransactionsLimit == 0 {
```
devbugging, line 1135:
```suggestion
                       if tx.executionEffort > totalAvailableEffort || totalTransactionsLimit == 0 {
```

contracts/FlowTransactionScheduler.cdc
```diff
@@ -0,0 +1,1440 @@
...
+            minimumExecutionEffort: UInt64,
+            maxDataSizeMB: UFix64,
+            priorityFeeMultipliers: {Priority: UFix64},
+            refundMultiplier: UFix64,
+            canceledTransactionsLimit: UInt,
+            collectionEffortLimit: UInt64,
+            collectionTransactionsLimit: Int
+        ) {
+            pre {
+                refundMultiplier >= 0.0 && refundMultiplier <= 1.0:
```
devbugging, line 335: I believe that alowing the refundMultiplier be set to 1.0 will cause issues in the refund function, since it will try to withdraw 0

joshuahannan: withdrawing zero is allowed, so I don't think this is a problem

tests/transactionScheduler_events_test.cdc
```diff
@@ -0,0 +1,261 @@
...
+
+    // move time forward to trigger execution eligibility
+    // Have to subtract to handle the automatic timestamp drift
+    // so that the medium transaction that got scheduled doesn't get marked as pendingExecution
+    Test.moveTime(by: Fix64(futureDelta - 6.0))
+    while getTimestamp() < timeInFuture {
+        Test.moveTime(by: Fix64(1.0))
+    }
+
+    // Simulate FVM process - should process since timestamp is in the past
```
devbugging, line 197: it would be nice to add a comment which transaction it would process since we don't schedule anything in this test, I guess it should be something like:

```
// process transaction scheduled in the testTransactionScheduleEventAndData 
```

tests/transactionScheduler_events_test.cdc
```diff
@@ -0,0 +1,261 @@
...
+                Test.assertEqual(pendingExecutionEvent.id, executedEvent.id)
+                Test.assertEqual(pendingExecutionEvent.priority, executedEvent.priority)
+                Test.assertEqual(pendingExecutionEvent.executionEffort, executedEvent.executionEffort)
+                Test.assertEqual(pendingExecutionEvent.transactionHandlerOwner, executedEvent.transactionHandlerOwner)
+                Test.assertEqual(pendingExecutionEvent.transactionHandlerTypeIdentifier, executedEvent.transactionHandlerTypeIdentifier!)
+                firstEvent = true
+            }
+        }
+
+        i = i + 1
```
devbugging, line 240: why do we need this? I think it would be wise to check i == 1 (I believe)
Epsecially it would be very good to check the ID of transaction to be executed and match it to the ID from the above schedule transaction, because this is critical, if these IDs for some reason get mixed it will cause critical failure.

tests/transactionScheduler_misc_test.cdc
```diff
@@ -0,0 +1,303 @@
...
+        if id != 2 {
+            cancelTransaction(
+                id: id,
+                failWithErr: nil
+            )
+        }
+
+        // make sure the getStatus logic is working correctly
+        // the second transaction should be canceled
+        if id < 12 {
```
devbugging, line 100: I just assume this is the cancel transactions limit value, maybe good to make it a const

tests/transactionScheduler_misc_test.cdc
```diff
@@ -0,0 +1,303 @@
...
+    )
+
+    // get the status of the newly scheduled transaction with ID 2
+    var status = getStatus(id: 2)
+    Test.assertEqual(statusScheduled, status!)
+    
+}
+
+
+access(all) fun testScheduledTransactionDestroyHandler() {
```
devbugging, line 176: nice

contracts/FlowTransactionScheduler.cdc
joshuahannan: Issue created here: https://github.com/onflow/flow-core-contracts/issues/538

joshuahannan:
Merged to master! Testnet here we come!

PR #491, opened 2025-07-10
Adds support for low priority callbacks

joshuahannan:
* Adds logic to handle low priority callbacks in the `process` function
* Adds fields to events to indicate priority and owner. The FVM can see which processed callbacks are low priority in any given block and wait to execute those until the high and medium priority ones have been executed.
* Updates the `slotQueue` field to be a dictionary of dictionaries to help with cleaning up callbacks
* Refactors the tests to use utility functions for many of the transactions and scripts


contracts/FlowCallbackScheduler.cdc
```diff
@@ -15,12 +15,28 @@ access(all) contract FlowCallbackScheduler {
...
+        fees: UFix64,
+        callbackOwner: Address
     )
 
-    access(all) event CallbackProcessed(id: UInt64, executionEffort: UInt64)
-    access(all) event CallbackExecuted(id: UInt64)
-    access(all) event CallbackCanceled(id: UInt64)
+    access(all) event CallbackProcessed(
+        id: UInt64,
+        priority: UInt8,
```
devbugging, line 24: this is not needed for FVM

contracts/FlowCallbackScheduler.cdc
```diff
@@ -465,31 +481,49 @@ access(all) contract FlowCallbackScheduler {
...
 
-            let highUsed = slotPriority[Priority.High] ?? 0
-            let mediumUsed = slotPriority[Priority.Medium] ?? 0
+            // Get how much effort has been used for each priority
+            let highUsed = slotPriorityEffortsUsed[Priority.High] ?? 0
+            let mediumUsed = slotPriorityEffortsUsed[Priority.Medium] ?? 0
+
+            // If it is low priority, return whatever effort is remaining
+            // under 5000
+            if priority == Priority.Low {
```
devbugging, line 507: Low priority should never be added as part of the effort available

contracts/FlowCallbackScheduler.cdc
```diff
@@ -554,20 +592,42 @@ access(all) contract FlowCallbackScheduler {
...
+            // and add low priority callbacks to the timestamp if there is space
             for timestamp in pastTimestamps {
-                let callbackIDs = self.slotQueue[timestamp] ?? []
+                let callbackIDs = self.slotQueue[timestamp] ?? {}
+
+                // Add low priority callbacks to the list
+                // until the low available effort is used up
+                // todo: This could get pretty costly if there are a lot of low priority callbacks
+                // in the queue. Figure out how to more efficiently go through the low priority callbacks
+                // Could potentially limit the size of the low priority callback queue?
```
devbugging, line 608: The low priority queue is limited with effort limit, so each timestamp will only have max X low priorities. Not sure what you mean with this?

contracts/FlowCallbackScheduler.cdc
joshuahannan: I want to keep it anyway in case there are third parties that want to follow events and need more info, like @bjartek was asking for

PR #494, opened 2025-07-14
Adds Canceling callbacks logic and emits events in priority order

joshuahannan:
* Adds logic to correctly cancel and remove reserved execution effort for callbacks
* Does not store status for executed callbacks
* Adds some tests for callback canceling
* Updates the process() function to emit the processed events in order of priority, with High first, medium second, and low third
    * The testing framework currently cannot get the order in which the events were emitted, so there isn't a good way to test this yet. We'll need to test in flow-go integration tests
* Updates initialization for low priority slot


contracts/FlowCallbackScheduler.cdc
```diff
@@ -646,10 +658,27 @@ access(all) contract FlowCallbackScheduler {
...
+            // Remove this callback id from its slot
+            let slotQueue = self.slotQueue[callback.scheduledTimestamp]!
+            slotQueue[id] = nil
+            self.slotQueue[callback.scheduledTimestamp] = slotQueue
+            
+            // Subtract the execution effort for this callback from the slot's priority
+            // Low priority effots don't count toward a slot's execution effort
+            // so we don't need to subtract anything for them
+            if callback.priority != Priority.Low {
+
```
devbugging, line 671:
```suggestion
```

contracts/FlowCallbackScheduler.cdc
```diff
@@ -682,16 +713,17 @@ access(all) contract FlowCallbackScheduler {
                         priority: callback.priority.rawValue,
                         callbackOwner: callback.handler.address
                     )
+
+                    // keep historic Canceled status for future queries after garbage collection
+                    // We don't keep executed statuses because we can just assume
+                    // they every ID that is less than the current ID counter
+                    // that is not Canceled, Scheduled, or Processed is executed
+                    let historic = HistoricStatus(timestamp: callback.scheduledTimestamp, status: status)
```
devbugging, line 721: nice

contracts/FlowCallbackScheduler.cdc
```diff
@@ -432,6 +422,8 @@ access(all) contract FlowCallbackScheduler {
             // if the callback is not found in the callbacks map, we check the callback status map for historic status
             if let historic = self.historicStatuses[id] {
                 return historic.status
+            } else if id < self.nextID {
```
devbugging, line 425: add a comment as to why this is so

contracts/FlowCallbackScheduler.cdc
```diff
@@ -613,21 +619,27 @@ access(all) contract FlowCallbackScheduler {
...
+                                callbackOwner: callback.handler.address
+                            )
+                        }
+                    } else {
+                        // This should ideally not happen if callbackIDs are correctly managed
+                        // but adding a panic for robustness in case of unexpected state
+                        panic("Invalid ID: \(id) callback not found during processing")
+                    }
                 }
             }
```
devbugging, line 644: I'm not sure sorting is justified; it wasn't part of the FLIP, and it adds complexity. It might seem like an okay feature, but I don't like the complexity it adds. 
Also we were discussing randomizations of callback execution. Until we agree on it I wouldn't make this change. You can leave this change on another branch for maybe later but don't see a good reason why to add it now, unless you can convince me otherwise.

devbugging (changes requested):
Left a comment on sorting

contracts/FlowCallbackScheduler.cdc
devbugging: I don't remember saying that. The ordering section in FLIP doesn't mention it https://github.com/devbugging/flips/blob/b0b49835eac427148cd4738290870969748da52a/protocol/20250609-scheduled-callbacks.md#order-of-execution 
For now I would stick to non-ordered way (beside ofc low priority being always last - but that's anyway how it needs to be). Later if we decide we want to randomise the order we can revisit. Why I'm also saying this is because I have a feeling there will be an ask to randomise as it was mentioned in the FLIP review.

devbugging: We did discuss it, you are right. We can keep this ordering, I don't mind then.

PR #497, opened 2025-07-29
Adds cleanup for failed callbacks and additional tests

joshuahannan:
* Adds a few lines to `process()` to automatically clean up callbacks that were processed but not executed. I believe doing it automatically here should be okay because `process()` will not be called until after all previously processed callbacks have been executed, but @devbugging please let me know if that is true or not. If that isn't true, then there is a risk of marking callbacks as executed before they have actually been executed by the FVM, and we definitely do not want that
* Rearranges/renames some of the functions and types in the contract for better readability and clarity
* Adds additional estimate tests and tests for marking failed callbacks as executed


contracts/FlowCallbackScheduler.cdc
```diff
@@ -71,7 +75,7 @@ access(all) contract FlowCallbackScheduler {
     /// Scheduled callback contains methods to cancel the callback and obtain the status. 
     /// It can only be created by the scheduler contract to prevent spoofing.
     access(all) struct ScheduledCallback {
-        access(self) let scheduler: Capability<auth(CancelCallback, ReadCallbackStatus) &SharedScheduler>
+        access(self) let scheduler: Capability<auth(CancelCallback) &SharedScheduler>
```
joshuahannan, line 78: There was no need for the `ReadCallbackStatus` entitlement since it was just a view function

contracts/FlowCallbackScheduler.cdc
```diff
@@ -307,13 +364,63 @@ access(all) contract FlowCallbackScheduler {
...
+        /// calculate fee by converting execution effort to a fee in Flow tokens.
+        access(all) view fun calculateFee(executionEffort: UInt64, priority: Priority): UFix64 {
+            // Use the official FlowFees calculation
+            let baseFee = FlowFees.computeFees(inclusionEffort: 1.0, executionEffort: UFix64(executionEffort))
+            
+            return baseFee * self.priorityFeeMultipliers[priority]!
+        }
+
+        /// getNextIDAndIncrement returns the next ID and increments the ID counter
+        access(self) fun getNextIDAndIncrement(): UInt64 {
```
joshuahannan, line 381: I just moved these to the top since they are all just getter functions

contracts/FlowCallbackScheduler.cdc
```diff
@@ -307,13 +364,63 @@ access(all) contract FlowCallbackScheduler {
...
             let nextID = self.nextID
             self.nextID = self.nextID + 1
             return nextID
         }
 
+        /// Gets a struct containing all the configurable metadata
+        /// of the Scheduler resource
+        access(all) fun getConfigMetadata(): SchedulerConfig {
```
joshuahannan, line 389: Allows anyone to query the metadata that governs the callback scheduler contract

contracts/FlowCallbackScheduler.cdc
```diff
@@ -614,10 +691,18 @@ access(all) contract FlowCallbackScheduler {
...
-                    if self.getPriority(id: id) == Priority.High {
-                        highPriorityIDs.append(id)
+                    let callback = self.borrowCallback(id: id)!
+
+                    // If any callbacks failed to execute properly because of an error,
+                    // mark them as executed
+                    if callback.status == Status.Processed {
+                        self.finalizeCallback(callback: callback, status: Status.Executed)
+                        continue
                     }
```
joshuahannan, line 701: This addition is what I am curious about. Is it okay to clean the failed callbacks here in `process()`, or does it need to be an explicit fvm transaction for each one?

joshuahannan: Sure, we need to discuss it more, but we can't just not handle failed transactions at the moment or the contract will get in a weird state with all the transactions that never leave the `Processed` state. I think we should either keep this or the manual way to mark failed callbacks as executed and create an issue to revisit it in the future to decide on a final solution.

PR #498, opened 2025-07-30
Adds calculation for storage fees

joshuahannan:
* Adds an additional fee to `calculateFee` for the cost of storing `data`. This will prevent users from spamming the contract with callback data that is prohibitively large to store. The storage fee is not subject to the priority multiplier though because it is just temporarily paying for the storage.


contracts/FlowCallbackScheduler.cdc
```diff
@@ -370,11 +371,17 @@ access(all) contract FlowCallbackScheduler {
         }
 
         /// calculate fee by converting execution effort to a fee in Flow tokens.
-        access(all) view fun calculateFee(executionEffort: UInt64, priority: Priority): UFix64 {
```
devbugging, line 373: how come this is not `view` anymore?

contracts/FlowCallbackScheduler.cdc
```diff
@@ -911,6 +918,22 @@ access(all) contract FlowCallbackScheduler {
...
+    access(all) fun getSizeofData(_ data: AnyStruct?): UFix64 {
+        if data == nil {
+            return 0.0
+        }
+        let storagePath = /storage/dataTemp
+        let storageUsedBefore = self.account.storage.used
+        self.account.storage.save(data!, to: storagePath)
+        let storageUsedAfter = self.account.storage.used
+        self.account.storage.load<AnyStruct>(from: storagePath)
+        return FlowStorageFees.convertUInt64StorageBytesToUFix64Megabytes(storageUsedAfter - storageUsedBefore)
```
bluesign, line 935: this can underflow, some atree internal optimizations.

PR #500, opened 2025-08-01
Events updates and PR comments

joshuahannan:
* Renames events and entitlements to just verbs to be more consistent with Cadence best practices
* Rearranges and adds parameters to events for more relevant metadata
* Moves event emission to more logical functions for cancel and execute so they can access the data needed for their new parameters
* Adds tests to verify event arguments
* Addresses most of Gio's comments in https://github.com/onflow/flow-core-contracts/pull/489


contracts/FlowCallbackScheduler.cdc
```diff
@@ -26,58 +26,70 @@ access(all) contract FlowCallbackScheduler {
...
 
-    access(all) event CallbackExecuted(
+    access(all) event Executed(
         id: UInt64,
         priority: UInt8,
+        executionEffort: UInt64,
+        fees: UFix64,
         callbackOwner: Address,
-        fees: UFix64
+        succeeded: Bool
```
devbugging, line 51: Cool idea, but I think if we implement failures for callbacks, we will likely also add a status for callback which will be succeeded / failed and executed status will be removed. So I believe we will have two events succeeded and failed. So not sure about adding this field.

contracts/FlowCallbackScheduler.cdc
```diff
@@ -26,58 +26,70 @@ access(all) contract FlowCallbackScheduler {
...
+    // Event listeners can listen to this and query the new configuration
+    // if they need to
+    access(all) event ConfigUpdated()
+
     /// Entitlements
-    access(all) entitlement ExecuteCallback
-    access(all) entitlement CancelCallback
+    access(all) entitlement Execute
+    access(all) entitlement Cancel
     access(all) entitlement UpdateMetadata
```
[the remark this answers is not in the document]
joshuahannan: I think I am going to stick with this one. Since the others refer to callbacks, it would be nice to keep it clear that this one is for the metadata

contracts/FlowCallbackScheduler.cdc
joshuahannan: What do you mean by "if we implement failures", don't we have to implement failures? I assumed that is was mandatory that we defined a standard way to handle failures.

PR #504, opened 2025-08-07
Use interface type for config storage

joshuahannan:
Updates the config storage in the scheduler to be an interface type instead of a concrete type so that we have more flexibility to add additional config fields in the future


devbugging (approved):
Great

PR #512, opened 2025-08-12
Removes Failed Callbacks so no shared state is written to during `execute`

joshuahannan:
Copied from https://github.com/onflow/flow-core-contracts/pull/511

Marking callbacks as failed is unfortunately not possible without sacrificing concurrent execution of callbacks, which is not acceptable.

The problem is that we cannot access shared state in the executeCallback(id) function; thus, we cannot mark the callback as succeeded after it was executed. Which in turn means there's no way to distinguish between failed or succeeded.

The success of callback execution would either have to be checked off-chain by monitoring executed events (which are only emitted after callback execution) or on-chain by implementing additional logic in the contract receiving callbacks. Since the callback function also passes the callback ID, the contract can keep their own list of callbacks that were executed successfully and those that werent.


contracts/FlowCallbackScheduler.cdc
```diff
@@ -850,6 +850,12 @@ access(all) contract FlowCallbackScheduler {
                 for id in callbackIDs.keys {
                     let callback = self.borrowCallback(id: id)!
 
+                    // Callbacks that are already executed are finalized and removed from the queue
```
devbugging, line 855: is there a reason we make it so the callback is removed from the queue only after next callback is executed? Meaning what was the thinking to seperate this from finalize being called when executed event is emitted?

contracts/FlowCallbackScheduler.cdc
```diff
@@ -962,18 +982,8 @@ access(all) contract FlowCallbackScheduler {
...
-        /// in the same block after it is processed so it won't get processed twice
-        access(contract) fun finalizeCallback(callback: &CallbackData, status: Status) {
-            pre {
-                status == Status.Executed || status == Status.Canceled: 
-                    "Invalid status: The provided status to finalizeCallback must be Executed or Canceled"
-            }
-
-            callback.setStatus(newStatus: status)
+        /// finalizes and garbage collects the callback by removing the callback resource and the slot map if it is empty.
+        access(contract) fun finalizeCallback(callback: &CallbackData) {
```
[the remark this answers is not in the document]
joshuahannan: good suggestion. I'll rename to `garbageCollect`.

I realized that the canceled state actually is meaningless here because there is no point where a callback is canceled but isn't removed from the queue, so marking it as canceled before garbage collecting was just using unnecessary computation, so it didn't really make sense to keep that check. I'll add a check that the status is not scheduled though.

contracts/FlowCallbackScheduler.cdc
devbugging: lol right!

PR #513, opened 2025-08-15
Add low prio slot and correct processing

joshuahannan:
The low priority callbacks were all being processed immediately instead of waiting until the callback they were scheduled for.  This change gives low priority callbacks their own slot and processes them differently so they only get processed at or after the slot they were scheduled for


contracts/FlowCallbackScheduler.cdc
```diff
@@ -528,9 +532,12 @@ access(all) contract FlowCallbackScheduler {
             return nextID
         }
 
-        /// get status of the scheduled callback, if the callback is not found Unknown is returned.
+        /// get status of the scheduled callback
+        /// @param id: The ID of the callback to get the status of
+        /// @return Status: The status of the callback, if the callback is not found Unknown is returned.
         access(all) view fun getStatus(id: UInt64): Status? {
```
devbugging, line 538: We have a bug here... the problem is that if we query for ID that is bigger than actually executed IDs it will return executed status. We must add another field that keeps track of latest executed ID, then we need to add a check `if id > latestExecutedID return Unknown`

contracts/FlowCallbackScheduler.cdc
```diff
@@ -541,11 +548,6 @@ access(all) contract FlowCallbackScheduler {
                 return Status.Canceled
             }
 
-            // if no callbacks were yet canceled the status is executed
-            if self.canceledCallbacks.length == 0 {
-                return Status.Executed
-            }
```
devbugging, line 547: We have a bug here... the problem is that if we query for ID that is bigger than actually executed IDs it will return executed status. We must add another field that keeps track of latest executed ID, then we need to add a check `if id > latestExecutedID return Unknown`

devbugging: Having latest executed ID we also don't have to check for first canceled ID, we can just check if ID is in canceled ID list, and if it's lower than last executed ID. We should remove init the canceled with 0

joshuahannan: I don't think that is necessary. We can just check if `id > self.nextID` and return `nil` if it is

devbugging (approved):

PR #518, opened 2025-08-29
Add additional callback test cases and reorg tests

joshuahannan:
* Adds additional tests to get full code coverage
* Makes fields and functions as restrictive as possible
* Adds getter for canceled callbacks and additional tests for filling up the canceled callbacks array
* Adds tests for a callback that schedules another callback
* Adds tests toe verify that fees were deposited properly
* Removes redundant code
* Reorganizes tests into three different files to avoid timestamp errors

### DIFF
--- a/tests/callbackScheduler_test.cdc
+++ b/tests/callbackScheduler_test.cdc
@@ -120,7 +120,7 @@ access(all) struct EstimateTestCase {
 
 access(all) fun testEstimate() {
 
-    let currentTime = getTimestamp()
+    let currentTime = getCurrentBlock().timestamp
     let futureTime = currentTime + 100.0
     let pastTime = currentTime - 100.0
     let farFutureTime = currentTime + 10000.0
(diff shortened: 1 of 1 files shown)


devbugging (approved):

tests/callbackScheduler_test.cdc
```diff
@@ -0,0 +1,767 @@
...
+        self.data = data
+        self.expectedFee = expectedFee
+        self.expectedTimestamp = expectedTimestamp
+        self.expectedError = expectedError
+    }
+}
+
+access(all) fun testEstimate() {
+
+    let currentTime = getTimestamp()
```
m-Peter, line 123: To avoid flakiness, it is better to use:
```cadence
let currentTime = getCurrentBlock().timestamp
```

Although I haven't had time to write proper documentation for this :sweat_smile: , the Cadence Testing Framework generally provides 2 types of testing strategies: **unit tests** & **integration tests**.
- An **integration test** will deploy some contract/s to Emulator, and interact with it using Cadence scripts & transactions.
- A **unit test** will simply import a contract in a test file, and call its public methods.

This test file (`tests/callbackScheduler_test.cdc`) has the form of a unit test.
```cadence
import "FlowCallbackScheduler"
...

let estimate = FlowCallbackScheduler.estimate(
    data: testCase.data,
    timestamp: testCase.timestamp,
    priority: testCase.priority,
    executionEffort: testCase.executionEffort
)
```

it imports the `FlowCallbackScheduler` contract, and later on calls its `estimate` function. Since there is no contract deployment involved, it is better to use the unit test APIs for fetching the timestamp etc, and avoid using scripts.

`getTimestamp()` would fetch the timestamp using a Cadence script, reading the Emulator's time.
`getCurrentBlock().timestamp` is the unit test API for fetching the testing environment's time. 

I tried the 2nd approach, and I didn't get any failing test, in more than 100+ attempts :pray: 
If this is too much to grasp, we can also arrange a call.

joshuahannan: that makes sense. I forgot that we weren't using scripts for this and just calling directly. thanks for checking

PR #521, opened 2025-09-05
Make ScheduledCallback a Resource

joshuahannan:
ScheduledCallback is a representation of monetary value since it can access the fees that were paid for the callback and it has privileged functionality, so it should be a resource instead of a struct.


contracts/FlowCallbackScheduler.cdc
```diff
@@ -145,11 +145,11 @@ access(all) contract FlowCallbackScheduler {
...
     /// Structs
 
-    /// ScheduledCallback is the struct that the user receives after scheduling a callback.
+    /// ScheduledCallback is the resource that the user receives after scheduling a callback.
     /// It allows them to get the status of their callback and can be passed back
     /// to the scheduler contract to cancel the callback if it has not yet been executed. 
     /// It can only be created by the scheduler contract to prevent spoofing.
-    access(all) struct ScheduledCallback {
+    access(all) resource ScheduledCallback {
         access(self) let scheduler: Capability<auth(Cancel) &SharedScheduler>
```
sisyphusSmiling, line 153: Since this is a resource, we can probably remove this field now

sisyphusSmiling (commented):
Pretty straightforward, just left one minor suggestion.

In the process, I found myself wondering how one could get information about a pending callback from the ID. For instance, could I find out the time, address & type of handler for a pending callback? I didn't see a route that would give me that info but it could be helpful, though it's not related to this PR.

joshuahannan:
@sisyphusSmiling Callbacks are marked as pending in the same block that they are executed, so adding functionality to get info about a pending callback wouldn't actually be useful

sisyphusSmiling:
@joshuahannan I see, thanks for highlighting that. It might be helpful to add methods on the struct like `getHandlerType(): Type?` and `getHandlerAddress(): Address` since the handler Cap isn't available outside of the contract (understandably so)

PR #523, opened 2025-09-08
Adds FlowCallbackUtils contract

joshuahannan:
* Adds a contract that contains utils for callbacks on flow
* Starts out with just CallbackManager, a resource to schedule and cancel callbacks from
* Potentially can have other utils added in the future

It is important to remember that since handlers still need to be created for the actual logic of the callback, this does not allow us to completely standardize the callback transactions because you still need to have code in your transactions to deal with your specific callback handler

I'm still looking for feedback on these changes, so please take a look and let me know what you think. I especially need feedback about usability. I think this improves usability, but I don't always have the best instincts about that.


devbugging:
I can see the utility of this, but don't like the need to change the scheduler contract. This could easily be just a util contract composed on top without changing the contract. 
This makes us more flexible in implementing this (instead of last minute change, that wasn't discussed much), and also waiting to see more real world use to implement the contract. 
I'm oposed to changing the scheduler contract to fit into util contract.

joshuahannan:
Looks like the consensus is that the schedule function should be `access(all)`, so I'll change it back to that. Does anyone have any feedback on the manager part specifically, now that it is just optional to use it?

PR #530, opened 2025-09-11
Adds FlowTransactionSchedulerUtils contract

joshuahannan:
Adds a contract that will contain utilities for the transaction scheduler contract
Currently it just has a scheduled transaction manager resource for now, but will have scheduled transaction metadata views and potentially other constructs in the future

This is nice from a developer education standpoint because it removes the unsafe patters we were using in the previous version with the public functions in the Handler contract and replaces them with the safer versions that use the manager resource


joshuahannan:
I was having some issue with git, so I am closing this in favor of https://github.com/onflow/flow-core-contracts/pull/542

PR #542, opened 2025-09-16
Adds FlowTransactionSchedulerUtils contract

joshuahannan:
Adds a contract that will contain utilities for the transaction scheduler contract
Currently it just has a scheduled transaction manager resource and a single metadata view, but we can add additional utility methods for reducing boilerplate in transactions and add additional resources for potentially managing handlers and other things.

This is nice from a developer education standpoint because it removes the unsafe patters we were using in the previous version with the public functions in the Handler contract and replaces them with the safer versions that use the manager resource.

### DIFF
--- a/contracts/FlowTransactionScheduler.cdc
+++ b/contracts/FlowTransactionScheduler.cdc
@@ -56,7 +56,8 @@ access(all) contract FlowTransactionScheduler {
         executionEffort: UInt64,
         fees: UFix64,
         transactionHandlerOwner: Address,
-        transactionHandlerTypeIdentifier: String
+        transactionHandlerTypeIdentifier: String,
+        transactionHandlerPublicPath: PublicPath?
     )
 
     /// Emitted when a scheduled transaction's scheduled timestamp is reached and it is ready for execution
@@ -75,7 +76,8 @@ access(all) contract FlowTransactionScheduler {
         priority: UInt8,
         executionEffort: UInt64,
         transactionHandlerOwner: Address,
-        transactionHandlerTypeIdentifier: String
+        transactionHandlerTypeIdentifier: String,
+        transactionHandlerPublicPath: PublicPath?
     )
 
     /// Emitted when a scheduled transaction is canceled by the creator of the transaction
@@ -762,6 +764,11 @@ access(all) contract FlowTransactionScheduler {
             // Deposit the fees to the service account's vault
             FlowTransactionScheduler.depositFees(from: <-fees)
 
+            let handlerRef = handlerCap.borrow()
+                ?? panic("Invalid transaction handler: Could not borrow a reference to the transaction handler")
+
+            let handlerPublicPath = handlerRef.resolveView(Type<PublicPath>()) as? PublicPath
+
             emit Scheduled(
                 id: transactionData.id,
                 priority: transactionData.priority.rawValue,
@@ -769,7 +776,8 @@ access(all) contract FlowTransactionScheduler {
                 executionEffort: transactionData.executionEffort,
                 fees: transactionData.fees,
                 transactionHandlerOwner: transactionData.handler.address,
-                transactionHandlerTypeIdentifier: transactionData.handlerTypeIdentifier
+                transactionHandlerTypeIdentifier: transactionData.handlerTypeIdentifier,
+                transactionHandlerPublicPath: handlerPublicPath
             )
 
             // Add the transaction to the slot queue and update the internal state
@@ -1325,12 +1333,16 @@ access(all) contract FlowTransactionScheduler {
             let transactionHandler = tx.handler.borrow()
                 ?? panic("Invalid transaction handler: Could not borrow a reference to the transaction handler")
 
+            let handlerPublicPath = transactionHandler.resolveView(Type<PublicPath>()) as? PublicPath
+
             emit Executed(
                 id: tx.id,
                 priority: tx.priority.rawValue,
                 executionEffort: tx.executionEffort,
                 transactionHandlerOwner: tx.handler.address,
-                transactionHandlerTypeIdentifier: transactionHandler.getType().identifier
+                transactionHandlerTypeIdentifier: transactionHandler.getType().identifier,
+                transactionHandlerPublicPath: handlerPublicPath
+
             )
             
             transactionHandler.executeTransaction(id: id, data: tx.getData())
--- a/contracts/FlowTransactionScheduler.cdc
+++ b/contracts/FlowTransactionScheduler.cdc
@@ -276,6 +276,14 @@ access(all) contract FlowTransactionScheduler {
         access(contract) view fun getData(): AnyStruct? {
             return self.data
         }
+
+        /// getUnentitledHandlerReference returns an un-entitled reference to the transaction handler
+        /// This allows users to query metadata views about the handler
+        /// @return: An un-entitled reference to the transaction handler
+        access(all) view fun getUnentitledHandlerReference(): &{TransactionHandler} {
+            return self.handler.borrow() as? &{TransactionHandler}
+                ?? panic("Invalid transaction handler: Could not borrow a reference to the transaction handler")
+        }
     }
 
     /// Struct interface representing all the base configuration details in the Scheduler contract
--- a/contracts/FlowTransactionSchedulerUtils.cdc
+++ b/contracts/FlowTransactionSchedulerUtils.cdc
@@ -43,6 +43,28 @@ access(all) contract FlowTransactionSchedulerUtils {
             self.handlers = {}
         }
 
+        /// scheduleByHandler schedules a transaction by a given handler that has been used before
+        /// @param handlerTypeIdentifier: The type identifier of the handler
+        /// @param data: Optional data to pass to the transaction when executed
+        /// @param timestamp: The timestamp when the transaction should be executed
+        /// @param priority: The priority of the transaction (High, Medium, or Low)
+        /// @param executionEffort: The execution effort for the transaction
+        /// @param fees: A FlowToken vault containing sufficient fees
+        /// @return: The ID of the scheduled transaction
+        access(Owner) fun scheduleByHandler(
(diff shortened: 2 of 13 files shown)


contracts/FlowTransactionSchedulerUtils.cdc
```diff
@@ -0,0 +1,335 @@
...
+        /// @param id: The ID of the transaction to retrieve
+        /// @return: A reference to the transaction handler, or nil if not found
+        access(all) view fun getTransactionHandler(id: UInt64): &{FlowTransactionScheduler.TransactionHandler}? {
+            let txData = self.getTransactionData(id: id)
+            return txData?.getUnentitledHandlerReference()
+        }
+
+        /// Get all the handler type identifiers that the manager has transactions scheduled for
+        /// @return: An array of all handler type identifiers
+        access(all) view fun getHandlerTypeIdentifiers(): [String] {
```
mfbz, line 216: Would it be possible to have a method to retrieve handler unentitled reference by type identifier?
This would allow us to show all the handlers data from an address:

1 - we get handler type identifiers by address (getHandlerTypeIdentifiers())
2 - then get handler reference from which we can access views for each handler (getHandlerByTypeIdentifier(typeIdentifier))

joshuahannan: Yeah, we can add that for the manager, but not globally unless you know the address of which manager you're querying

contracts/FlowTransactionSchedulerUtils.cdc
```diff
@@ -0,0 +1,335 @@
...
+        access(Owner) fun cleanup(): [UInt64] {
+            let currentTimestamp = getCurrentBlock().timestamp
+            var transactionsToRemove: [UInt64] = []
+
+            let pastTimestamps = self.sortedTimestamps.getBefore(current: currentTimestamp)
+            for timestamp in pastTimestamps {
+                let ids = self.idsByTimestamp[timestamp] ?? []
+                for id in ids {
+                    let status = FlowTransactionScheduler.getStatus(id: id)
+                    if status == nil || status == FlowTransactionScheduler.Status.Unknown {
```
chasefleming, line 179: Do you need to add a cancelled or executed check here?

contracts/FlowTransactionSchedulerUtils.cdc
```diff
@@ -0,0 +1,335 @@
...
+
+        /// Schedule a transaction and store it in the manager's dictionary
+        /// @param handlerCap: A capability to a resource that implements the TransactionHandler interface
+        /// @param data: Optional data to pass to the transaction when executed
+        /// @param timestamp: The timestamp when the transaction should be executed
+        /// @param priority: The priority of the transaction (High, Medium, or Low)
+        /// @param executionEffort: The execution effort for the transaction
+        /// @param fees: A FlowToken vault containing sufficient fees
+        /// @return: The ID of the scheduled transaction
+        access(Owner) fun schedule(
```
chasefleming, line 76: Should we add any guards here to check for bad inputs?

contracts/FlowTransactionSchedulerUtils.cdc
```diff
@@ -0,0 +1,335 @@
...
+            if self.scheduledTransactions.containsKey(id) {
+                return FlowTransactionScheduler.getTransactionData(id: id)
+            }
+            return nil
+        }
+
+        /// Get an un-entitled reference to a transaction handler of a given ID
+        /// @param id: The ID of the transaction to retrieve
+        /// @return: A reference to the transaction handler, or nil if not found
+        access(all) view fun getTransactionHandler(id: UInt64): &{FlowTransactionScheduler.TransactionHandler}? {
```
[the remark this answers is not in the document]
mfbz: I was meaning that this method doesn't necessarily need the manager to return the data so it could be better suited maybe for the other core contract.

```cadence
let txData = FlowTransactionScheduler.getTransactionData(id: transactionId) {
let handler = txData.getUnentitledHandlerReference()
```

In this case i got an unentitled handler reference from which i can get the views of an handler associated to an id but i don't need the manager to get this data. Am i missing something?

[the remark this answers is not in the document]
joshuahannan: oh yeah, that is definitely possible to do in the main contract. I can add a utility method for that

contracts/FlowTransactionSchedulerUtils.cdc
joshuahannan: I was thinking that we'd like to keep the executed transaction statuses in the manager for some amount of time after they are executed in case the owner wants to query the status or information about it or something like that, so right now, we're only removing transactions if they no longer have a status. 

It might be better idea to set a limit on the number here because depending on the behavior of the main contract, the storage could start to get a little out of hand. But I think it probably is better to just monitor it for now and we can always add those limits later if needed

contracts/FlowTransactionScheduler.cdc
```diff
@@ -75,7 +76,8 @@ access(all) contract FlowTransactionScheduler {
         priority: UInt8,
         executionEffort: UInt64,
         transactionHandlerOwner: Address,
-        transactionHandlerTypeIdentifier: String
+        transactionHandlerTypeIdentifier: String,
```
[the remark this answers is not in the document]
joshuahannan: yes, good idea. I'll add the uuid to the events

contracts/FlowTransactionScheduler.cdc
```diff
@@ -75,7 +76,8 @@ access(all) contract FlowTransactionScheduler {
...
+        transactionHandlerPublicPath: PublicPath?
```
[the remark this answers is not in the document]
joshuahannan: I anticipate tools will listen for events like this to get information about what handlers are used by what callbacks and with the public path and the other info that is already there, they have all the information they need to find the handler object and query metadata views about that handler. I know it can be faked, but I don't see that being a problem. I think the benefit is worth more than the risk

contracts/FlowTransactionSchedulerUtils.cdc
```diff
@@ -0,0 +1,519 @@
...
+
+            // Store the handler capability in the handlers dictionary for later retrieval
+            if let handlers = self.handlerInfos[handlerTypeIdentifier] {
+                if let handlerInfo = handlers[handlerUUID] {
+                    handlerInfo.addTransactionID(id: id)
+                    handlers[handlerUUID] = handlerInfo
+                } else {
+                    let handlerInfo = HandlerInfo(typeIdentifier: handlerTypeIdentifier, capability: handlerCap)
+                    handlerInfo.addTransactionID(id: id)
+                    handlers[handlerUUID] = HandlerInfo(typeIdentifier: handlerTypeIdentifier, capability: handlerCap)
```
mfbz, line 193: Shouldn't save the updated handlerInfo at uuid?

```suggestion
                    handlers[handlerUUID] = handlerInfo
```

joshuahannan: yes, good catch. Hadn't gotten to that test case yet

contracts/FlowTransactionScheduler.cdc
```diff
@@ -75,7 +80,12 @@ access(all) contract FlowTransactionScheduler {
         priority: UInt8,
         executionEffort: UInt64,
         transactionHandlerOwner: Address,
-        transactionHandlerTypeIdentifier: String
+        transactionHandlerTypeIdentifier: String,
+        transactionHandlerUUID: UInt64,
+
+        // The public path of the transaction handler that can be used to resolve views
```
devbugging, line 86: couldn't this be retrieved from the Utils contract by the event ID? I feel we should store metadata in the Utils contract and then make the tools retrieve it from there using the ID. This makes the event not contain any non-critical info, and also allows for extensibility down the line, because we now for sure can't predict all the usecases

contracts/FlowTransactionSchedulerUtils.cdc
```diff
@@ -0,0 +1,519 @@
...
+                }
+            }
+            
+            return transactionsInTimeframe
+        }
+
+        /// Get the status of a transaction by its ID
+        /// @param id: The ID of the transaction
+        /// @return: The status of the transaction, or Status.Unknown if not found in manager
+        access(all) view fun getTransactionStatus(id: UInt64): FlowTransactionScheduler.Status? {
```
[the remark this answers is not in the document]
joshuahannan: I'm a little confused by what you're saying. Could you elaborate or maybe provide a code example?

contracts/FlowTransactionSchedulerUtils.cdc
```diff
@@ -0,0 +1,519 @@
...
+        access(all) fun getHandlerViews(handlerTypeIdentifier: String, handlerUUID: UInt64?): [Type] 
+        access(all) fun resolveHandlerView(handlerTypeIdentifier: String, handlerUUID: UInt64?, viewType: Type): AnyStruct?      
+        access(all) fun getHandlerViewsFromTransactionID(_ id: UInt64): [Type]
+        access(all) fun resolveHandlerViewFromTransactionID(_ id: UInt64, viewType: Type): AnyStruct? 
+        access(all) view fun getTransactionIDs(): [UInt64]
+        access(all) view fun getTransactionIDsByHandler(handlerTypeIdentifier: String, handlerUUID: UInt64?): [UInt64]
+        access(all) view fun getTransactionIDsByTimestamp(_ timestamp: UFix64): [UInt64]
+        access(all) fun getTransactionIDsByTimestampRange(startTimestamp: UFix64, endTimestamp: UFix64): {UFix64: [UInt64]}
+        access(all) view fun getTransactionStatus(id: UInt64): FlowTransactionScheduler.Status?
+    }
```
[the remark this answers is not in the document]
mfbz: yep it make sense 👍

contracts/FlowTransactionSchedulerUtils.cdc
joshuahannan: added

PR #545, opened 2025-10-01
Add missing FungibleTokenMetadataViews dependency

turbolent:
When checking out the repository and installing all dependencies, the project still does not seem to check. Tracking the problem down, it looks like the dependency `FungibleTokenMetadataViews` is missing.

Add it to the `flow.json` config.


turbolent:
@joshuahannan Nice work fixing the tests 👍

PR #546, opened 2025-10-07
Scheduled Txs: Updates limits and add removal limit

joshuahannan:
This PR focuses on optimizing the Scheduled Transactions feature.

The key changes involve:

Limit Updates: Lowers cumulative effort limits and all priority effort limits by half
Removal limit: Adds a limit to the number of transactions that can be removed during `removeExecutedTransactions()` so it doesn't hit the gas limit


contracts/FlowTransactionScheduler.cdc
```diff
@@ -1220,8 +1221,38 @@ access(all) contract FlowTransactionScheduler {
...
+
+                    // if the priority queue is now empty remove it from the map
+                    if transactionIDs.keys.length == 0 {
+                        transactionPriorities.remove(key: priority)
+                    } else {
+                        transactionPriorities[priority] = transactionIDs
+                    }
+
+                    if numRemoved >= self.config.collectionTransactionsLimit {
+                        break
```
devbugging, line 1242: add an event to be emitted here `ItterationLimitReached`

contracts/FlowTransactionScheduler.cdc
```diff
@@ -1220,8 +1221,38 @@ access(all) contract FlowTransactionScheduler {
                         // charge the full fee for transaction execution
                         destroy tx.payAndRefundFees(refundMultiplier: 0.0)
 
-                        self.removeTransaction(txData: tx)
+                        // remove transaction object
+                        let transactionObject = self.transactions.remove(key: id)!
+
+                        transactionIDs.remove(key: id)
+
+                        if numRemoved >= self.config.collectionTransactionsLimit {
```
devbugging, line 1229: I would just check this limit as first line in the inner for loop (before borrowTransaction) and if reached emit an even and actually return, so we leave to next itteration to clear remaining, this way its cleaner to me

PR #553, opened 2025-10-27
Adds COA handler and schedule txs

joshuahannan:
Adds a handler type that can schedule transactions for a COA.

Currently can schedule deposit FLOW, withdraw FLOW, and `call()`, but could possibly do more. This would make the handler more complicated though and may not be worth the effort


devbugging:
> EVM is accessible in the emulator, right? We don't have to do anything else to make it work properly? I just want to make sure that is true before trying to write too many tests

The emulator should have EVM enabled in the same way as other networks. Is anything not working?

contracts/FlowTransactionSchedulerUtils.cdc
```diff
@@ -561,6 +564,217 @@ access(all) contract FlowTransactionSchedulerUtils {
...
+        /// Not required for the Call transaction type
+        access(all) let amount: UFix64?
+
+        /// The following fields are only required for the Call transaction type
+        access(all) let callToEVMAddress: EVM.EVMAddress?
+        access(all) let data: [UInt8]?
+        access(all) let gasLimit: UInt64?
+        access(all) let value: EVM.Balance?   
+
+        init(txType: UInt8, amount: UFix64?, callToEVMAddress: [UInt8; 20]?, data: [UInt8]?, gasLimit: UInt64?, value: UFix64?) {
```
sisyphusSmiling: IMO `callToEVMAddress` should either be `String` or `EVM.EVMAddress` - fixed byte arrays can be a pain to manipulate between txn args & this `init`. In `init` scope, you can deserialize from `String` with `EVM.addressFromString()` though FYI the method will revert on failure.

I also think `value` would benefit from `UInt` precision - UFix64 only supports 8 while EVM-native FLOW is 18 decimals. The can be some cases where that precision is needed for a successful call. I vote we either we make this `UInt?` or we accept `EVM.Balance?`

contracts/FlowTransactionSchedulerUtils.cdc
```diff
@@ -561,6 +564,217 @@ access(all) contract FlowTransactionSchedulerUtils {
...
+        /// @param data: Optional data passed to the transaction execution. In this case, the data must be a COAHandlerParams struct with valid values.
+        access(FlowTransactionScheduler.Execute) fun executeTransaction(id: UInt64, data: AnyStruct?) {
+            let coa = self.coaCapability.borrow()
+            if coa == nil {
+                emit COAHandlerExecutionError(id: id, owner: self.owner?.address ?? Address(0x0), coaAddress: nil,
+                                              errorMessage: "COA capability is invalid or expired for scheduled transaction with ID \(id)")
+                return
+            }
+
+            if let params = data as? COAHandlerParams {
```
sisyphusSmiling: I think this should either be `[COAHandlerParams]` or the params struct should be updated to include a sequenced array of calls. In either case, sequenced calls would be very helpful for even basic EVM scenarios - e.g. DEX swap == `approve` & `swap`.

I also think there should be a mechanism by which a scheduler can define whether failure between calls forces a reversion.

Take two scenarios:
1. I want to schedule listing 5 NFTs for sale
2. I want to perform a DCA swap

In scenario 1., I likely don't care if one of the listings fail - I just want al listings executed and I may deal with the failed listings later.

In scenario 2., I probably want to abort the whole scheduled operation if my initial `approve` call fails.

As I'm writing, I'm also imagining another scenario where I want to schedule an arbitrage. In this scenario, I likely need the output of one call to be input of another call. You could argue that a scheduler should just leverage a solidity contract to perform all of this which is valid. But I think it's worth at least considering the scenario where it would be helpful to pipe data between calls and making a design decision about whether it's worth supporting or not.

contracts/FlowTransactionSchedulerUtils.cdc
```diff
@@ -561,6 +564,217 @@ access(all) contract FlowTransactionSchedulerUtils {
...
+        }
+    }
+
+    /// Enum for COA handler execution type
+    access(all) enum COAHandlerTxType: UInt8 {
+        access(all) case DepositFLOW
+        access(all) case WithdrawFLOW
+        access(all) case Call
+
+        // TODO: Should we have other transaction types??
```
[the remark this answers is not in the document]
joshuahannan: yeah, I didn't know what to do with those resources so I left it out for now. I guess we could have like a dictionary of arbitrary collection or vault capabilities that can be filled up, but I also just had a hard time believing that people would need to schedule transactions to bridge arbitrary NFTs. Maybe for tokens though for some defi purposes

contracts/FlowTransactionSchedulerUtils.cdc
joshuahannan: Thanks. String is definitely a better choice there so I'll update it.

I usually try to make passing the values as arguments use simple types so people don't have to construct complex types in their app code, so I think `UInt` is a good choice for `value`. I can do that

joshuahannan: If I change it to `UInt`, do I just provide that to the initializer? like this?
```
self.value = EVM.Balance(attoflow: unwrappedValue)
```
I'm reaching the limits of my evm knowledge. I gotta brush up on this

sisyphusSmiling: Yeah that would do it

contracts/FlowTransactionSchedulerUtils.cdc
```diff
@@ -561,6 +564,252 @@ access(all) contract FlowTransactionSchedulerUtils {
...
+                            }
+                            let vault = self.flowTokenVaultCapability.borrow()
+                            if vault == nil {
+                                self.emitError(id: id, errorMessage: "FlowToken vault capability is invalid or expired for scheduled transaction with ID \(id) and index \(index)")
+                                return
+                            }
+                            if txParams.amount! > vault!.balance && !txParams.revertOnFailure {
+                                self.emitError(id: id, errorMessage: "Insufficient FLOW in FlowToken vault for deposit into COA for scheduled transaction with ID \(id) and index \(index)")
+                                continue
+                            }
```
sisyphusSmiling: I feel like these early return cases are inconsistent. I would think the particular command would revert if `revertOnFailure`, otherwise continue. But it looks like we're returning early. Same applies to the other txn types. Maybe this is a matter of interpretation on `revertOnFailure`.

My impression is that `false` would attempt to catch any condition that would force a failure (e.g. `amount == nil` or `amount > vault.balance`) and continue on to the next txnType. On the other hand `true` would allow the entire scheduled execution to revert entirely in a failure case. Am I understanding the intended implementation correctly?

sisyphusSmiling (commented):
Couple remaining questions in my mind about addressing the `revertOnFailure` field

contracts/FlowTransactionSchedulerUtils.cdc
```diff
@@ -561,6 +564,252 @@ access(all) contract FlowTransactionSchedulerUtils {
...
+        /// @param data: Optional data passed to the transaction execution. In this case, the data must be a COAHandlerParams struct with valid values.
+        access(FlowTransactionScheduler.Execute) fun executeTransaction(id: UInt64, data: AnyStruct?) {
+            let coa = self.coaCapability.borrow()
+            if coa == nil {
+                emit COAHandlerExecutionError(id: id, owner: self.owner?.address ?? Address(0x0), coaAddress: nil,
+                                              errorMessage: "COA capability is invalid or expired for scheduled transaction with ID \(id)")
+                return
+            }
+
+            if let transactions = data as? [COAHandlerParams] {
```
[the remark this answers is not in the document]
joshuahannan: yeah, that is a good idea

contracts/FlowTransactionSchedulerUtils.cdc
```diff
@@ -579,6 +582,269 @@ access(all) contract FlowTransactionSchedulerUtils {
...
+            self.revertOnFailure = revertOnFailure
+            if self.txType == COAHandlerTxType.DepositFLOW {
+                assert(amount != nil, message: "Amount is required for deposit but was not provided")
+            }
+            if self.txType == COAHandlerTxType.WithdrawFLOW {
+                assert(amount != nil, message: "Amount is required for withdrawal but was not provided")
+            }
+            if self.txType == COAHandlerTxType.Call {
+                assert(callToEVMAddress != nil && callToEVMAddress!.length == 40, message: "Call to EVM address is required for EVM call but was not provided or is invalid length (must be 40 hex chars)")
+                assert(data != nil, message: "Data is required for EVM call but was not provided")
```
m-Peter: Technically, for a `call`, you need either `data` for smart contract calls, or `value` for transfers between accounts. It's also possible that you can have both, but it's not mandatory.

PR #554, opened 2025-10-27
Make data field `access(all)`

### LINKED ISSUE #552: Scheduled Txs: expose data
### Issue To Be Solved
At the moment it is not possible to know what data something that is scheduled will run with, this means that in flowscan for instance we have no way of knowing or showing this. 

### (Optional): Suggest A Solution
Make the getData method not be access(contract)
  
### (Optional): Context

visualizing scheduled transctions in flowscan

joshuahannan:
Closes #552 

We defaulted to making the data field of a scheduled transaction `access(contract)`, but there isn't a compelling reason to restrict this, so this PR updates the getter function to be `access(all)` so it can be more easily read from anywhere.


contracts/FlowTransactionScheduler.cdc
devbugging: do we then need this getter?

devbugging (approved):

contracts/FlowTransactionScheduler.cdc
joshuahannan: oops, actually I meant to make this getter public. Getters are good because they copy the value instead of returning a reference. So this will be public and the field will remain `access(contract)`

PR #562, opened 2025-11-06
Add events for withholding node rewards

### LINKED ISSUE #446: Adds Events for nodes that get rewards slashed
### Issue To Be Solved
When nodes' stakes or rewards are slashed, there should be events that indicate what was slashed and how much.

### Suggest A Solution
* An event when the staking admin sets the nodes to slash that indicates which nodes are being slashed
* Updates to the rewards events that indicate how much was slashed for each node

May need some more discussion about what kind of events and new parameters are needed
  
### Context

Block explorers want to show this information so that Flow can be more transparent

joshuahannan:
Closes #446 

Adds two events to indicate when nodes are marked to get their rewards withheld and when the withheld rewards are actually calculated


contracts/FlowIDTableStaking.cdc
```diff
@@ -38,6 +38,7 @@ access(all) contract FlowIDTableStaking {
     /// Epoch
     access(all) event NewEpoch(totalStaked: UFix64, totalRewardPayout: UFix64, newEpochCounter: UInt64)
     access(all) event EpochTotalRewardsPaid(total: UFix64, fromFees: UFix64, minted: UFix64, feesBurned: UFix64, epochCounterForRewards: UInt64)
+    access(all) event NodeRewardsSetToSlash(nodeSlashPercentagesRemaining: {String: UFix64})
```
jordanschalm: I think it's useful to differentiate between withholding rewards and slashing. In my view, slashing is a punitive measure for a protocol violation where part of a node's initial stake amount is confiscated. 

I don't think we should refer to withholding rewards as slashing in general. In particular if we plan to pursue performance-based rewards, that blurs the line even further between "getting more rewards for good performance" and "getting less rewards for worse performance".

```suggestion
    // The single parameter is a map from node ID to the percentage of rewards that will be withheld from
    // node at each subsequent payout, until the withholding map is modified again.
    access(all) event NodeRewardWithholdingsUpdated(nodeRewardWithholdings: {String: UFix64})
```

joshuahannan: yes, you're right, I forgot about the naming distinction. Thanks for the correction! I'll update it

vishalchangrani:
should we include a ReasonCode? and then have a table either on chain or just in the docs to begin with which has the mapping of 
code to reason.
e.g.
ReasonCode 1 - "Node was down for more than 80% of time during the last epoch"
ReasonCode 2 - "Node not updated within 24hr of the Height coordinated upgrade"
Eventually we have many more reasons e.g. incorrect signature on vote for consensus nodes.

joshuahannan:
That is a good idea. I will probably need you to create that page in the docs site though because you'll know those better than I do. Once that page is on the docs site, I can add a link to it in a comment in the contract.

Do we think just having the reason code as a `UInt8` is okay?

PR #569, opened 2025-12-04
Improve addTransaction performance and add calculateFee

joshuahannan:
Includes some performance improvements to `addTransaction()` to use references instead of copying and adds a `calculateFee()` function so that users can call that to calculate the fee instead of having to call `estimate()`, which is pretty costly


contracts/FlowTransactionScheduler.cdc
```diff
@@ -1053,8 +1053,8 @@ access(all) contract FlowTransactionScheduler {
             }
 
             // Add this transaction id to the slot
-            let slotQueue = self.slotQueue[slot]!
-            if let priorityQueue = slotQueue[txData.priority] {
+            let slotQueue = &self.slotQueue[slot]! as auth(Mutate) &{Priority: {UInt64: UInt64}}
+            if let priorityQueue = *slotQueue[txData.priority] { 
                 priorityQueue[txData.id] = txData.executionEffort
                 slotQueue[txData.priority] = priorityQueue
             } else {
```
sisyphusSmiling: I was going to suggest that we avoid copying the `priorityQueue` mapping here but it looks like we can't get an `auth(Mutate)` reference to the inner mapping which is unfortunate. I would think the outer authorized reference enables us to access the inner values as authorized as well.

joshuahannan: yeah, I thought that would be true too, but I could see it potentially unintentionally exposing something.

PR #571, opened 2025-12-11
Remove old timestamps in scheduleld transaction manager cleanup

joshuahannan:
Fixes an issue in `FlowTransactionSchedulerUtils.Manager.cleanup` where we were not removing old timestamps from the array when their transactions were cleaned up. This was causing that array to grow without bounds and causing some apps to reach the computation limit.

Additionally, I removed the check for the transaction status to be `nil` or Unknown before removing because this is not sufficient to remove old transactions in time. We originally included this because we thought it might be useful for people to be able to query their executed transactions for their status, but the highest priority is actually to keep the computation limit low and apps can figure out different ways to monitor their transactions.

There is also the issue that some of these managers have already reached the computation limit, so if we just changed this check, it may not actually work to get those into a working state. Because of this, I included a conditional in the cleanup function that breaks the loop after 100 ids so that the old timestamps can be cleaned up in batches until it is in a manageable state. Then we can remove that check in an upgrade. Not sure it 100 ids is the right number because I just picked an arbitrary one. I'm open to feedback on other options for number of ids or other ways to batch it.

Another option would be to remove the cleanup function from `schedule()` entirely, but then there is no way to keep the maps from growing indefinitely without calling cleanup manually, so I would still like to keep it in there if possible.


contracts/FlowTransactionSchedulerUtils.cdc
```diff
@@ -320,9 +322,14 @@ access(all) contract FlowTransactionSchedulerUtils {
                 let ids = self.idsByTimestamp[timestamp] ?? []
                 for id in ids {
                     let status = FlowTransactionScheduler.getStatus(id: id)
-                    if status == nil || status == FlowTransactionScheduler.Status.Unknown {
-                        transactionsToRemove.append(id)
-                    }
+                    transactionsToRemove.append(id)
+                }
```
sisyphusSmiling: by removing the conditional, I think we're appending adding ids that could be scheduled. is that right?

joshuahannan: It wouldn't remove scheduled ids because it is only doing timestamps in the past, but now that you mention it, I think it might remove ids that are supposed to execute in the same block but haven't actually been executed yet. I need to think about it a bit more, but I don't think that is an issue because this just removes the scheduled transaction resource and does not affect the handler

joshuahannan: I'll still add a check that the status can't be scheduled though just to be safe

contracts/FlowTransactionSchedulerUtils.cdc
```diff
@@ -290,6 +291,7 @@ access(all) contract FlowTransactionSchedulerUtils {
                 ids.remove(at: index!)
                 if ids.length == 0 {
                     self.idsByTimestamp.remove(key: timestamp)
+                    self.sortedTimestamps.remove(timestamp: timestamp)
```
sisyphusSmiling: Since these were not removed prior, do we need some sort of backfill logic to remove irrelevant sorted timestamps?

contracts/FlowTransactionSchedulerUtils.cdc
```diff
@@ -320,9 +322,14 @@ access(all) contract FlowTransactionSchedulerUtils {
                 let ids = self.idsByTimestamp[timestamp] ?? []
                 for id in ids {
```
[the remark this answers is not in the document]
joshuahannan: yes, I was also going through and looking for places where I can borrow references. I'll do that

contracts/FlowTransactionSchedulerUtils.cdc
```diff
@@ -284,27 +286,26 @@ access(all) contract FlowTransactionSchedulerUtils {
         /// @param timestamp: The timestamp of the transaction to remove
         /// @param handlerTypeIdentifier: The type identifier of the handler of the transaction to remove
         access(self) fun removeID(id: UInt64, timestamp: UFix64, handlerTypeIdentifier: String) {
+            pre {
+                self.handlerInfos.containsKey(handlerTypeIdentifier): "Invalid handler type identifier: Handler with type identifier \(handlerTypeIdentifier) not found in manager"
+                self.idsByTimestamp.containsKey(timestamp): "Invalid timestamp: Timestamp \(timestamp) not found in manager for transaction with ID \(id)"
+            }
```
sisyphusSmiling: How does this impact the ability to remove an ID for a handler that's been removed?

joshuahannan: That isn't possible because a handler can't be removed unless all the transaction IDs for that handler have already been removed. I just put this in because before I was using `if let` to get the handler infos, but now that I am borrowing a reference, I needed to force unwrap, so I have these now just as a backup but these should never trigger

contracts/FlowTransactionSchedulerUtils.cdc
```diff
@@ -318,10 +319,21 @@ access(all) contract FlowTransactionSchedulerUtils {
...
                 for id in ids {
                     let status = FlowTransactionScheduler.getStatus(id: id)
-                    if status == nil || status == FlowTransactionScheduler.Status.Unknown {
+                    if status == nil || status! != FlowTransactionScheduler.Status.Scheduled {
                         transactionsToRemove.append(id)
+                        // Need to temporarily limit the number of transactions to remove
+                        // because some managers on mainnet have already hit the limit and we need to batch them
+                        // to make sure they get cleaned up properly
+                        // This will be removed eventually
+                        if transactionsToRemove.length > 50 {
```
joshuahannan: This limit is only temporary. The cases that we've become aware of going over the computation limit were in the 100s of iterations, so we believe that this should be sufficient to allow them to clean up without hitting the limit again. We'll be doing some more testing also to make sure

janezpodhostnik (approved):

PR #573, opened 2025-12-30
Add governance EVM direct call

m-Peter:


contracts/FlowServiceAccount.cdc
```diff
@@ -187,6 +188,11 @@ access(all) contract FlowServiceAccount {
         }
     }
 
+    access(all)
```
ramtinms: This should not be access all in my view, should be only callable by the service account

contracts/FlowServiceAccount.cdc
```diff
@@ -185,6 +186,11 @@ access(all) contract FlowServiceAccount {
                 emit IsAccountCreationRestrictedUpdated(isRestricted: enabled)
             }
         }
+
+        access(all)
+        fun governanceDirectCall(from: String, to: String, amount: UInt): EVM.Result {
+            return EVM.reclaimFundsFromAttackerEOAs(from: from, to: to, amount: amount)
```
joshuahannan: Can this only reclaim FLOW? Don't we need to do this for other token types too?

turbolent: `reclaimFundsFromAttackerEOAs` is `access(account)` (https://github.com/onflow/flow-go/pull/8280/commits/dde3be6f718a325f73c9ab0061e0d962ee60010a#diff-02837edc597afad5bd8f7d1962c129bded25adcc21a9889382c7aaad9a0e5b34R1035). Are the `EVM` contract and the `FlowServiceAccount` contract deployed to the same account?

m-Peter: And the `flow cli` command:
```bash
flow transactions send transactions/reclaim_EOA_tokens.cdc --network=mainnet-fork --signer=mainnet-fork-service "0x9D9247F5C3F3B78F7EE2C480B9CDaB91393Bf4D6" "0x00000000000000000000000235aE95896583818d" 2_030_023_297_036_000_000_000_000
```

PR #575, opened 2026-01-15
Enable concurrent fee collection

janezpodhostnik:
This PR enables concurrent fee collection by introducing child fee accounts that can collect transaction fees in parallel, reducing contention on the single FlowFees vault during high-throughput scenarios.

Changes:

- Child Fee Accounts: Fees are now distributed across multiple child accounts using getTransactionIndex() % numChildAccounts, allowing concurrent deposits without vault contention
- Updated getFeeBalance(): Now aggregates balances from the main vault plus all child fee accounts
- Updated withdrawTokensFromFeeVault(): Withdraws from the main vault first, then iterates through child accounts if more funds are needed
- New collectFeesOnChildAccounts() function: Routes deposited fees to the appropriate child account based on transaction index
- New transaction add_fee_child_accounts.cdc: Allows creating and registering child fee accounts

Falls back to the original single-vault behavior when no child accounts are configured, ensuring existing deployments continue to work.

Missing:
- [x] Fix for the cadence code checker
- [ ] check epoch transitions work normally


joshuahannan (commented):
I assume the purpose here is to have many different fee accounts so that depositing fees to the service account doesn't cause conflicts that would prevent concurrent execution, right? How many accounts do we think we would have and is that configurable based on network load or something?

joshuahannan:
@janezpodhostnik I just updated the branch with the latest from master. Are you okay with this getting merged since it is already deployed on testnet?

PR #585, opened 2026-03-12
Fix division by zero in calculateRewards when all nodes are non-operational

joshuahannan:
## Summary

- `calculateRewards` in `FlowIDTableStaking.cdc` unconditionally divides `sumRewardsWithheld` by `totalStaked - sumStakeFromNonOperationalStakers`
- If all staked nodes are marked non-operational, that denominator is `0.0`, which panics in Cadence
- This would permanently block epoch reward distribution until the non-operational list is corrected

## Fix

Guard the division with a zero-check. When there are no operational nodes, `withheldRewardsScale` is `0.0` — there are no eligible recipients to redistribute the withheld rewards to, so the scale is correctly zero. The second loop (line 1394) that distributes rewards skips all nodes via the `nonOperationalNodes[nodeID] != nil` check anyway, so the withheld rewards simply go undistributed in this edge case.

## Test plan

- [x] All existing Cadence tests pass (`make test`)
- [x] All Go tests pass (`make -C lib/go test`)
- [x] Manually verify: the reward calculation still produces correct output when a subset of nodes is non-operational (existing test coverage)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jordanschalm (approved):
Dividing by zero when all nodes are non-operational does seem kind of fitting 😀

PR #615, opened 2026-07-24
Expose FlowFees fee receiver addresses via getter and change event

### LINKED ISSUE #614: Expose FlowFees fee-receiver addresses via a getter and a change event
## Context

Since #575 (concurrent fee collection), `FlowFees.deductTransactionFee` deposits fees round-robin into child fee accounts, held as a capability list at `/storage/ChildFeeAccounts` on the FlowFees account.

Off-chain consumers (indexers, explorers, exchanges, reconciliation tools) need to know which accounts receive fees, e.g. to classify `FlowToken.TokensDeposited` events as fees rather than ordinary transfers.

This is already possible today — the list can be read from the FlowFees account's storage in a script via `getAuthAccount`, or hardcoded per network (as flow-go does in `systemcontracts.FlowFeesReceivers`). But neither is convenient: the storage read must spell the capability's exact entitlement list and silently breaks if that ever changes, and hardcoded lists go stale when receivers are added — with no on-chain signal, since `add_fee_child_accounts.cdc` writes `/storage/ChildFeeAccounts` directly rather than through a contract function that could emit an event.

## Proposal

Two small quality-of-life additions:

1. A public getter on `FlowFees`:

   ```cadence
   /// Returns the addresses of all accounts that may receive transaction
   /// fee deposits: the FlowFees account plus any child fee accounts.
   access(all) fun getFeeReceiverAddresses(): [Address]
   ```

2. An event whenever the child fee account list changes, e.g. `ChildFeeAccountsChanged(addresses: [Address])`, by routing the admin mutation through a contract function (e.g. on `Administrator`) instead of a raw storage write.

Together these would let integrators fetch the receiver set with one stable script and react to changes, instead of maintaining per-network address lists or entitlement-sensitive storage reads.

janezpodhostnik:
Closes #614

- Add `FlowFees.getFeeReceiverAddresses()`: returns the addresses that may receive
  transaction fee deposits (the FlowFees account plus all child fee accounts), so
  integrators can fetch the receiver set with one stable script instead of
  entitlement-sensitive storage reads or hardcoded per-network lists.
- Add a `ChildFeeAccountsChanged(addresses: [Address])` event, emitted with the full
  new list whenever child fee accounts are added.
- Route the admin mutation through the new
  `Administrator.createChildFeeAccounts(count:payer:)` instead of a raw storage write
  in the transaction. The contract creates the accounts itself, so the stored
  capabilities are valid, keyless, and fully owned by FlowFees by construction.


contracts/FlowFees.cdc
```diff
@@ -47,6 +52,34 @@ access(all) contract FlowFees {
...
+        if let childFeeAccounts = self.borrowChildFeeAccounts() {
+            var i = 0
+            while i < childFeeAccounts.length {
+                // concat is used because mutating functions
+                // like append cannot be called in view functions
+                addresses = addresses.concat([childFeeAccounts[i].address])
+                i = i + 1
+            }
+        }
+        return addresses
```
turbolent: The manual `while` loop iterating over the accounts can be replaced with a `for-in` loop.

It's unfortunate that `append` cannot be used, the concatenation workaround is very expensive, especially as it grows with the size of the number of accounts. 

Does `getFeeReceiverAddresses` have to be `view`? I don't see the function being used anywhere in the contracts, only in a script, where it could be non-`view`

janezpodhostnik: applied 36380dc

turbolent (approved):
Nice

